PR #18, opened 2022-12-04
allow update backupmariadbs finalizers

stefan-bergstein:
Creating backup CrinJob failed because the operator did not have the permission to update backupmariadbs finalizers  

```
$ kubectl get backupmariadbs
NAME               COMPLETE   STATUS                   MARIADB   AGE
backup-scheduled   False      Error creating CronJob   mariadb   3s
```

Error logs in operator:
```
creating CronJob: cronjobs.batch is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on
```
The problem occurred on OpenShift 4.11 

Added permission to update backupmariadbs/finalizers

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -198,6 +198,12 @@ rules:
 - apiGroups:
   - database.mmontes.io
   resources:
+  - backupmariadbs/finalizers
+  verbs:
+  - update
+- apiGroups:
+  - database.mmontes.io
+  resources:
   - usermariadbs/status
   verbs:
   - get


mmontes11:
Good catch! Thanks for the PR.

I guess that Openshift requires a slightly different set of RBAC permissions maybe?:
- https://sdk.operatorframework.io/docs/faqs/#after-deploying-my-operator-why-do-i-see-errors-like-is-forbidden-cannot-set-blockownerdeletion-if-an-ownerreference-refers-to-a-resource-you-cant-set-finalizers-on-

Our e2e tests run in vanilla Kubernetes, so we didn't catch this one. Also, we don't oficially support Openshift yet, but it is in our roadmap:
- https://github.com/mmontes11/mariadb-operator/issues/10

Contributions are welcomed!

PR #56, opened 2023-03-05
mdbClient: improve sql statement in Grant/Revoke

### LINKED ISSUE #55: [Bug] Grant controller fails to grant/revoke privileges if database name contains dashes
**Describe the bug**

Given the following CRs
```
---
kind: Database
metadata:
  name: ak-test
spec:
  mariaDbRef:
    name: mariadb
  characterSet: utf8
  collate: utf8_general_ci
---
kind: User
metadata:
  name: ak-test-user
spec:
  mariaDbRef:
    name: mariadb
  passwordSecretKeyRef:
    name: ak-test-user
    key: password
---
kind: Grant
metadata:
  name: ak-test-grant
spec:
  mariaDbRef:
    name: mariadb
  privileges:
    - "ALL"
  database: "ak-test"
  table: "*"
  username: ak-test-user
  grantOption: false
---
apiVersion: v1
kind: Secret
metadata:
  name: ak-test-user
data:
  password: NjRiYzJlZWYyYzgyN2ZlY2JmYjFiYzMzOWIyMTFkZTQ=
```

the controller fails to add grant with the following error

```
error reconciling in TemplateReconciler: error creating ak-test-grant: 1 error occurred:\n\t* error granting privileges in MariaDB: Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '-test.* TO 'ak-test-user'@'%' WITH GRANT OPTION'
```

**Expected behaviour**

I expect the controller to successfully assign the grant

**Environment details**:
- Kubernetes version: 1.25.
- mariadb-operator version: mmontes11/mariadb-operator:v0.0.9
- Install method: helm
- Install flavour: recommended

macno:
Fixes #55

### DIFF
--- a/controllers/grant_controller_test.go
+++ b/controllers/grant_controller_test.go
@@ -5,7 +5,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+	http://www.apache.org/licenses/LICENSE-2.0
 
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


macno:
@mmontes11 any idea why the CI is failing? Do I have to do something on my side?

mmontes11:
Hey @macno !

LGTM! 🚀 
Thanks for contributing to `mariadb-operator`

> any idea why the CI is failing? Do I have to do something on my side?

It's something on our side, don't worry I'm looking into it 👀

mmontes11:
Hey @macno ! 

The tests are running now but they are failing because now we are escaping the `*`:
- https://github.com/mmontes11/mariadb-operator/blob/main/controllers/grant_controller_test.go#L89..L90

Both cases should work, `*` and literal tables/databases/users with dashes, so I suggess we can wrap them with this function:
```golang
func escape(s string) string {
	if s == "*" {
		return s
	}
	return fmt.Sprintf("`%s`", s)
}
```
Does this makes sense?

It would be awesome if you could add some tests covering this tests cases here:
- https://github.com/mmontes11/mariadb-operator/blob/main/controllers/grant_controller_test.go

Thanks a lot and sorry for the issues with the CI! 🙏🏻 🙇🏻

macno:
> Hey @macno !
> 
> The tests are running now but they are failing because now we are escaping the `*`:
> 
> * https://github.com/mmontes11/mariadb-operator/blob/main/controllers/grant_controller_test.go#L89..L90
> 
> Both cases should work, `*` and literal tables/databases/users with dashes, so I suggess we can wrap them with this function:
> 
> ```go
> func escape(s string) string {
> 	if s == "*" {
> 		return s
> 	}
> 	return fmt.Sprintf("`%s`", s)
> }
> ```
> 
> Does this makes sense?
> 

I'm checking
> It would be awesome if you could add some tests cases covering this here:
> 
Ok, on it

> * https://github.com/mmontes11/mariadb-operator/blob/main/controllers/grant_controller_test.go
> 
> Thanks a lot and sorry for the issues with the CI! 🙏🏻 🙇🏻

no problem

mmontes11 (approved):
Thanks for your contribution @macno !

PR #60, opened 2023-03-06
User sample: add comment to maxUserConnections

macno:
Since this property is immutable and defaults to 10, let state it clearly

### DIFF
--- a/config/samples/mariadb_v1alpha1_user.yaml
+++ b/config/samples/mariadb_v1alpha1_user.yaml
@@ -8,4 +8,5 @@ spec:
   passwordSecretKeyRef:
     name: user
     key: password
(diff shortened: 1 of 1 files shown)


macno:
A possible improvement could be to remove the immutability flag on this property and update the value

mmontes11:
Hey @macno, thanks for your contribution! 🚀 

At the moment, we are keeping the `maxUserConnections` field inmutable to be able to reusre the [TemplateReconciler](https://github.com/mmontes11/mariadb-operator/blob/main/pkg/controller/template/controller.go). Otherwise, we will need to have a custom controller to perform `ALTER ...` whenever the field changes.

Anyway, good point, we will keep that in mind for the future 👍🏻

PR #83, opened 2023-04-05
Fix wrong status condition type for MariaDB restore

Vidalee:
It seems to me that there might be an issue with the condition types for the MariaDB resource during the backup restoration process. Could it be possible that the types were swapped?

I think that the type for the "Restoring backup" condition should be `Bootstrapped `as it is part of the bootstrapping process. Please let me know if I am mistaken.

### DIFF
--- a/controllers/mariadb_controller.go
+++ b/controllers/mariadb_controller.go
@@ -414,13 +414,13 @@ func (r *MariaDBReconciler) patcher(ctx context.Context, mariaDb *mariadbv1alpha
 			})
 		} else {
 			c.SetCondition(metav1.Condition{
-				Type:    mariadbv1alpha1.ConditionTypeReady,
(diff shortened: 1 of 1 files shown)


mmontes11:
Hey @Vidalee !

The condition type for the "Restoring backup" condition is "Ready" because we want to output this to the user using the printer columns defined here:
- https://github.com/mariadb-operator/mariadb-operator/blob/main/api/v1alpha1/mariadb_types.go#L202

Basically, we only display the "Ready" one to the user in this case, and not the "Bootstrapped" one, altought it is kinda related in this case.

There's room for improvovement regarding the status conditions, so happy to keep the discussion in an issue if needed. Thanks for reporting this 👍🏻

Vidalee:
Makes sense, it's not a problem on my end, I was just looking at the possible conditions of this resource to gain a more thorough understanding of all the possible states 🙂

Issue #86, opened 2023-04-13
[Feature] Add possibility to specify user name in User CRD spec


Vidalee:
**Is your feature request related to a problem? Please describe.**

Currently the name of the user in the database will be the name of the CRD. I have multiple MariaDB running on the same namespace and I want some users in the databases to have the same name without sharing the same CRD (since a CRD name is unique in its namespace).

**Describe the solution you'd like**

Add a (optional?) "username"/"name" field in the spec of the CRD that will define the name of the user in MariaDB.

**Describe alternatives you've considered**

Splitting the MariaDB CRDs in multiple namespace, not possible on my end.

**Additional context**

Example: If I create a "foo" user in the database A, I cannot create a user "foo" in the database B since there is already a User CRD named "foo" in the namespace.

**Environment details**:
- mariadb-operator version: v0.0.11

PR #87, opened 2023-04-14
feat: add optional username to User CRD

Vidalee:
Solves #86 by adding an optional `name` field on User.spec.

Works well on my end, although it seems like this field is not inmutable. I managed to change the name value and patch the CRD, which created a new user in the database while keeping the first one. Is the `webhook:"inmutable"`` tag not enough ?

This is my first time writing Go code so I may be wrong about how tags work in Go.

I added a 80 characters limit to the user names following [the MariaDB documentation](https://mariadb.com/kb/en/identifier-names/#maximum-length).


mmontes11 (approved):
Thanks for your contribution @Vidalee ! 🚀 

The changes LGTM. The validating webhook that takes care of inmutability is not deployed when developing locally, but the annotation you added should be enough.

PR #105, opened 2023-05-09
feat: reconcile User using custom username if defined

### LINKED ISSUE #104: [Bug] User with username defined in spec does not get dropped at reconciliation
**Describe the bug**

User with username defined in spec (spec.name) are not dropped at reconciliation or when the CRD is deleted.

**Expected behaviour**

**Additional context**

Seems like it is because the DropUser method of the controller use the CRD name even if `spec.name` is set.

**Environment details**:
- Kubernetes version:
- mariadb-operator version: 0.0.12
- Install method: helm, OLM or static manifests
- Install flavour: minimal, recommended or custom

Vidalee:
Fixes #104 

It is my bad on this one, forgot that the behavior should be applied on the user deletion too.

### DIFF
--- a/api/v1alpha1/user_types.go
+++ b/api/v1alpha1/user_types.go
@@ -64,6 +64,13 @@ type User struct {
 	Status UserStatus `json:"status,omitempty"`
 }
 
+func (u *User) UsernameOrDefault() string {
+	if u.Spec.Name != "" {
+		return u.Spec.Name
+	}
+	return u.Name
+}
+
 func (u *User) IsBeingDeleted() bool {
 	return !u.DeletionTimestamp.IsZero()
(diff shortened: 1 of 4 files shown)


mmontes11 (changes requested):
This will work, but can we add a helper to the `User` struct so we can reuse the same logic in both the creation and the deletion?, something like:
```go
func (u *User) UsernameOrDefault() string {
	if u.Spec.Name != "" {
		return u.Spec.Name
	}
	return u.Name
}
```

Vidalee:
Done 😄

mmontes11:
LGTM! Thanks for your contribution 🙇🏻 🙏🏻

Issue #108, opened 2023-05-11
[Bug] root password is not set


tbe:
**Describe the bug**
After a fresh install, created clusters fail to get ready.

The healtcheck reports permission denied. Execing into the pod, and trying to log in with the credentials from the provided secret results in the same error.

It seems like the entrypoint does not set the correct password.

Log from the first start of the pod:
```
2023-05-11 12:47:17+00:00 [Note] [Entrypoint]: Entrypoint script for MariaDB Server 1:10.7.4+maria~focal started.
2023-05-11 12:47:17+00:00 [Note] [Entrypoint]: Switching to dedicated user 'mysql'
2023-05-11 12:47:17+00:00 [Note] [Entrypoint]: Entrypoint script for MariaDB Server 1:10.7.4+maria~focal started.
2023-05-11 12:47:18+00:00 [Note] [Entrypoint]: Initializing database files


PLEASE REMEMBER TO SET A PASSWORD FOR THE MariaDB root USER !
To do so, start the server, then issue the following command:

'/usr/bin/mysql_secure_installation'

which will also give you the option of removing the test
databases and anonymous user created by default.  This is
strongly recommended for production servers.

See the MariaDB Knowledgebase at https://mariadb.com/kb

Please report any problems at https://mariadb.org/jira

The latest information about MariaDB is available at https://mariadb.org/.

Consider joining MariaDB's strong and vibrant community:
https://mariadb.org/get-involved/

2023-05-11 12:47:41+00:00 [Note] [Entrypoint]: Database files initialized
2023-05-11 12:47:41+00:00 [Note] [Entrypoint]: Starting temporary server
2023-05-11 12:47:41+00:00 [Note] [Entrypoint]: Waiting for server startup
2023-05-11 12:47:41 0 [Note] mariadbd (server 10.7.4-MariaDB-1:10.7.4+maria~focal) starting as process 117 ...
2023-05-11 12:47:41 0 [Note] InnoDB: Compressed tables use zlib 1.2.11
2023-05-11 12:47:41 0 [Note] InnoDB: Number of transaction pools: 1
2023-05-11 12:47:41 0 [Note] InnoDB: Using crc32 + pclmulqdq instructions
2023-05-11 12:47:41 0 [Note] InnoDB: Using Linux native AIO
2023-05-11 12:47:41 0 [Note] InnoDB: Initializing buffer pool, total size = 134217728, chunk size = 134217728
[...]

PR #113, opened 2023-05-13
Bump MariaDB default version to LTS

mmontes11:
Bump default image to MariaDB LTS as @grooverdan  suggested [here](https://github.com/mariadb-operator/mariadb-operator/issues/108#issuecomment-1544925693)

### DIFF
--- a/api/v1alpha1/mariadb_webhook_test.go
+++ b/api/v1alpha1/mariadb_webhook_test.go
@@ -240,7 +240,7 @@ var _ = Describe("MariaDB webhook", func() {
 					},
 					Image: Image{
 						Repository: "mariadb",
-						Tag:        "10.7.4",
+						Tag:        "10.11.3",
 					},
 					Port: 3306,
 					VolumeClaimTemplate: corev1.PersistentVolumeClaimSpec{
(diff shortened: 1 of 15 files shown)


grooverdan:
I was thinking having `lts` as the tag.

I think you're trying to compromise between good practices of pinning to a version and having a codebase that doesn't need to be frequently changed. Is the minor version going to be bumped every ~3 months? Is `10.11` with a comment above on version pinning.

mmontes11:
Hey @grooverdan! The motivation behind using a pinned version instead of `lts` is that the operator doesn't currently handle [version upgrades](https://github.com/mariadb-operator/mariadb-operator/issues/118), meaning that they would be handled by Kubernetes `StatefulSet`, which is not aware of MariaDB state so it could potentially lead to downtime. 

> I think you're trying to compromise between good practices of pinning to a version and having a codebase that doesn't need to be frequently changed. Is the minor version going to be bumped every ~3 months?

Yes, I can commit to bump the version every ~3 months, What would be the best place to keeep track of the releases? Maybe here?:
- https://mariadb.org/mariadb/all-releases/

> Is 10.11 with a comment above on version pinning.

Can do that, will add pointing to the releases page when confirming the above ☝🏻

grooverdan:
API around https://downloads.mariadb.org/rest-api/mariadb/10.11.

Alternately a GH API around https://github.com/docker-library/official-images/blob/master/library/mariadb changes. Also docker hub api on library/mariadb.

PR #120, opened 2023-05-16
feat: extended information on connection secret

grilix:
There are times when an application expects the connection information to be specified as database, host, port, username and password instead of a single DSN string.

These applications, would expect an environment that is something like the following:

```yml
env:
  - name: APP_DATABASE_HOST
    valueFrom:
      secretKeyRef:
        name: db-secret
        key: host
  - name: APP_DATABASE_PORT
    valueFrom:
      secretKeyRef:
        name: db-secret
        key: port
  - name: APP_DATABASE_NAME
    valueFrom:
      secretKeyRef:
        name: db-secret
        key: name
  - name: APP_DATABASE_USER
    valueFrom:
      secretKeyRef:
        name: db-secret
        key: username
  - name: APP_DATABASE_PASS
    valueFrom:
      secretKeyRef:
        name: db-secret
        key: password
```

For achieving this, we can include the extra keys on the secretTemplate for the connection:

```yaml
kind: MariaDB
metadata:
  name: myapp-mariadb
spec:
  # ...
  connection:
    secretName: db-connection
    secretTemplate:
      hostKey: host
      portKey: port
      databaseKey: name
      usernameKey: user
      passwordKey: pass
```

This will simplify working with such applications without having to duplicate the knowledge about these values over the configuration.


mmontes11 (approved):
LGTM. Thanks for your contribution @grilix 🦭 🙇🏻

PR #134, opened 2023-06-01
examples: added quotes in liveness/readiness probes

roccazzella:
In the "plain" mariadb example manifest, we have a custom liveness/readiness probe.
The probes call the `mysql` tool passing the root password as a parameter: the password is not quoted.

If the (custom) root password contains spaces, the probes will fail.


mmontes11:
Thanks for your contribution @roccazzella ! 🙏🏻

PR #135, opened 2023-06-05
set labels for svc so servicemonitor can find it

wittdennis:
The Prometheus Operator can't currently pick up mariadb instances because the created mariadb service doesn't have the labels the ServiceMonitor looks for

### DIFF
--- a/pkg/builder/service_builder.go
+++ b/pkg/builder/service_builder.go
@@ -33,6 +33,7 @@ func (b *Builder) BuildService(mariadb *mariadbv1alpha1.MariaDB, key types.Names
 		metadata.NewMetadataBuilder(key).
 			WithMariaDB(mariadb).
 			WithAnnotations(opts.Annotations).
+			WithLabels(opts.Selectorlabels).
 			Build()
 	selectorLabels :=
 		labels.NewLabelsBuilder().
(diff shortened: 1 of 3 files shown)


pkg/builder/service_builder.go
```diff
@@ -33,6 +33,7 @@ func (b *Builder) BuildService(mariadb *mariadbv1alpha1.MariaDB, key types.Names
 		metadata.NewMetadataBuilder(key).
 			WithMariaDB(mariadb).
 			WithAnnotations(opts.Annotations).
+			WithLabels(opts.Selectorlabels).
```
mmontes11, line 36: I think it would be easier to do the following:
```go
	selectorLabels :=
		labels.NewLabelsBuilder().
			WithMariaDBSelectorLabels(mariadb).
			WithLabels(opts.Selectorlabels).
			Build()
	objMeta :=
		metadata.NewMetadataBuilder(key).
			WithMariaDB(mariadb).
			WithAnnotations(opts.Annotations).
			WithLabels(selectorLabels).
			Build()

```
What do you think?

mmontes11 (changes requested):
Thanks for your contribution @wittdennis ! 🙏🏻 

LGTM, left a suggestion, could you PTAL?

pkg/builder/service_builder.go
wittdennis: Looks good to me. Updated the solution

mmontes11:
Thanks again for your contribution @wittdennis ! 🚀

PR #157, opened 2023-07-18
Use [mariadb] as group for MariaDB server

grooverdan:
[mariadb] has been read by the MariaDB server for a long time.  (5.5 and probably 5.3)

As this operator is container to MariaDB use its a good place to remove some references to the diverging heritage of mysql.

### DIFF
--- a/controllers/mariadb_controller_test.go
+++ b/controllers/mariadb_controller_test.go
@@ -393,7 +393,7 @@ var _ = Describe("MariaDB replication", func() {
 						},
 					},
 					MyCnf: func() *string {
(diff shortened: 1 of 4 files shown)


mmontes11:
Thanks for your contributions @grooverdan ! changes LGTM

To complete the work, I will be making the same changes in the `initContainer` used by galera, which popuplates the Galera configuration file:

https://github.com/mariadb-operator/init/blob/307ececa23d847b158836d3d29263b867e1e2aaf/pkg/config/config.go#L39

I will take care of this after merging 👍🏻

Issue #65, opened 2023-03-17
Functionality to customize commands when executing mysqldumps (backups)


parkerg21:
**Is your feature request related to a problem? Please describe.**
When performing backups with large databases we would like the ability to not lock tables.
Before using operator we were creating backups using: `mysqldump --single-transaction`

**Describe the solution you'd like**

The ability to customize the logical backup options such as locking tables, using single-transaction, other flags

Alternatives would be running a different backup process, using standalone cron job

**Environment details**:
- Kubernetes version: v1.23.5
- Openshift version: 4.10.0-0.okd-2022-07-09-073606
- mariadb-operator version: v0.0.5
- Install method: static manifest synced with argoCD
- Install flavour: custom

mmontes11:
Hey @parkerg21 ! This is defintely a feature we a feature we should support eventually, adding it to the mid term roadmap. We probably might need a refactor in the `Backup` CRDs to cleanly support this. 

Thanks! 🙏🏻

mmontes11:
The ability to provide different opts to a backup has been implemented in this PR:
- https://github.com/mariadb-operator/mariadb-operator/pull/158

In order to close this issue, we need to expose the options in the `Backup` CRD

mmontes11:
Hey @parkerg21 ! This has been implemented and it will be part of the next release.

Now you are able to provide `spec.args` in the `Backup` CRD:
https://github.com/mariadb-operator/mariadb-operator/blob/main/examples/manifests/mariadb_v1alpha1_backup.yaml#L15..L17

By default, the following args are used if none are provided:
https://github.com/mariadb-operator/mariadb-operator/blob/main/pkg/command/backup/logical.go#L17

PR #162, opened 2023-07-18
Support for backup arguments

mmontes11:
Closes https://github.com/mariadb-operator/mariadb-operator/issues/65

cc @grooverdan


grooverdan:
thank you for the clean up and exposing the argument.

PR #163, opened 2023-07-19
healthcheck.sh for liveness/readiness

grooverdan:
Per
https://mariadb.org/mariadb-server-docker-official-images-healthcheck-without-mysqladmin/, mysql isn't included in mariadb:11.0+ images. And without enforcing a tcp option there are initialization stages that may report true ahead of the actual readiness.

ref: https://mariadb.com/kb/en/using-healthcheck-sh-script/

### DIFF
--- a/controllers/suite_test_data.go
+++ b/controllers/suite_test_data.go
@@ -99,9 +99,9 @@ func createTestData(ctx context.Context, k8sClient client.Client) {
 					ProbeHandler: v1.ProbeHandler{
 						Exec: &v1.ExecAction{
 							Command: []string{
-								"bash",
-								"-c",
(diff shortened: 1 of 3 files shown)


grooverdan:
Evidently not working correctly. Timeout is occuring.

grooverdan:
Adding galera_online test to official image container to match the existing test

https://github.com/MariaDB/mariadb-docker/commit/1792faf9319427c365cbc1918e575e223ae55470

mmontes11:
> Adding galera_online test to official image container to match the existing test
> https://github.com/MariaDB/mariadb-docker/commit/1792faf9319427c365cbc1918e575e223ae55470

Wouldn't we need to bump to this server image version for the liveness/readiness probes to work properly? What about the users using previous images of the server?

grooverdan:
Yes, I'll work out a backward compatible way.

PR #174, opened 2023-07-28
Fixed resource deletion

matthieugusmini:
Moved the logic adding the finalizer so it is only added if the resource was successfully created.


mmontes11:
Thanks again for your contributions @matthieugusmini ! 🦭 🚀

Issue #169, opened 2023-07-27
[Feature] Support `metadata` in `volumeClaimTemplates` 


mmontes11:
**Describe the solution you'd like**
Be able to specify `metadata` in the `volumeClaimTemplates` fields, both in the MariaDB and the Galera one. This can be achieved by changing from `corev1.PersistentVolumeClaimSpec` to `corev1.PersistentVolumeClaim` in the following types:
- https://github.com/mariadb-operator/mariadb-operator/blob/a820746605ed15aadd7c969e230808ca4f5cf704/api/v1alpha1/mariadb_types.go#L104
- https://github.com/mariadb-operator/mariadb-operator/blob/a820746605ed15aadd7c969e230808ca4f5cf704/api/v1alpha1/mariadb_galera_types.go#L167

Alternatively, consider creating our own `mariadbv1alpha1.PersistentVolumeClaim`.

**Additional context**
Sometimes annotations need to be added to PVCs:
- https://www.vcluster.com/docs/architecture/storage#how-does-it-work

vavrajosef:
I would like to work on this, please assign it to me.

mmontes11:
Absolutely, go ahead. Reach out if you have any questions!

vavrajosef:
I have created [PR](https://github.com/mariadb-operator/mariadb-operator/pull/175), where I replaced Spec. Please review it. Also, you were talking about changing some annotations and I have read the article in the description of this issue, but I don't think I understand what annotations should be updated.
When I forked this repository I had trouble running some tests, so I am not 100% sure my change didn't break anything.

mmontes11:
Hey @vavrajosef, just left a [comment](https://github.com/mariadb-operator/mariadb-operator/pull/175#pullrequestreview-1553270879) in the PR regarding this!

PR #175, opened 2023-07-29
Feature/169 metadata support

vavrajosef:
Closes issue #169

### DIFF
--- a/pkg/builder/statefulset_builder.go
+++ b/pkg/builder/statefulset_builder.go
@@ -151,21 +151,15 @@ func buildStsUpdateStrategy(mariadb *mariadbv1alpha1.MariaDB) appsv1.StatefulSet
 }
 
 func buildStsVolumeClaimTemplates(mariadb *mariadbv1alpha1.MariaDB) []corev1.PersistentVolumeClaim {
+	vctpl := mariadb.Spec.VolumeClaimTemplate
+	vctpl.ObjectMeta.Name = StorageVolume
 	pvcs := []corev1.PersistentVolumeClaim{
-		{
-			ObjectMeta: metav1.ObjectMeta{
-				Name: StorageVolume,
-			},
-			Spec: mariadb.Spec.VolumeClaimTemplate.Spec,
-		},
+		vctpl,
 	}
 	if mariadb.Galera().Enabled {
-		pvcs = append(pvcs, corev1.PersistentVolumeClaim{
-			ObjectMeta: metav1.ObjectMeta{
-				Name: galeraresources.GaleraConfigVolume,
-			},
-			Spec: mariadb.Galera().VolumeClaimTemplate.Spec,
-		})
+		galeraVctpl := *mariadb.Galera().VolumeClaimTemplate
+		galeraVctpl.ObjectMeta.Name = galeraresources.GaleraConfigVolume
+		pvcs = append(pvcs, galeraVctpl)
 	}
 	return pvcs
 }
--- a/pkg/builder/statefulset_builder.go
+++ b/pkg/builder/statefulset_builder.go
@@ -156,7 +156,7 @@ func buildStsVolumeClaimTemplates(mariadb *mariadbv1alpha1.MariaDB) []corev1.Per
 			ObjectMeta: metav1.ObjectMeta{
 				Name: StorageVolume,
 			},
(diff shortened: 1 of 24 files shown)


pkg/builder/statefulset_builder.go
```diff
@@ -156,15 +156,15 @@ func buildStsVolumeClaimTemplates(mariadb *mariadbv1alpha1.MariaDB) []corev1.Per
...
+			Spec: mariadb.Spec.VolumeClaimTemplate.Spec,
 		},
 	}
 	if mariadb.Galera().Enabled {
 		pvcs = append(pvcs, corev1.PersistentVolumeClaim{
 			ObjectMeta: metav1.ObjectMeta{
 				Name: galeraresources.GaleraConfigVolume,
 			},
-			Spec: *mariadb.Galera().VolumeClaimTemplate,
+			Spec: mariadb.Galera().VolumeClaimTemplate.Spec,
```
mmontes11, line 167:
```suggestion
func buildStsVolumeClaimTemplates(mariadb *mariadbv1alpha1.MariaDB) []corev1.PersistentVolumeClaim {
	vctpl := mariadb.Spec.VolumeClaimTemplate
	vctpl.ObjectMeta.Name = StorageVolume
	pvcs := []corev1.PersistentVolumeClaim{
		vctpl,
	}
	if mariadb.Galera().Enabled {
		galeraVctpl := *mariadb.Galera().VolumeClaimTemplate
		galeraVctpl.ObjectMeta.Name = galeraresources.GaleraConfigVolume
		pvcs = append(pvcs, galeraVctpl)
	}
	return pvcs
}
```

mmontes11 (changes requested):
Thanks for your contribution, a couple of things:
- We need to add the `metadata` field to the `volumeClaimTemplates`, which is the main purpose of this feature. I've added a commit suggestion for that
- Run `make gen` to generate code and CRDs.  The CI is failing because of this
- To run test locally:
```bash
make cluster
make install
make net
make test
```

Let me know if you have any question!

pkg/builder/statefulset_builder.go
mmontes11, line 167: can we do something like this?:

```golang
func buildStsVolumeClaimTemplates(mariadb *mariadbv1alpha1.MariaDB) []corev1.PersistentVolumeClaim {
	vctpl := mariadb.Spec.VolumeClaimTemplate
	vctpl.ObjectMeta.Name = StorageVolume
	pvcs := []corev1.PersistentVolumeClaim{
		vctpl,
	}
	if mariadb.Galera().Enabled {
		galeraVctpl := *mariadb.Galera().VolumeClaimTemplate
		galeraVctpl.ObjectMeta.Name = galeraresources.GaleraConfigVolume
		pvcs = append(pvcs, galeraVctpl)
	}
	return pvcs
}
```

mmontes11 (changes requested):

pkg/builder/statefulset_builder.go
vavrajosef: I have applied this suggestion.

vavrajosef:
Thanks for looking into this and thanks for the feedback. I have updated the PR. Please let me know if this is what you had in mind.

mmontes11:
Hey @vavrajosef ! thanks for your changes, they look fine!

However, I've just realized that the [issue](https://github.com/mariadb-operator/mariadb-operator/issues/169) definition implies breaking changes in the CRDs, something we want to avoid whenever possible. What I would suggest is defining our own `VolumeClaimTemplate` type embedding the `corev1.PersistentVolumeClaimSpec ` and adding annotations and labels, which is the main purpose of this issue. It would be something like:

```golang
type VolumeClaimTemplate struct {
    corev1.PersistentVolumeClaimSpec 
    Labels      map[string]string `json:"labels,omitempty"`
    Annotations map[string]string `json:"annotations,omitempty"`
}
``` 
This way, the existing `MariaDB` definitions in our users clusters are still valid and no changes from their side are needed to upgrade.

Sorry for misguiding you! I will be picking your changes to another branch and adding you as a co-author.

mmontes11:
Closed in favor of:

https://github.com/mariadb-operator/mariadb-operator/pull/185

Thanks for your contribution @vavrajosef

PR #176, opened 2023-07-31
Configure account name's hostname component

matthieugusmini:


api/v1alpha1/grant_types.go
```diff
@@ -83,6 +85,13 @@ func (g *Grant) MariaDBRef() *MariaDBRef {
 	return &g.Spec.MariaDBRef
 }
 
+func (g *Grant) HostnameOrDefault() string {
```
mmontes11, line 88: What about adding the `AccountName` method to the `Grant` for the sake of unification?

api/v1alpha1/user_types.go
```diff
@@ -64,11 +68,8 @@ type User struct {
 	Status UserStatus `json:"status,omitempty"`
 }
 
-func (u *User) UsernameOrDefault() string {
-	if u.Spec.Name != "" {
-		return u.Spec.Name
-	}
-	return u.Name
+func (u *User) AccountName() string {
+	return fmt.Sprintf("'%s'@'%s'", u.usernameOrDefault(), u.hostnameOrDefault())
```
mmontes11, line 72: great 👍🏻

mmontes11 (changes requested):
I think it would be worth adding `AccountName` method to `Grant` so it looks like `User`.  This way we can pass it directly here, as you did with `CreateUser` and `DropUser`: 
https://github.com/mariadb-operator/mariadb-operator/blob/f25831199002c637fdb67f258e333ce91e48f9c5/pkg/client/client.go#L220

Other than that, LGTM!

api/v1alpha1/grant_types.go
matthieugusmini: @mmontes11 Sure could do that but will also imply that we change the `GrantOpts` struct definition, replacing the `Username` and `Host` field by a `AccountName` string.

Also it seems like the `Privileges`, `Database`, `Table` and `AccountName` fields of the `GrantOpts` are always required. What do you think about changing the signature of the `Grant` and `Revoke` methods so they take these values as regular parameters and we would only keep the optional arguments of the option in the `GrantOpts` 🤔

Also does not seem like the [`REVOKE` statement](https://mariadb.com/kb/en/revoke/) takes any optional argument so what about dropping the `GrantOpts` from the signature? And if needed we later we could add a dedicated `RevokeOpts` struct 😄 

Something such as:
```go
func (c *Client) Grant(ctx context.Context, privileges []string, database, table, accountName string, opts GrantOpts) error

func (c *Client) Revoke(ctx context.Context, privileges []string, database, table, accountName string) error
```

mmontes11: In that case the `GrantOpts` would be just the `GrantOption` right?
```golang
type GrantOpts struct {
	GrantOption bool
}
```
We should still support revoking the `GRANT OPTION` so we would need to pass the `GrantOpts` to `Revoke` as well. For indicating that is nullable, we could pass it with a pointer, as follows:
```golang
func (c *Client) Grant(ctx context.Context, privileges []string, database, table, accountName string, opts *GrantOpts) error

func (c *Client) Revoke(ctx context.Context, privileges []string, database, table, accountName string,  opts *GrantOpts) error
```
What do you think?

matthieugusmini: Yes only the `GrantOption` will remain. Didn't take a closer look at the `GRANT` statement document but there's might be more support options by MariaDB itself though.

Passing a pointer to suggest the fact that the `GrantOpts` is optional sounds good. We could  also pass it as a [functional option](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) so other developers who don't need options when calling these methods doesn't even need to pass any value when calling it.

I'll push an implementation example to make it easier to visualize 😉

mmontes11: Agreed, let's go with an options pattern interface. You have an example here:
- https://github.com/mariadb-operator/mariadb-operator/blob/main/pkg/controller/galera/controller.go#L53

mmontes11:
Some conflicts after merging this [PR](https://github.com/mariadb-operator/mariadb-operator/pull/178), sorry about that. Let me know if you have any questions!

pkg/client/client.go
```diff
@@ -222,42 +222,66 @@ func (c *Client) UserExists(ctx context.Context, username string) (bool, error)
...
+	}
+}
+
+func (c *Client) Grant(
+	ctx context.Context,
+	privileges []string,
+	database string,
+	table string,
+	accountName string,
+	opts ...GrantOption,
```
mmontes11, line 243: awesome 🚀

matthieugusmini:
@mmontes11 Yeah sure you can merge this PR if it looks good to you 👍🏼

pkg/client/client.go
```diff
@@ -188,8 +188,8 @@ type CreateUserOpts struct {
 	MaxUserConnections int32
 }
 
-func (c *Client) CreateUser(ctx context.Context, username string, opts CreateUserOpts) error {
-	query := fmt.Sprintf("CREATE USER IF NOT EXISTS '%s'@'%s' ", username, "%")
+func (c *Client) CreateUser(ctx context.Context, accountName string, opts CreateUserOpts) error {
```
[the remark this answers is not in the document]
mmontes11: > `client.Client`

This is the kubernetes client provided by `controller-runtime`

> What about renaming the package to something more explicit about its intent?

Agreed the naming is not great given the number of different clients we have. I would suggest `mariadbclient` to be more explicit. Feel free to raise another PR

mmontes11 (approved):

PR #177, opened 2023-07-31
Controller/Connection: Close DB connection

lgohyipex:
Close the DB connection after the healthcheck otherwise it will exhaust the user connections set by `max_user_connections`


mmontes11:
Thanks a lot for your contribution @lgohyipex !

PR #180, opened 2023-08-02
Add custom dsn format to Connections

lgohyipex:
This PR adds the possibility to specify a custom format for the DSN in a different key than `secretKey`.

The use case is applications parsing other style of DSN like : `mysql://[user]:[pass]@[host]:[port]/[database]?[params]


lgohyipex:
@mmontes11 What do you think ?

api/v1alpha1/base_types.go
```diff
@@ -41,14 +41,16 @@ type MariaDBRef struct {
...
-	UsernameKey *string           `json:"usernameKey,omitempty"`
-	PasswordKey *string           `json:"passwordKey,omitempty"`
-	HostKey     *string           `json:"hostKey,omitempty"`
-	PortKey     *string           `json:"portKey,omitempty"`
-	DatabaseKey *string           `json:"databaseKey,omitempty"`
+	Labels       map[string]string `json:"labels,omitempty"`
+	Annotations  map[string]string `json:"annotations,omitempty"`
+	Key          *string           `json:"key,omitempty"`
+	FormatKey    *string           `json:"formatKey,omitempty"`
+	FormatString *string           `json:"format,omitempty"`
```
mmontes11, line 48: What about using always the same `Key` field for simplicity? If the user provides the `FormatString`,  we would override the `Key` field content with the given format, there is no need to have multiple fields for the same purpose.

Also, can we rename `FormatString` -> `Format` for simplicity as well?

controllers/connection_controller_test.go
```diff
@@ -131,6 +132,58 @@ var _ = Describe("Connection controller", func() {
 					},
 					WantDsn: "test:test@tcp(mariadb-test.default.svc.cluster.local:3306)/test?parseTime=true",
 				},
+				{
```
mmontes11, line 135: Thanks for the test case ❤️

mmontes11 (changes requested):
Thanks a lot for your contribution @lgohyipex, good stuff!  This adds a lot of flexibility for configuring connection strings for applications. 🚀 

I think you are on the right path, I've letf some suggestions regarding the model and I have a couple more asks if you don't mind:
- Validate the `format` field in the validation webhook:
```golang
_, err := template.New("").Parse(*formatString)
if err != nil {
    // deny the request
}
```
The validation should happen when the `MariaDB` is [created](https://github.com/mariadb-operator/mariadb-operator/blob/main/api/v1alpha1/mariadb_webhook.go#L60) and [updated](https://github.com/mariadb-operator/mariadb-operator/blob/main/api/v1alpha1/mariadb_webhook.go#L78). This is for detecting invalid formats as early as possible, preventing the controllers from reconciling invalid templates.You can add a test for the webhook [here](https://github.com/mariadb-operator/mariadb-operator/blob/main/api/v1alpha1/mariadb_webhook_test.go) as well 
- Add an example of usage in the [examples](https://github.com/mariadb-operator/mariadb-operator/tree/main/examples/manifests): `mariadb_v1alpha1_connection_custom_dsn.yaml` 

Let me know what you think 👍🏻

api/v1alpha1/base_types.go
lgohyipex: It was my first idea, but the `Key` is used to get the connection details to perform the healthcheck.
https://github.com/mariadb-operator/mariadb-operator/blob/main/controllers/connection_controller.go#L228

I didn't look more into how to get the connection details an other way.
Maybe pushing the call to `healthCheck` later in `reconcileSecret` would allow to pass the built ([from here mariadbv1alpha1.Connection](https://github.com/mariadb-operator/mariadb-operator/blob/main/controllers/connection_controller.go#L184)) DSN directly as argument instead of relying on the content of the secret.

mmontes11: That's a good point. I think the real problem here is that the `healthCheck` is coupled to the `Secret` and its DSN format. I would suggest breaking this dependency so we can have a format agnostic healthchecks and potentially have any DSN format in the `Key` field. We need a couple of changes though:

First of call we need to provide a way for the client (`pkg/client/client.go`) to perform a connection and not depend on the DSN:
```golang
func ConnectWithOpts(opts Opts) (*sql.DB, error) {
	dsn, err := BuildDSN(opts)
	if err != nil {
		return nil, fmt.Errorf("error building DNS: %v", err)
	}
	return Connect(dsn)
}
``` 
Then, in the `pkg/connection_controller.go`, the first think will be making the `healthCheck` unaware of the DSN format by adding an extra `mariadbclient.Opts` argument:
```golang
func (r *ConnectionReconciler) healthCheck(ctx context.Context, conn *mariadbv1alpha1.Connection, clientOpts mariadbclient.Opts) error {
	log.FromContext(ctx).V(1).Info("Checking connection health")
	db, err := mariadbclient.ConnectWithOpts(clientOpts)
	if err != nil {
		var connErr *multierror.Error
		connErr = multierror.Append(connErr, err)

		patchErr := r.patchStatus(
			ctx,
			conn,
			r.ConditionReady.PatcherHealthy(fmt.Errorf("failed to connect: %v", err)),
		)
		return multierror.Append(connErr, patchErr)
	}
	defer db.Close()

	if err := r.patchStatus(ctx, conn, r.ConditionReady.PatcherHealthy(nil)); err != nil {
		return fmt.Errorf("error patching connection status: %v", err)
	}
	return nil
}
``` 
Finally, we can hoist the creation of `mariadbclient.Opts` to the top of `reconcileSecret` and use it for both performing the healthheck and creating the DSN, so we have the same source truth:
```golang
func (r *ConnectionReconciler) reconcileSecret(ctx context.Context, conn *mariadbv1alpha1.Connection,
	mdb *mariadbv1alpha1.MariaDB) error {
	key := types.NamespacedName{
		Name:      conn.SecretName(),
		Namespace: conn.Namespace,
	}
	password, err := r.RefResolver.SecretKeyRef(ctx, conn.Spec.PasswordSecretKeyRef, conn.Namespace)
	if err != nil {
		return fmt.Errorf("error getting password for connection DSN: %v", err)
	}
	var host string
	if conn.Spec.ServiceName != nil {
		objMeta := metav1.ObjectMeta{
			Name:      *conn.Spec.ServiceName,
			Namespace: mdb.ObjectMeta.Namespace,
		}
		host = statefulset.ServiceFQDN(objMeta)
	} else {
		host = statefulset.ServiceFQDN(mdb.ObjectMeta)
	}
	mdbOpts := mariadbclient.Opts{
		Username: conn.Spec.Username,
		Password: password,
		Host:     host,
		Port:     mdb.Spec.Port,
		Params:   conn.Spec.Params,
	}
	if conn.Spec.Database != nil {
		mdbOpts.Database = *conn.Spec.Database
	}

	var existingSecret corev1.Secret
	if err := r.Get(ctx, key, &existingSecret); err == nil {
		if err := r.healthCheck(ctx, conn, mdbOpts); err != nil {
			log.FromContext(ctx).Info("Error checking connection health", "err", err)
			return errConnHealthCheck
		}
		return nil
	}

	dsn, err := mariadbclient.BuildDSN(mdbOpts)
       ....
}
```
Thoughts?

lgohyipex: Done, exactly i was thinking, you read my mind :)

mmontes11: We both think in (probably) the best option 🤣

mmontes11 (approved):
This is fantastic @lgohyipex ! 🚀 🎖️ Thanks a lot for your contribution! ❤️

PR #243, opened 2023-09-28
feat(chart): liveness, readiness & startupProbes for webhook & operator

Nold360:
This feature adds probes to all pods.

currently hardcoded, do you which me to make them configureable? I didn't want to add them to the values file.


mmontes11 (approved):
Hey there @Nold360 ! 

I don't see the point of making them configurable either, as we are hardcoding the port as well.

Thanks for your contribution 🚀

Issue #226, opened 2023-09-07
[Bug] initcontainer spec collides with kustomize imagetagtransformer


Guzzi:
Hi @mmontes11 ,
I just tested 0.20 release with the initcontainers spec and run into a problem when using kustomize for overlay overrides.
If I add the imageTagTransformer in the overlay, I get an error and the manifests cannot be rendered by kustomize:

Failed to load target state: failed to generate manifest for source 1 of 1: rpc error: code = Unknown desc = `kustomize build .cgroup/ctraffic/overlay/azure-dev1` failed exit status 1: Error: wrong Node Kind for expected: ScalarNode was MappingNode: value: {pullPolicy: IfNotPresent repository: mariadb tag: 10.11.5}

https://kubectl.docs.kubernetes.io/references/kustomize/builtins/#_imagetagtransformer_

It seems, that kustomize interprets the image spec as standard kubernetes image definition and this has a different data structure than the one in the mdb operator (image: source:tag versus image:{properties})

initContainers:
    - image:
        repository: mariadb
        tag: "10.11.5"
        pullPolicy: IfNotPresent

To be compatible with kustomize imageTagTransformer either the object for image should comply to the standarf image object - or should have a different naming I suppose - what do you think?

mmontes11:
Hey there @Guzzi ! Thanks for bringing this up.

> To be compatible with kustomize imageTagTransformer either the object for image should comply to the standarf image object - or should have a different naming I suppose - what do you think?

That makes complete sense. We should align with the `Pod` specification to better integrate with tools like kustomize.

This will require breaking changes though, but I think it still worth it.

Guzzi:
Hi @mmontes11 ,
any chance to get this change released? Without it, I am unable to integrate mdb into kustomize which collides with the current object definitions. If you hesistate to apply breaking change, you could rename the mdb objects, so they do not collide with default kubernetes object names!?

mmontes11:
Hey there @Guzzi ! I will try to squeeze it in `v0.0.21`. 

> f you hesistate to apply breaking change, you could rename the mdb objects, so they do not collide with default kubernetes object names!?

That's actually a good suggestion, and actually the strategy followed by CNPG:
- https://cloudnative-pg.io/documentation/1.16/quickstart/

The downside I guess is that we will be losing integration with tooling like kustomize.

Guzzi:
[...]

PR #248, opened 2023-10-10
BREAKING CHANGE: standarize image to make it look like Pod image.

mmontes11:
Fixes https://github.com/mariadb-operator/mariadb-operator/issues/226

__BREAKING CHANGE__:  standarize image to make it look like Pod image.

## Migration guide
 
 Follow this steps If you have created a `MariaDB` resource using a `mariadb-operator` version < `0.0.21` and you are willing to migrate to >= `0.0.21`.
- `MariaDB` standalone migration
 ```diff
 - image:
-   repository: mariadb
-   tag: "11.0.3"
-   pullPolicy: IfNotPresent
+ image: mariadb:11.0.3
+ imagePullPolicy: IfNotPresent
 ```
 - `MariaDB` galera migration
  ```diff
 - image:
-   repository: mariadb
-   tag: "11.0.3"
-   pullPolicy: IfNotPresent
+ image: mariadb:11.0.3
+ imagePullPolicy: IfNotPresent
galera:
  agent:
-   image:
-     repository: ghcr.io/mariadb-operator/agent
-     tag: "v0.0.2"
-     pullPolicy: IfNotPresent
+   image: ghcr.io/mariadb-operator/agent:v0.0.2
+   imagePullPolicy: IfNotPresent
  initContainer:
-   image:
-     repository: ghcr.io/mariadb-operator/init
-     tag: "v0.0.5"
-     pullPolicy: IfNotPresent
+   image: ghcr.io/mariadb-operator/init:v0.0.5
+   imagePullPolicy: IfNotPresent
 ```


evanrich:
this causes a comparisonerror in argocd trying to upgrade, even when trying force/skip schema validation, etc.  will have to work on it to get it to upgrade

mmontes11:
Thanks for reporting this @evanrich and sorry for the inconvenience

> Failed to compare desired state to live state: failed to calculate diff: error calculating structured merge diff: error building typed value from config resource: .spec.image: expected map, got &{mariadb:10.11.5}

It seems like ArgoCD is expecting a map in the `spec.image`, could it be because the `MariaDB` CRD hasn't yet been updated?  We used to have an object there.

Not an ArgoCD expert, but for instance helm doesn't manage the CRDs upgrade, not sure if it is a similar situation here. Please check our migration guide in case you haven't already:
- https://github.com/mariadb-operator/mariadb-operator/blob/main/docs/UPGRADE_v0.0.21.md

And please let us know how it went your upgrade, we aim to integrate with ArgoCD as better as possible.

evanrich:
@mmontes11  thanks for the reploy.   I had tried following the upgrade guide, by first scaling down the deployment for the operator and webhook, then applying the CRD upgrade manually, then trying to let argocd deploy the updated chart.     I actually ran into a few more errors, such as the format of the mysql-exporter image, (it said something similar about  expecting a string, but getting some other value).  I tried screwing around with it for about an hour, including trying to force argocd to force deploy the update while disabling schema validation and having it replace the existing chart, which normally works, but just ran into a number of issues.    I ended up rolling back to 0.20.0 and everything was fine after that.  I'll probably end up just doing a removal of the entire mariadb operator at some point and then attempting a fresh install of 0.21.0, then re-importing my database (actually I have my PV set to retain, so it should just pick it back up).   

I feel like this is just some oddity in how argocd is trying to template the new chart over the old one or something, not a huge deal.   thanks!

mmontes11:
@evanrich thanks for looking into this and your in-depth investigation. 

> then applying the CRD upgrade manually

have you used `kubectl replace -f` ? Sometimes they refure to be updated. Also, once updated, have you checked the actual state of the image fiels i.e.  `spec.image`  and `spec.metrics.exporter.image`. By the time I was testing this, they got wiped out to `{}` which would indicate that the CRD was successfully updated for this case.

> then re-importing my database (actually I have my PV set to retain, so it should just pick it back up

that should work as well and probably more convenient. Let me know how it goes

> I feel like this is just some oddity in how argocd is trying to template the new chart over the old one or something, not a huge deal. thanks!

Sorry for the breaking change, it was done for good in order to align with the `Pod` spec. Our aim is to keep the ArgoCD experience (and GitOps in general) as better as possible, so feel free to file another issue with the problems you find along the way!

Issue #256, opened 2023-10-16
[Bug] Infinite loop updating NodePort services


kvaps:
**Describe the bug**

```
# kubectl get svc -n db-mariadb -w
NAME            TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)          AGE
may             ClusterIP   172.23.54.220    <none>        3306/TCP         16m
may-internal    ClusterIP   None             <none>        3306/TCP         16m
may-primary     NodePort    172.23.253.144   <none>        3306:32442/TCP   16m
may-secondary   NodePort    172.23.253.205   <none>        3306:31597/TCP   16m
may-primary     NodePort    172.23.253.144   <none>        3306:32648/TCP   16m
may-secondary   NodePort    172.23.253.205   <none>        3306:32686/TCP   16m
may-primary     NodePort    172.23.253.144   <none>        3306:32641/TCP   16m
may-secondary   NodePort    172.23.253.205   <none>        3306:32690/TCP   16m
may-primary     NodePort    172.23.253.144   <none>        3306:30184/TCP   16m
may-secondary   NodePort    172.23.253.205   <none>        3306:32200/TCP   16m
may-primary     NodePort    172.23.253.144   <none>        3306:30307/TCP   16m
may-secondary   NodePort    172.23.253.205   <none>        3306:32255/TCP   16m
may-primary     NodePort    172.23.253.144   <none>        3306:30279/TCP   16m
may-secondary   NodePort    172.23.253.205   <none>        3306:30817/TCP   16m
may-primary     NodePort    172.23.253.144   <none>        3306:31172/TCP   16m
```

**Expected behaviour**

Service will use stable nodePort

**Steps to reproduce the bug**

```
apiVersion: mariadb.mmontes.io/v1alpha1
kind: MariaDB
metadata:
  name: mariadb-may
spec:
  rootPasswordSecretKeyRef:
    name: mariadb-may
    key: root-password

  image: "mariadb:11.0.2"

  port: 3306

  replicas: 3
     
  replication:
    enabled: true
    #primary:
    #  podIndex: 0
    #  automaticFailover: true

  volumeClaimTemplate:
    resources:
      requests:
        storage: 100Gi
    accessModes:
      - ReadWriteOnce

  #service:
  #  type: NodePort
  primaryService:
    type: NodePort
    externalTrafficPolicy: Local
  secondaryService:
    type: NodePort
    externalTrafficPolicy: Local
```

**Additional context**

[...]

PR #258, opened 2023-10-20
Fix annotations and labels reconciliation for services

kvaps:
Now specified labels and annotations will be merged with the service, but not replaced.

This is possible fix for https://github.com/mariadb-operator/mariadb-operator/issues/256

Alternatively it might be better to use three-way-merge technics to determine managed and non-managed fields by mariadb-operator.

### DIFF
--- a/pkg/controller/service/controller.go
+++ b/pkg/controller/service/controller.go
@@ -36,7 +36,12 @@ func (r *ServiceReconciler) Reconcile(ctx context.Context, desiredSvc *corev1.Se
 	existingSvc.Spec.Ports = desiredSvc.Spec.Ports
 	existingSvc.Spec.Selector = desiredSvc.Spec.Selector
 	existingSvc.Spec.Type = desiredSvc.Spec.Type
-	existingSvc.Annotations = desiredSvc.Annotations
-	existingSvc.Labels = desiredSvc.Labels
+	for k, v := range desiredSvc.Annotations {
+		existingSvc.Annotations[k] = v
+	}
+	for k, v := range desiredSvc.Labels {
+		existingSvc.Labels[k] = v
+	}
+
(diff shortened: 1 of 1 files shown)


mmontes11 (approved):
Thanks a lot for your contribution and taking the time to debug this issue, very good catch 🦭 🙇🏻 

As you described in https://github.com/mariadb-operator/mariadb-operator/issues/256, it seems like mariadb-operator and metallb were clashing when reconciling the `Service`. The PR most likely will fix it I would say, as it will preserve the additional annotations introduced by metallb

> Alternatively it might be better to use three-way-merge technics to determine managed and non-managed fields by mariadb-operator.

I'm not too familiar with this tecnique, do you have some more info/example? I would agree that currently there's nothing in place to determine whether the field x should be managed by the operator or not

LGTM 🙏🏻

kvaps:
> I'm not too familiar with this tecnique, do you have some more info/example? I would agree that currently there's nothing in place to determine whether the field x should be managed by the operator or not

Simple, this is why the "last-applied" annotation was invented for.

During the apply operation you can encode applicable object into JSON string and save into specific annotation.
During the next apply operation you use this encoded object to calculate patch for live object.
Some operators use gzip+base64 for such annotations, but it is optional.

Ref into kubectl:
- documentation: https://kubernetes.io/docs/tasks/manage-kubernetes-objects/declarative-config/#how-apply-calculates-differences-and-merges-changes
- code: https://github.com/kubernetes/kubectl/blob/b5fe0f6e9c65ea95a2118746b7e04822255d76c2/pkg/util/apply.go#L67-L117

There is a potential problem right now: if a user shows some annotations on a mariadb object and then deletes them, these annotations will never be deleted from live object.

PR #259, opened 2023-10-20
Introduce AllocateLoadBalancerNodePorts option

kvaps:
I don't use external loadbalancer, so allocation node ports for my services is not necessary.
I would like to control this behavior by specifying `allocateLoadBalancerNodePorts: false` option.

see https://github.com/kubernetes/kubernetes/issues/69845 and https://github.com/kubernetes/kubernetes/pull/92744 for more details.


pkg/builder/service_builder.go
```diff
@@ -68,6 +68,9 @@ func (b *Builder) BuildService(mariadb *mariadbv1alpha1.MariaDB, key types.Names
 	if opts.SessionAffinity != nil {
 		svc.Spec.SessionAffinity = *opts.SessionAffinity
 	}
+	if opts.AllocateLoadBalancerNodePorts != nil {
+		svc.Spec.AllocateLoadBalancerNodePorts = *&opts.AllocateLoadBalancerNodePorts
```
mmontes11, line 72: As the linter suggests, this can be simplified to:
```suggestion
		svc.Spec.AllocateLoadBalancerNodePorts = opts.AllocateLoadBalancerNodePorts
```
Is there any reason for not doing so?

mmontes11 (changes requested):
Thanks a lot for your contribution 🦭 ✨ 

We should definitely provide a way to opt out from NodePort provision, like Kubernetes does.

LGTM, left a minor nitpick

mmontes11 (approved):

mmontes11:
LGTM 👍🏻 thanks for your contribution

Issue #11, opened 2022-11-03
Physical backups via `mariabackup` 


mmontes11:
MariaDB has its own backup command to perform backups, we should migrate to use it:
- https://mariadb.com/kb/en/mariabackup-overview/

Usage can be found here under the `Creating backups with Mariabackup` section:
- https://hub.docker.com/_/mariadb/

```bash
docker run --user mysql -v some-mariadb-socket:/var/run/mysqld -v some-mariadb-backup:/backup -v /my/own/datadir:/var/lib/mysql --rm mariadb:latest mariabackup --backup --target-dir=/backup
```

grooverdan:
There's still a role for the venerable mysqldump as a logical backup as its a good mitigator for the odd corruption bugs of database tables that mariabackup (and others like aria-hotcopy) physical backup programs don't mitigate. So I'd consider an addition rather than replacement.

mmontes11:
@grooverdan thanks for the clarification, it does sound great. We could extend our CRD so the user could choose the `type` of backup, defaulting to the logical one:
```yaml
apiVersion: database.mmontes.io/v1alpha1
kind: BackupMariaDB
metadata:
  name: backup
spec:
  type: (logical | physical)
  mariaDbRef:
    name: mariadb
  storage:
    persistentVolumeClaim:
      resources:
        requests:
          storage: 100Mi
      storageClassName: standard
      accessModes:
        - ReadWriteOnce
```

mmontes11:
Reopening, see:
- https://github.com/mariadb-operator/mariadb-operator/pull/97

mmontes11:
I've managed to restore a physical backup in a `MariaDB` instance with Galera. Steps followed:

- Provision a `MariaDB` instance with Galera
```yaml
apiVersion: mariadb.mmontes.io/v1alpha1
kind: MariaDB
metadata:
  name: mariadb-galera
spec:
  rootPasswordSecretKeyRef:
    name: mariadb
    key: root-password

  database: mariadb
  username: mariadb
  passwordSecretKeyRef:
    name: mariadb
    key: password

  image:
    repository: mariadb
    tag: "11.0.3"
    pullPolicy: IfNotPresent

  port: 3306

  replicas: 3
     
  galera:
    enabled: true

  volumeClaimTemplate:
    resources:
      requests:
        storage: 1Gi
    accessModes:
      - ReadWriteOnce
  volumes: 
    - name: mariabackup
      persistentVolumeClaim:
[...]

Issue #246, opened 2023-10-08
[Feature] Replication - Bootstrap replica from `mariabackup`


mmontes11:
**Describe the solution you'd like**
When spining up a new replica in a replication setup, take a physical backup in the primary and restore it in the new replica as described here:
- https://mariadb.com/kb/en/configuring-mariadb-replication-between-two-mariadb-galera-clusters/

It can happen that the primary binary logs get purged, making the new replica unable to bootstrap or just missing operations.

**Describe alternatives you've considered**
N/A

**Additional context**
Related on:
- https://github.com/mariadb-operator/mariadb-operator/issues/141

Depends on:
- https://github.com/mariadb-operator/mariadb-operator/issues/11

**Environment details**:
- Kubernetes version:
- mariadb-operator version: 
- Install method: helm, OLM or static manifests
- Install flavour: minimal, recommended or custom

kvaps:
Ideally cluster should be able to recover after executing:
```
FLUSH LOGS;
PURGE BINARY LOGS BEFORE NOW();
```
on master.

kvaps:
Just wanted to mention that running commands via exec mechanism in Kubernetes is not intended to be reliable mechanism.

However it can be simple replaced by spawning additional pod/job as described in https://github.com/mariadb-operator/mariadb-operator/pull/273#discussion_r1398254512

Issue #141, opened 2023-06-19
[Bug] Cannot add a semi-sync replica to existing cluster


fchiacchiaretta:
**Describe the bug**
I have a running semi-sync replica cluster with 1 master and 1 replica.
Trying to add a new replica fails since it cannot find binlogs required for replica
```
2023-06-19  9:47:12 12 [Note] Master connection name: 'mariadb-operator'  Master_info_file: 'master-mariadb@002doperator.info'  Relay_info_file: 'relay-log-mariadb@002doperator.info'
2023-06-19  9:47:12 12 [Note] 'CHANGE MASTER TO executed'. Previous state master_host='', master_port='3306', master_log_file='', master_log_pos='4'. New state master_host='mariadb-myns-0.mariadb-myns.myns.svc.cluster.local', master_port='3306', master_log_file='', master_log_pos='4'.
2023-06-19  9:47:12 12 [Note] Previous Using_Gtid=No. New Using_Gtid=Current_Pos
2023-06-19  9:47:12 13 [Note] Master 'mariadb-operator': Slave I/O thread: Start semi-sync replication to master 'repl@mariadb-myns-0.mariadb-myns.myns.svc.cluster.local:3306' in log '' at position 4
2023-06-19  9:47:12 14 [Note] Master 'mariadb-operator': Slave SQL thread initialized, starting replication in log 'FIRST' at position 4, relay log './mariadb-myns-relay-bin-mariadb@002doperator.000001' position: 4; GTID position ''
2023-06-19  9:47:12 13 [Note] Master 'mariadb-operator': Slave I/O thread: connected to master 'repl@mariadb-myns-0.mariadb-myns.myns.svc.cluster.local:3306',replication starts at GTID position ''
2023-06-19  9:47:12 13 [ERROR] Master 'mariadb-operator': Error reading packet from server: Could not find GTID state requested by slave in any binlog files. Probably the slave state is too old and required binlog files have been purged. (server_errno=1236)
2023-06-19  9:47:12 13 [ERROR] Master 'mariadb-operator': Slave I/O: Got fatal error 1236 from master when reading data from binary log: 'Could not find GTID state requested by slave in any binlog files. Probably the slave state is too old and required binlog files have been purged.', Internal MariaDB error code: 1236
[...]

Issue #6, opened 2022-08-19
Support for S3 storage in `BackupMariaDB` and `RestoreMariaDB`


mmontes11:
Be able to take and restore backups using AWS S3 as storage.

ox1depl:
its will support only AWS or any storage with S3 API, like azure blob, monio and others ?

mmontes11:
Hey @ox1depl !

The intention is supporting any S3 compatible API by levaraging this library:
- https://github.com/minio/minio-go

Let me know if you have other suggestions.

mmontes11:
At the moment, S3 backups can be achieved by installing the following CSI driver:
- https://github.com/yandex-cloud/k8s-csi-s3

and specify its `storageClassName` in the `Backup`/`Restore` resources:
```yaml
apiVersion: mariadb.mmontes.io/v1alpha1
kind: Backup
metadata:
  name: backup
spec:
  mariaDbRef:
    name: mariadb
  storage:
    persistentVolumeClaim:
      storageClassName: csi-s3
      resources:
        requests:
          storage: 100Mi
      accessModes:
        - ReadWriteOnce
  args:
    - --single-transaction
    - --all-databases
 ```
 The S3 support is a priority for the next releases though

PR #273, opened 2023-11-12
Add Support to mariadbbackup

harunkucuk5:
Closes https://github.com/mariadb-operator/mariadb-operator/issues/11

This PR adds mariadbbackup support.

~Note: scheduled backups are not supported for now as the physical backup requires to be triggered on the database pod and when cronjob starts there is nothing that can trigger backup on the database pod. But for the on-demand backups mariadb-backup reconciler triggers the backup on the database pod~

The new logic allows scheduled mariadbbackups. Thanks @kvaps.

logs from backup
```bash
 💾 Taking physical backup                                                                                                                                                                                                                                           
[00] 2023-12-26 20:22:06 Connecting to MariaDB server host: mariadb-repl-primary, user: root, password: set, port: not set, socket: /run/mysqld/mysqld.sock                                                                                                         
[00] 2023-12-26 20:22:06 Using server version 11.0.3-MariaDB-1:11.0.3+maria~ubu2204-log                                                                                                                                                                             
mariadb-backup based on MariaDB server 11.0.3-MariaDB debian-linux-gnu (aarch64)
...

[00] 2023-12-26 20:38:29         ...done                                                                                                                                                                                                                            
[00] 2023-12-26 20:38:29 Redo log (from LSN 443413019 to 443598086) was copied.                                                                                                                                                                                     
[00] 2023-12-26 20:38:29 completed OK!                                                                                                                                                                                                                              
📜 Backup completed                                                                                                                                                                                                                                                 
🧹 Cleaning up old backups
📜 Backup history
...
87M    /backup/mariabackup-2023-12-25T18:32:39Z                                                                                                                                                                                                                     
87M    /backup/mariabackup-2023-12-26T19:47:43Z                                                                                                                                                                                                                     
87M    /backup/mariabackup-2023-12-26T20:22:06Z


```

logs from restore

```bash
  Checking if a restore already performed...                                                                                                                                                                                                                          
 Cleaning current data folder...                                                                                                                                                                                                                                     
 Creating a temporary data folder...                                                                                                                                                                                                                                 
 Moving backup to the temporary data folder...                                                                                                                                                                                                                       
 sending incremental file list                                                                                                                                                                                                                                       
 ./                                                                                                                                                                                                                                                                  
 aria_log.00000001                                                                                                                                                                                                                                                   
 aria_log_control                                                                                                                                                                                                                                                    
 backup-my.cnf
 ...
 sent 90,102,988 bytes  received 4,233 bytes  25,744,920.29 bytes/sec                                                                                                                                                                                                
 total size is 90,064,114  speedup is 1.00                                                                                                                                                                                                                           
 Preparing backup...                                                                                                                                                                                                                                                 
 mariadb-backup based on MariaDB server 11.0.3-MariaDB debian-linux-gnu (aarch64)                                                                                                                                                                                    
 [00] 2023-12-26 20:25:30 cd to /var/lib/mysql/mariabackup/                                                                                                                                                                                                          
 [00] 2023-12-26 20:25:30 open files limit requested 0, set to 1048576                                                                                                                                                                                               
 [00] 2023-12-26 20:25:30 Renaming ./mysql/gtid_slave_pos.new to ./mysql/gtid_slave_pos.ibd
...
 [00] 2023-12-26 20:25:31 Last binlog file ./mariadb-repl-bin.000003, position 851                                                                                                                                                                                   
 [00] 2023-12-26 20:25:31 completed OK!                                                                                                                                                                                                                              
 Moving restored backup to the data directory...                                                                                                                                                                                                                     
 Removing temporary folder...                                                                                                                                                                                                                                        
 Creating a flag file to avoid multiple restoration after restart...                                                                                                                                                                                                 
 Changing ownership of the files to mysql...
...

```

![image](https://github.com/mariadb-operator/mariadb-operator/assets/8478133/5b2c0041-df22-432a-a143-5e3a4700723c)

![image](https://github.com/mariadb-operator/mariadb-operator/assets/8478133/d68e593e-120f-4964-96ff-08c33d20e5e0)

![image](https://github.com/mariadb-operator/mariadb-operator/assets/8478133/65898386-737b-46f1-ae26-c2748ec8adbc)


![image](https://github.com/mariadb-operator/mariadb-operator/assets/8478133/67aa78c5-e3e7-4a70-b27b-0d7d516a37c9)


controllers/mariabackup_controller.go
```diff
@@ -0,0 +1,218 @@
...
+			SubResource("exec").
+			VersionedParams(&corev1.PodExecOptions{
+				Container: "mariadb",
+				Command:   cmd,
+				Stdin:     true,
+				Stdout:    true,
+				Stderr:    true,
+			}, runtime.NewParameterCodec(r.Scheme))
+
+		exec, err := remotecommand.NewSPDYExecutor(r.RESTConfig, "POST", execReq.URL())
```
mmontes11, line 170: Love this ! 💯 ❤️ 💯 ❤️ 💯 ❤️ 💯 ❤️ 

Thank you very much much! This gives a lot of flexibility, haven't thought about this before!

controllers/mariabackup_controller.go
```diff
@@ -0,0 +1,218 @@
...
+			break
+		} else if pod.Status.Phase == "Succeeded" {
+			podIP = ""
+			break
+		}
+	}
+
+	if podIP != "" {
+		// Start mariadb-backup on the primary pod
+		// TODO: make the backup source configurable primary or replicas
```
mmontes11, line 148: To get a Pod replica IP, you may want to look into this:
https://github.com/mariadb-operator/mariadb-operator/blob/78a26624e51aec394bfd679abc8ad261d5a1f403/pkg/health/health.go#L78

controllers/mariabackup_controller.go
```diff
@@ -0,0 +1,218 @@
...
+		// TODO: add s3 configuration parameters to get backup to s3 directly
```
mmontes11, line 149: S3 backup storage is in our roadmap already and something that our community is looking forward to have: https://github.com/mariadb-operator/mariadb-operator/issues/6

If you feel like the PR gets too big feel free to open a new one, or we can take care of it as a separate issue, I leave it up to you. We probably would need to generalize it, not only for physical but also for logical backups.

mmontes11:
Hey there @harunkucuk5 ! Huge kudos for starting working on this, looks promising. It will be a massive contribution for the project  🎖️ 💯 ❤️ 💯 ❤️ 🎖️ 

> Note: scheduled backups are not supported for now as the physical backup requires to be triggered on the database pod and when cronjob starts there is nothing that can trigger backup on the database pod. But for the on-demand backups mariadb-backup reconciler triggers the backup on the database pod

Regarding the scheduled backups, I would say don't stress too much, we can take care of it as a separate issue or even implement it myself later. Having `mariabackup` support is a huge step forward.

I think this PR is attempting to automate [this suggestion](https://github.com/mariadb-operator/mariadb-operator/issues/141#issuecomment-1804760231) from @kvaps, is that correct? Kudos to him as well for this fantastic guide 🏅 

I am curious about how are you planning to support the physical backup restoration? I think you are in a point where you have the physical backup already in a PVC, how are you planning to ship it to a replica?
- Send the backup through streaming to a replica
- Mounting the `mariabackup` PVC

Once the replica has access to the physical backup, how do you plan to actually copy it to `/var/lib/mysql`:
- By executing a remote command and follow the @kvaps approach?
- By introducing and extra initContainer and restart the `StatefulSet`

Having an initial support for `mariabackup` would be massive, but in the long term we might want to use it to fix replication scenarios, see: https://github.com/mariadb-operator/mariadb-operator/issues/246#issuecomment-1804727359

Again, thanks a lot for working on this, you have my full support. Let me know if you need a hand for the implementation! 🙏🏻

mmontes11 (commented):
Hey there @harunkucuk5 ! Huge kudos for starting working on this, looks promising. It will be a massive contribution for the project  🎖️ 💯 ❤️ 💯 ❤️ 🎖️ 

> Note: scheduled backups are not supported for now as the physical backup requires to be triggered on the database pod and when cronjob starts there is nothing that can trigger backup on the database pod. But for the on-demand backups mariadb-backup reconciler triggers the backup on the database pod

Regarding the scheduled backups, I would say don't stress too much, we can take care of it as a separate issue or even implement it myself later. Having `mariabackup` support is a huge step forward.

I think this PR is attempting to automate [this suggestion](https://github.com/mariadb-operator/mariadb-operator/issues/141#issuecomment-1804760231) from @kvaps, is that correct? Kudos to him as well for this fantastic guide 🏅 

I am curious about how are you planning to support the physical backup restoration? I think you are in a point where you have the physical backup already in a PVC, how are you planning to ship it to a replica?
- Send the backup through streaming to a replica
- Mounting the `mariabackup` PVC

Once the replica has access to the physical backup, how do you plan to actually copy it to `/var/lib/mysql`:
- By executing a remote command and follow the @kvaps approach?
- By introducing and extra initContainer and restart the `StatefulSet`

Having an initial support for `mariabackup` would be massive, but in the long term we might want to use it to fix replication scenarios, see: https://github.com/mariadb-operator/mariadb-operator/issues/246#issuecomment-1804727359

Left a couple of suggestions in the PR!

Again, thanks a lot for working on this, you have my full support. Let me know if you need a hand for the implementation! 🙏🏻

controllers/mariabackup_controller.go
```diff
@@ -0,0 +1,218 @@
...
+	podList := &corev1.PodList{}
+
+	// Create a label selector to filter pods with a specific label
+	labelSelector := labels.SelectorFromSet(labels.Set(map[string]string{
+		"backup-ref": backup.Name,
+	}))
+
+	// Create a ListOptions with the label selector
+	listOptions := &client.ListOptions{
+		LabelSelector: labelSelector,
```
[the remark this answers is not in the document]
harunkucuk5: Thanks @mmontes11, I will fix this

controllers/mariabackup_controller.go
harunkucuk5: well, my initial idea was using the same logic that we have in xtrabackup + xbcloud [here](https://docs.percona.com/percona-xtrabackup/8.0/xbcloud-swift.html) that percona has built-in support to get backup directly to s3 and also restore as well. However, it seems there is no such utility like mariadbcloud utility for mariadbbackup. I need to think about this a bit more.

Just as an idea, I could use xtrabackup [image](https://hub.docker.com/r/percona/percona-xtrabackup) in the job pod and it would receive mariabackup stream through socat and upload to bucket with xbcloud like mariadbbackup + xbcloud. As mariadbbackup is ported from xtrabackup hopefully they are compatible.

harunkucuk5:
> I think this PR is attempting to automate [this suggestion](https://github.com/mariadb-operator/mariadb-operator/issues/141#issuecomment-1804760231) from @kvaps, is that correct? Kudos to him as well for this fantastic guide 🏅
> 
> I am curious about how are you planning to support the physical backup restoration? I think you are in a point where you have the physical backup already in a PVC, how are you planning to ship it to a replica?
> 
> * Send the backup through streaming to a replica
> * Mounting the `mariabackup` PVC
> 
> Once the replica has access to the physical backup, how do you plan to actually copy it to `/var/lib/mysql`:
> 
> * By executing a remote command and follow the @kvaps approach?
> * By introducing and extra initContainer and restart the `StatefulSet`

Hey @mmontes11, thanks for your great review! Regarding your questions, this will be a backup and restore solution for now, so it won't handle scaling up an existing cluster and copying data from master to replica for now, but we can implement this as well later on.

So, this PR creates persistent volume if it is not already there and puts backups to that volume. When it comes to restoring I am planning to introduce the init container which will be responsible for the MariaDB backup preparation and copying it to the data folder then it will complete its task. When the actual container starts to run, it will have the data already in the data folder and just start running.

controllers/mariabackup_controller.go
kvaps: Not sure about this approach,

From my experience, I can say that exec mechanism depends on a large number of components (primarily apiserver, CRI and kubelet) and if at least one stops working, the execution of the command will be interrupted with an unprediucted result. No logs are collected.

Please consider spawning a new pod or job with the same volume and podAffinity to the target pod.
It should normally work with ReadWriteOnce volumes as well, since the local filesystem can be simple shared with multiple pods across the single node.

kvaps: In order not to be unfounded, here is a working proposal:

```
MASTER=mydb-1

image=$(kubectl get pod -o yaml "$MASTER" -o json | jq -rc '.spec.containers[] | select(.name=="mariadb").image')
volumeMounts=$(kubectl get pod -o yaml "$MASTER" -o json | jq -rc '.spec.containers[] | select(.name=="mariadb").volumeMounts')
env=$(kubectl get pod -o yaml "$MASTER" -o json | jq -rc '.spec.containers[] | select(.name=="mariadb").env')
volumes=$(kubectl get pod -o yaml "$MASTER" -o json | jq -rc '.spec.volumes')
ip=$(kubectl get pod -o yaml "$MASTER" -o json | jq -r '.status.podIP')
cmd="[\"mariadb-backup\", \"--backup\", \"--user=root\", \"--password=\$(MARIADB_ROOT_PASSWORD)\", \"--compress\", \"--target-dir=/tmp\", \"--protocol=tcp\", \"--host=$ip\"]"

overrides="$(
cat <<EOT
{
  "spec": {
    "containers": [
      {
        "image": "$image",
        "name": "mariadb-backup",
        "stdin": true,
        "stdinOnce": true,
        "tty": true,
        "command": $cmd,
        "env": $env,
        "volumeMounts": $volumeMounts
      }
    ],
    "tolerations": [{ "operator": "Exists" }],
    "volumes": $volumes
  }
}
EOT
)"

kubectl run --image "$image" --restart=Never --overrides="$overrides" -ti mariadb-backup
```

Of course you don't need all those `stdin` and `tty` options when running job from operator

api/v1alpha1/base_types.go
```diff
@@ -279,6 +279,14 @@ type RestoreSource struct {
 	// +optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	TargetRecoveryTime *metav1.Time `json:"targetRecoveryTime,omitempty" webhook:"inmutable"`
+	// FileName is the file within the source to be restored.
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	FileName *string `json:"fileName,omitempty" webhook:"inmutableinit"`
```
mmontes11, line 285: Is there any way we could integrate the physical backup feature with the PITR? This is quite recent, see the docs:
- https://github.com/mariadb-operator/mariadb-operator/blob/main/docs/BACKUP.md

If that's not the case, could we expand the `FileName` explanation a bit further to indicate that this is supposed to be used only in physical backups.

We don't necessarily have to do it in this PR, I can take a look as part of a different PR.

api/v1alpha1/base_types.go
```diff
@@ -279,6 +279,14 @@ type RestoreSource struct {
...
+	// Type is the type of backup. (mariabackup or mariadump)
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	Type *string `json:"type,omitempty" webhook:"inmutableinit"`
```
mmontes11, line 289: Maybe we can leave as `string` and default it here?
https://github.com/mariadb-operator/mariadb-operator/blob/0179c43be88b483168cee3d54939d743d8dabcc5/api/v1alpha1/base_types.go#L295

This would reduce the risk of nil pointers, which by the way I noticed we are getting some in the CI tests:
- https://github.com/mariadb-operator/mariadb-operator/actions/runs/7332592001/job/19966968808?pr=273

In this line specifically:
https://github.com/mariadb-operator/mariadb-operator/pull/273/files#diff-23dfecbfdd15c2c0a28d94dfb1bae6d933f91a37124325caf0257ce5aa9cfcafR106

Also, what about creating an enum and document the values like we do here:
https://github.com/mariadb-operator/mariadb-operator/blob/0179c43be88b483168cee3d54939d743d8dabcc5/api/v1alpha1/mariadb_galera_types.go#L17

api/v1alpha1/mariabackup_types.go
```diff
@@ -0,0 +1,121 @@
...
+
+import (
+	"errors"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type MariaBackupStorage struct {
```
mmontes11, line 27: I would love to have S3 storage backups here, we recently introduced this for logical backups:
https://github.com/mariadb-operator/mariadb-operator/blob/main/docs/BACKUP.md

Again, this can be done in a different PR and I can look into it.

controller/mariadb_controller.go
```diff
@@ -316,7 +316,7 @@ func (r *MariaDBReconciler) reconcileGalera(ctx context.Context, mariadb *mariad
 }
 
 func (r *MariaDBReconciler) reconcileRestore(ctx context.Context, mariadb *mariadbv1alpha1.MariaDB) (ctrl.Result, error) {
-	if mariadb.Spec.BootstrapFrom == nil {
+	if mariadb.Spec.BootstrapFrom == nil || *mariadb.Spec.BootstrapFrom.Type == "mariabackup" {
```
mmontes11, line 319: See my previous comment about `Type` nullability, same applies here.

pkg/builder/statefulset_container_builder.go
```diff
@@ -102,6 +103,49 @@ func (b *Builder) buildGaleraAgentContainer(mariadb *mariadbv1alpha1.MariaDB) co
...
 func buildStsInitContainers(mariadb *mariadbv1alpha1.MariaDB) []corev1.Container {
 	initContainers := []corev1.Container{}
+	if mariadb.Spec.BootstrapFrom != nil && *mariadb.Spec.BootstrapFrom.Type == "mariabackup" && mariadb.Spec.BootstrapFrom.BackupRef != nil {
+
+		cmdsCheck := []string{
+			"echo 'Checking if a restore already performed...'",
+			"[ -e /var/lib/mysql/restore_success ] && { echo 'No restore required, exiting...'; exit 0; }",
+		}
+
+		cmdsRestore := []string{
```
mmontes11, line 113: I believe this is inspired in a previous @kvaps suggestion 👍🏻 :
https://github.com/mariadb-operator/mariadb-operator/pull/273#discussion_r1398243571
Would love to have some input 🙏🏻

mmontes11 (changes requested):
Thanks a lot for the effort @harunkucuk5 ! This is looking better, very much appreciated 🙏🏻 

The CI tests are failing, could you please take a look? I think we are de-referencing the `Type` field, which in some cases is nil. As I said in the comments, I think it will be worth changing it to a `string` and default it in the `SetDefaults` method to a logical backup to be backwards compatible.

Regarding the `Check Diff` step in the CI, just run `make gen` to re-generate code, CRDs and all the required stuff.

pkg/builder/statefulset_container_builder.go
harunkucuk5: this is the part the init container performs a restore from mariadbbackup and once the restore is complete it creates a file to avoid another restore once the pod gets restarted.

So after every restart it checks if this file exists, if this file exists then that means the restore is already performed and just skips restore and then the actual container comes up.

```bash
restore-init Checking if a restore is already performed...                                                                                                                                                                                                     
restore-init No restore required, exiting...
```

mmontes11: Yes, I think it looks good. As an alternative we could pre-populate the PVCs with a Job, but I think this is much simpler.

This would probably work when upscaling the cluster and creating new replicas, as long as the backup is in the right location.

api/v1alpha1/base_types.go
harunkucuk5: Hey @mmontes11, since mariadbbackup is a physical backup solution, it does not support PITR. But here I think we use PITR in the wrong sense. PITR means restoring the database to the exact time, which can be achieved using binary logs. But here we just use the time to find the closest backup file and import the entire backup. The same can be done for the mariadbbackup but we should not call this PITR. this may lead to misunderstanding

controller/restore_controller.go
```diff
@@ -43,6 +43,12 @@ func (r *RestoreReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	}
 
 	mariaDb, err := r.RefResolver.MariaDB(ctx, &restore.Spec.MariaDBRef, restore.Namespace)
+
+	if mariaDb.Spec.BootstrapFrom != nil && *mariaDb.Spec.BootstrapFrom.Type == "mariabackup" {
+		// skip restore reconciler for the mariadbbackup restore
```
mmontes11, line 48: What about about deleting the `/var/lib/mysql/restore_success` from the Job and restoring the backup? Afterwards the `Pod` should be restarted.

This could be helpful is situations where the replicas go down and need a physical backup to be restored. For example:
- https://github.com/mariadb-operator/mariadb-operator/issues/141

pkg/builder/statefulset_container_builder.go
harunkucuk5: yes, this works, and it starts replicating from the master as long as the master has not purged the binary logs

> This would probably work when upscaling the cluster and creating new replicas, as long as the backup is in the right location.

mmontes11 (commented):
Left a few more suggestions

pkg/builder/statefulset_container_builder.go
mmontes11: > purged the binary logs

exactly, we have some issues open about this:
https://github.com/mariadb-operator/mariadb-operator/issues/141

api/v1alpha1/base_types.go
mmontes11: right, understood. PITR implies the entire state of the database not just the data (logical state). I think we should rename it to "target recovery time" ?

controllers/mariabackup_controller.go
mmontes11: Regarding S3, we have a builtin mechanism based of the `mariadb-operator` image:
- https://github.com/mariadb-operator/mariadb-operator/tree/main/cmd/backup
So we don't need to rely on xbcloud

api/v1alpha1/base_types.go
mmontes11: Thanks a lot for the suggestion, documentation updated accordingly:
https://github.com/mariadb-operator/mariadb-operator/blob/main/docs/BACKUP.md

pasztorl:
@harunkucuk5 sorry, may I ask an update about this? mariabackup perhaps the most wanted awesome feature

mmontes11:
@pasztorl I encourage you to take a look at this:
- https://github.com/mariadb-operator/mariadb-operator/issues/507

Following a similar approach as @harunkucuk5  we are intentding to refactor this and support PITR based on mariabackup + binary logs.

pasztorl:
> @pasztorl I encourage you to take a look at this:
> 
> * [[Feature] Point In Time Recovery (PITR) #507](https://github.com/mariadb-operator/mariadb-operator/issues/507)
> 
> Following a similar approach as @harunkucuk5, we are intentding to refactor this and support PITR based on mariabackup + binary logs.

sounds even better :)

harunkucuk5:
@mmontes11 , @pasztorl, sorry that I could not find time to focus on this again.

mmontes11:
@harunkucuk5 no prob! I will use and refactor these changes to implement [PITR](https://github.com/mariadb-operator/mariadb-operator/issues/507) and add you as a co-author if that's ok

mmontes11:
This has been supported via this PR: https://github.com/mariadb-operator/mariadb-operator/pull/1266

Thanks for raising the initial changes @harunkucuk5 ! You have been added as co-author.

PR #275, opened 2023-11-14
Bump agent version

matthieugusmini:
Upgraded the agent version to v0.0.4-beta1.
See https://github.com/mariadb-operator/agent/pull/5#pullrequestreview-1727838242


matthieugusmini:
I replaced the `agent` module import to my fork to check if the test passed. Will remove the `replace` directive once we tag a new version 👍🏼

PR #276, opened 2023-11-16
feat: add labels to job pods

Trojan295:
This PR adds `.spec.labels` field into the `Backup`, `Restore` and `SqlJob` CRDs. These labels are set on the `CronJob`, `Job` and `Pod` resources created by the controllers.

Currently, it's not possible to add custom labels to the job pods, which makes it hard to use with NetworkPolicies.


Trojan295:
@mmontes11 could you take a look at this?

mmontes11:
Hey there @Trojan295 ! Thanks for raising this PR, I completely agree that setting up custom labels would definetly be helpful to configure network policies.

We currently support adding user provided metadata (i.e. labels and annotations) via `spec.inheritMetadata`, like in this example:
https://github.com/mariadb-operator/mariadb-operator/blob/7bf25741e00ee33e2319a78ef3101152e00906a1/examples/manifests/mariadb_v1alpha1_mariadb_metadata.yaml#L7

If you create a `Backup`, `Restore` or `SqlJob` on a `MariaDB` resource which has `spec.inheritMetadata` set, the child `Job` resources will also inherit this metadata, see:
https://github.com/mariadb-operator/mariadb-operator/blob/7bf25741e00ee33e2319a78ef3101152e00906a1/pkg/builder/batch_builder.go#L36
https://github.com/mariadb-operator/mariadb-operator/blob/7bf25741e00ee33e2319a78ef3101152e00906a1/pkg/builder/metadata/metadata.go#L28

Closing! I believe this should cover your case, feel free to reopen otherwise.

Thanks a lot for the PR anyway, I really appreciate your effort! 🙏🏻

Trojan295:
Oh, thanks, I did not see this field to inherit the metadata.

Issue #291, opened 2023-12-11
[Bug] Failover breaks replication if a super privileged user writes data during failover


harunkucuk5:
**Describe the bug**

We make the active primary read-only before failing it over. However, this is not enough to keep the data safe if there is a user that has SUPER privileges. So if a user has the SUPER privilege it can write to the database even if the database is read-only. So in that case when we mark the active primary read-only the current primary still accepts write requests from these users and when the switchover is complete the new master will be missing these data and the gtid position will be wrong when connecting the old primary as a replica to the new primary and the replication stream gets broken with the following error or another conflict

```bash
Last_SQL_Errno: 1062
                Last_SQL_Error: Could not execute Write_rows_v1 event on table test.users; Duplicate entry '15799' for key 'PRIMARY', Error_code: 1062; handler error HA_ERR_FOUND_DUPP_KEY; the event's master log mariadb-cluster-bin.000001, end_log_pos 192
```

**Is the bug related to Galera?**
No, it is related to the traditional MariaDB replication, not to Galera.

- [x] I have checked the [Galera documentation](https://github.com/mariadb-operator/mariadb-operator/blob/main/docs/GALERA.md)
- [x] I have checked the [Galera troubleshooting guide](https://github.com/mariadb-operator/mariadb-operator/blob/main/docs/GALERA.md#troubleshooting)

**Expected behavior**

When failover is started the active primary must not accept any write transaction from any user to keep the data unchanged. This can be done with the following command that described [here](https://mariadb.com/kb/en/flush/#purpose-of-flush-tables-with-read-lock:~:text=Purpose%20of%20FLUSH%20TABLES%20WITH%20READ%20LOCK%C2%B6)

```bash
flush tables with read lock;
```

**Steps to reproduce the bug**

Create a basic table

```bash
 CREATE TABLE `users` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `username` varchar(50) NOT NULL,
  `email` varchar(100) NOT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB AUTO_INCREMENT=15802 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci
```

Create a simple bash script to write this table,

```bash

#!/bin/bash

while true; do
    # Generate random values for username and email
    username=$(head /dev/urandom | tr -dc A-Za-z0-9 | head -c 8)
    email="${username}@example.com"

[...]

PR #311, opened 2023-12-19
Lock primary during replication failover

mmontes11:
Related to:
- https://github.com/mariadb-operator/mariadb-operator/issues/291

### DIFF
--- a/pkg/controller/replication/switchover.go
+++ b/pkg/controller/replication/switchover.go
@@ -44,13 +44,11 @@ func (r *ReplicationReconciler) reconcileSwitchover(ctx context.Context, req *re
 		return fmt.Errorf("error patching MariaDB status: %v", err)
 	}
 
-	err, close := r.lockPrimaryWithReadLock(ctx, req.mariadb, req.clientSet, logger)
-	if err != nil {
-		return fmt.Errorf("error locking primary: %v", err)
-	}
-	defer close()
-
 	phases := []switchoverPhase{
+		{
+			name:      "Lock primary with read lock",
+			reconcile: r.lockPrimaryWithReadLock,
+		},
 		{
 			name:      "Set read_only in primary",
 			reconcile: r.setPrimaryReadOnly,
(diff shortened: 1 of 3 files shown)


pkg/sql/sql.go
```diff
@@ -352,6 +352,14 @@ func (c *Client) SetSystemVariables(ctx context.Context, keyVal map[string]strin
 	return nil
 }
 
+func (c *Client) LockTablesWithReadLock(ctx context.Context) error {
+	return c.Exec(ctx, "FLUSH TABLES WITH READ LOCK;")
```
harunkucuk5, line 356: my concern here is that once this connection is terminated it unlocks the tables automatically. So we should make sure this connection is active until the failover is complete.

running the following command locks and unlocks tables automatically,

```bash
mariadb -uroot -p${MARIADB_ROOT_PASSWORD} -e 'FLUSH TABLES WITH READ LOCK'
```

pkg/controller/replication/switchover.go
```diff
@@ -46,8 +46,8 @@ func (r *ReplicationReconciler) reconcileSwitchover(ctx context.Context, req *re
 
 	phases := []switchoverPhase{
 		{
-			name:      "Set read_only in current primary",
```
harunkucuk5, line 49: I think we should keep this phase as well, so we mark the current primary read-only and it will stay read-only after the failover as it will become a replica.

pkg/sql/sql.go
mmontes11: That's right, I completely forgot ththat the tables are unlocked after the connection is closed. In our case, the connections last one reconciliation cycle, so it should be fine. See:

https://github.com/mariadb-operator/mariadb-operator/blob/0fa82f1aa5a55e6a0eea2625482519cef7ab1ec3/pkg/controller/replication/controller.go#L99

https://github.com/mariadb-operator/mariadb-operator/blob/0fa82f1aa5a55e6a0eea2625482519cef7ab1ec3/pkg/controller/replication/controller.go#L123

I will delete the unlock method, as tables are implicitly unlocked 👍🏻

pkg/controller/replication/switchover.go
mmontes11: What would be the benefit of setting it here? 

The current primary has already been locked in the previous step, so no new writes will happen in this instance. Also, as part or the primary and replica reconciliation we always set `read_only` idempotently, so I believe it should cover this case?


Primary:
https://github.com/mariadb-operator/mariadb-operator/blob/b02782032fcc9e459ef53ae2c51d28c014fd8468/pkg/controller/replication/config.go#L50

Replica:
https://github.com/mariadb-operator/mariadb-operator/blob/b02782032fcc9e459ef53ae2c51d28c014fd8468/pkg/controller/replication/config.go#L75

Maybe I am missing something?

pkg/sql/sql.go
mmontes11: Hmmm, maybe the unlock is required:
- https://github.com/mariadb-operator/mariadb-operator/actions/runs/7273373846/job/19817219567?pr=311

Will take a closer look later

pkg/controller/replication/switchover.go
harunkucuk5: Well, my point was here that as you know when the connection is closed it unlocks tables automatically. As you already said it should be fine. But if somehow this connection is closed in the middle of the failover process for some reason like network etc. having the old primary as read_only would be safer than the writeable one

> In our case, the connections last one reconciliation cycle, so it should be fine

mmontes11: > But if somehow this connection is closed in the middle of the failover process for some reason like network etc. having the old primary as read_only would be safer than the writeable one

Great, I think this is a valid point, I will add and extra step for this 👍🏻

pkg/sql/sql.go
mmontes11: Second attempt, this time explicitly unlocking the primary:
- https://github.com/mariadb-operator/mariadb-operator/actions/runs/7297147476/job/19885960383?pr=311

mmontes11: Judging by the logs, it seems like we need to unlock the old primary before coverting it into a replica.

harunkucuk5: right, it seems like that. I think this is safe to unlock tables at that phase as it is the last step and the primary service should be using the new primary already

```bash
Error 1192: Can't execute the given command because you have active locked tables or an active transaction
```

mmontes11: Third time is the charm:
https://github.com/mariadb-operator/mariadb-operator/actions/runs/7297603999/job/19887135274?pr=311

mmontes11:
@harunkucuk5 ready to re-review! LGTM from my side

harunkucuk5 (approved):
LGTM, thanks for your quick implementation :thumbsup:

mmontes11:
@harunkucuk5 thanks a lot for your feedback reviewing this PR 🙏🏻 

This will be released in `v0.0.24`, stay tuned 🚀

PR #331, opened 2024-01-08
Add an option for ephemeral storage

### LINKED ISSUE #322: [Feature] Allow non-pv MariaDB environment

as of 0.24.0, `MariaDB` requires the `volumeClaimTemplate` field. (see https://github.com/mariadb-operator/mariadb-operator/blob/v0.0.24/examples/manifests/mariadb_v1alpha1_mariadb_minimal.yaml)

In some scenarios such as running inside a CI environment or test environment, non-pv MariaDB instance is much easier to handle and can be more flexible.

`volumeClaimTemplate` to be optional

**Environment details**:
- Kubernetes version: [Version number]
- Kubernetes distribution: [Vanilla, EKS, GKE, AKS, Rancher, OpenShift, k3s, KIND...]
- mariadb-operator version: 0.24.0
- Install method: helm
- Install flavor: recommended

**Additional context**

Hoega:
Closes #322 

Makes volumeClaimTemplate parameter optional with the addition of an ephemeralStorage option.

```
kind: MariaDB
metadata:
  name: mariadb-min
spec:
  ephemeralStorage: true
```

If ephemeralStorage=true, the volume for mariaDB storage will be an emptyDir:
```
  Volumes:
   config:
    Type:       EmptyDir (a temporary directory that shares a pod's lifetime)
    Medium:     
    SizeLimit:  <unset>
   storage:
    Type:       EmptyDir (a temporary directory that shares a pod's lifetime)
    Medium:     
    SizeLimit:  <unset>
```

Add a webhook validation to make sure either ephemeralStorage or volumeClaimTemplate is defined but not both.

- When neither is set: 
`Error from server (Forbidden): error when creating "examples/manifests/mariadb_v1alpha1_mariadb_minimal.yaml": admission webhook "vmariadb.kb.io" denied the request: spec.volumeClaimTemplate: Invalid value: v1alpha1.VolumeClaimTemplate{PersistentVolumeClaimSpec:v1.PersistentVolumeClaimSpec{AccessModes:[]v1.PersistentVolumeAccessMode(nil), Selector:(*v1.LabelSelector)(nil), Resources:v1.ResourceRequirements{Limits:v1.ResourceList(nil), Requests:v1.ResourceList(nil), Claims:[]v1.ResourceClaim(nil)}, VolumeName:"", StorageClassName:(*string)(nil), VolumeMode:(*v1.PersistentVolumeMode)(nil), DataSource:(*v1.TypedLocalObjectReference)(nil), DataSourceRef:(*v1.TypedObjectReference)(nil)}, Labels:map[string]string(nil), Annotations:map[string]string(nil)}: 'spec.ephemeralStorage' or 'spec.volumeClaimTemplate' must be defined`

- When both are set:
`Error from server (Forbidden): error when creating "examples/manifests/mariadb_v1alpha1_mariadb_minimal.yaml": admission webhook "vmariadb.kb.io" denied the request: spec.ephemeralStorage: Invalid value: true: 'spec.ephemeralStorage' must be disabled when 'spec.volumeClaimTemplate' is specified`


Hoega:
Still need to fix unit tests

api/v1alpha1/mariadb_types.go
```diff
@@ -238,10 +238,14 @@ type MariaDBSpec struct {
...
+	// EphemeralStorage indicates whether to use ephemeral storage for the instances.
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	EphemeralStorage *bool `json:"ephemeralStorage,omitempty" webhook:"inmutable"`
 	// VolumeClaimTemplate provides a template to define the Pod PVCs.
-	// +kubebuilder:validation:Required
+	// +optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
-	VolumeClaimTemplate VolumeClaimTemplate `json:"volumeClaimTemplate" webhook:"inmutable"`
+	VolumeClaimTemplate *VolumeClaimTemplate `json:"volumeClaimTemplate" webhook:"inmutable"`
```
mmontes11, line 248: You can remove the pointer here.

Either we use ephemeral or regular storage, the struct will be filled at some point, so allowing the field to be nil does not worth it IMHO. This way we also reduce the risk of a nil pointer exception.

api/v1alpha1/mariadb_webhook.go
```diff
@@ -201,3 +202,16 @@ func (r *MariaDB) validatePodDisruptionBudget() error {
 	}
 	return nil
 }
+
+func (r *MariaDB) validateStorage() error {
+	logger.V(1).Info("Validate MariaDB storage", "mariadb", r.Name)
+
+	if !r.IsEphemeralStorageEnabled() && r.Spec.VolumeClaimTemplate == nil {
```
mmontes11, line 209: You can compare `r.Spec.VolumeClaimTemplate` with an empty struct.

examples/manifests/mariadb_v1alpha1_mariadb_minimal.yaml
```diff
@@ -3,9 +3,4 @@ kind: MariaDB
 metadata:
   name: mariadb-min
 spec:
-  volumeClaimTemplate:
-    resources:
-      requests:
-        storage: 1Gi
-    accessModes:
-      - ReadWriteOnce
+  ephemeralStorage: true
```
mmontes11, line 6: Thanks, indeed, it makes sense to update the minimal example with ephemeral storage and empty root password!

mmontes11 (changes requested):
Thanks a lot for your effort, this is looking good 👍🏻 💯 

Left a couple of comments.

api/v1alpha1/mariadb_webhook.go
```diff
@@ -201,3 +203,24 @@ func (r *MariaDB) validatePodDisruptionBudget() error {
 	}
 	return nil
 }
+
+func (r *MariaDB) validateStorage() error {
+
+	// spec.ephemeralStorage and spec.volumeClaimTemplate shall be mutually exclusive
+	if !r.IsEphemeralStorageEnabled() && reflect.ValueOf(r.Spec.VolumeClaimTemplate).IsZero() {
```
Hoega, line 210: I think it is better to enforce that both ephemeralStorage and VolumeClaimTemplate are mutually exclusive to avoid confusion.

If neither is configured the pod will never be created with this failure but It could be better to add an explicit error in the webhook.
`  Warning  FailedCreate  7s (x12 over 17s)  statefulset-controller  create Pod mariadb-min-0 in StatefulSet mariadb-min failed error: failed to create PVC storage-mariadb-min-0: PersistentVolumeClaim "storage-mariadb-min-0" is invalid: [spec.accessModes: Required value: at least 1 access mode is required, spec.resources[storage]: Required value]`

 How do you test the webhook + controller locally? I have a port conflict when I launch both. @mmontes11

api/v1alpha1/mariadb_webhook.go
Hoega: Done using the reflect package, I'm not a go expert but comparing nested structs with slices seems tricky.

mmontes11: Good shout, I was thinking of:
```go
r.Spec.VolumeClaimTemplate == (corev1.PersistentVolumeClaim{})
```
but you are right, the compiler complains if the struct has slices

api/v1alpha1/mariadb_webhook.go
mmontes11: > I think it is better to enforce that both ephemeralStorage and VolumeClaimTemplate are mutually exclusive to avoid confusion.

Makes sense to me as well. When `ephemeralStorage` is `false`, the `volumeClaimTemplate` must be set as well.

> If neither is configured the pod will never be created with this failure but It could be better to add an explicit error in the webhook.

Right, we should explicitly cover this case in the webhook to avoid non meaningful errors for the user like this one:
 `Required value: at least 1 access mode is required`. This would be very hard to debug for the end user.
 
> How do you test the webhook + controller locally? I have a port conflict when I launch both

Probably you have been using the `make webhoo` I am guessing? This target is only intented to spin up a webhook locally to ensure it runs correctly and tthe TLS configuration is fine. The Kubernetes API server is not configured to route request to the local webhook i.e. no `ValidatingWebhookConfiguration` nor `MutatingWebhookConfiguration` are creating pointing to the local webhook, connectivity-wise it would be hard it would be tricky,

Here it is what you can do:
- Build the `mariadb-operator` image and load it in kind. The tag to be built is specified in the Makefile [here](https://github.com/mariadb-operator/mariadb-operator/blob/e654465e675963603159980844fd9f1853cd1c0a/Makefile#L16):
```bash
make docker-dev
```
- Install the operator locally using the helm chart pointing to the previous tag:
```
helm upgrade --install --set image.tag=v0.0.24 mariadb-operator deploy/charts/mariadb-operator
```

This will run and operator and a webhook with your local changes.

mmontes11: In any case, please add test cases for the webhook suite covering the ephemeral storage functionality

Hoega: Thanks I was able to valide the webhooks, I have updated the PR description with the errors returned.

mmontes11 (changes requested):
Outstanding job! 💯 🙇🏻 

Just some typos to be fixed. Could we also remove this example?
https://github.com/mariadb-operator/mariadb-operator/blob/mxs/examples/manifests/mariadb_v1alpha1_mariadb_empty_root_password.yaml

We now have the `mariadb-minimal` which is more representative. I feel like our example suite is getting big.

mmontes11 (approved):
LGTM! Very much appreciated contributions! 🦭 🙇🏻 

FYI @outductor

PR #334, opened 2024-01-10
Add a webhook validation for root password

Hoega:
When rootEmptyPassword=true and rootPasswordSecretKeyRef are both defined, the secret will be ignored. This can be confusing for the user. Therefore, it is proposed to make both parameters mutually exclusive.

Webhook rejecting a manifest with rootEmptyPassword and rootPasswordSecretKeyRef both set:
`Error from server (Forbidden): error when creating "examples/manifests/mariadb_v1alpha1_mariadb_minimal.yaml": admission webhook "vmariadb.kb.io" denied the request: spec.rootEmptyPassword: Invalid value: true: 'spec.rootEmptyPassword' must be disabled when 'spec.rootPasswordSecretKeyRef' is specified`

When neither is defined, a root password will be randomly generated, eliminating the need for a validation webhook.

### DIFF
--- a/api/v1alpha1/mariadb_types.go
+++ b/api/v1alpha1/mariadb_types.go
@@ -431,6 +431,11 @@ func (m *MariaDB) IsRootPasswordEmpty() bool {
 	return m.Spec.RootEmptyPassword != nil && *m.Spec.RootEmptyPassword
 }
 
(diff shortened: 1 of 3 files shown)


mmontes11 (approved):
Thanks again for your contribution! 🙏🏻 

Totally agree, this aligns pretty much with the ephemeral storage validation

PR #335, opened 2024-01-10
Set securityContext and podSecurityContext for Backup/Restore/Sql Jobs

addreas:
There is currently no way to configure security contexts for jobs that are created by the operator. At first I considered extending the relevant CRDs with security context fields, but then I realized (at least for my situation) the required information already exists in the MariaDB spec.

This made for a very simple implementation, though it does not allow jobs to be ran with security contexts that are different from the MariaDB instance they reference. In my case the issue was pod-security settings, which can be solved with the same security context. I can't come up a reason to want to configure it separately for the Jobs, though I'm not against having it configured separately just to be consistent.


mmontes11 (changes requested):
Thanks for your contribution @addreas ! 🙏🏻 

I think it would make more sense that all the Job tyes had their own `PodSecurityContext` and `SecurityContext` field. They are workloads independent from the `MarkaDB` so the same security restrictions does not necessarily apply. Also, in the long term, we are looking to increase the scope of the `mariadbRef` and point to external database, so there will be no `mariadb.Spec.PodSecurityContext` nor `mariadb.Spec.SecurityContex` available.

addreas:
Very good point! I've updated things so that there is a separate security context for each CRD. I took the liberty to slightly simplify the struct definitions such that they all use the same `ContainerTemplate` and `PodTemplate` struct, not sure if this is the right way to go so opinions are welcome.

mmontes11:
Thank you! Makes sense to me as long as the test pass. It is a pending refactor I was planning to do, but feel free to include just `PodSecurityContext` and `SecurityContext` if it gets too complicated.

addreas:
Saw that the diff ended up very big for this PR if that refactoring was included, so i changed it to simply add the new fields. Since you were already planning to do it, would you like me to make a separate PR for it?

mmontes11 (changes requested):
Changes LGTM! 

Could you please take a look at the linter?:
https://github.com/mariadb-operator/mariadb-operator/actions/runs/7527566252/job/20504999127?pr=335

addreas:
Lint issue was the `jobMariadbOperatorContainer` function signature,  I'm not really a fan of my solution but at least the linter is happy. Perhaps a candidate for some type of refactor in the future.

mmontes11:
> Lint issue was the jobMariadbOperatorContainer function signature, I'm not really a fan of my solution but at least the linter is happy. Perhaps a candidate for some type of refactor in the future.

Agreed, too many arguments. Seems like a good candidate for functional options

mmontes11 (approved):
LGTM 🚀  Thanks for your contribution! 🙏🏻

PR #337, opened 2024-01-10
Add a priorityClassName option

### LINKED ISSUE #265: [Feature] Add priorityClassName for operator and for deployed pods that are part of Mariadb CRD
If in a k8s cluster I have priority classes, including a default one, after bootstrapping mariadb-operator its pods will fail in the default one which usually have a low priority so in case of resource pressure they will be evicted.

Allow setting priorityClassName for operator and for deployed pods that are part of Mariadb CRD.

Deployment for mariadb-operator can be manually edited to include priorityClassName but is not a nice solution.
For Mariadb CRD I have tried mutating pods using kyverno/gatekeeper but in case a pod is killed it will never be recreated.

**Additional context**

**Environment details**:

Hoega:
Closes #265 

Add an option to specify a priority class to use.

```
kind: MariaDB
metadata:
  name: mariadb-min
spec:
  priorityClassName: high-priority
```

If the priority class is missing, the Statefulset will fail until a priority class is created. Does not create a priority class automatically.
`Warning  FailedCreate  4s (x12 over 14s)  statefulset-controller  create Pod mariadb-min-0 in StatefulSet mariadb-min failed error: pods "mariadb-min-0" is forbidden: no PriorityClass with name high-priority was found`

`k get po mariadb-min-0 -o yaml | grep priorityClass -B 2`
```
  preemptionPolicy: PreemptLowerPriority
  priority: 1000000
  priorityClassName: high-priority
```


mmontes11 (approved):
Thanks for your contribution! 🙏🏻 LGTM 🚀

PR #338, opened 2024-01-10
Add prefix for S3 backup/restore

addreas:
Closes https://github.com/mariadb-operator/mariadb-operator/issues/344

Something we noticed while setting things up is that backups in S3 buckets are stored only in the root, without any reference to the database name. This adds a `prefix` field to the S3 config so that backups can be placed wherever one might want them.

It might be of interest to have some kind of default path that contains more things to identify the backup, though I don't have any strong opinions about how that should look.

### DIFF
--- a/api/v1alpha1/base_types.go
+++ b/api/v1alpha1/base_types.go
@@ -266,6 +266,10 @@ type S3 struct {
 	// +optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	TLS *TLS `json:"tls,omitempty"`
+	// Prefix allows backups to be placed under a specific prefix in the bucket.
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	Prefix string `json:"prefix"`
 }
 
 // RestoreSource defines a source for restoring a MariaDB.
--- a/cmd/backup/main.go
+++ b/cmd/backup/main.go
@@ -24,6 +24,7 @@ var (
 	s3Region       string
 	s3TLS          bool
 	s3CACertPath   string
+	s3Prefix       string
 	maxRetention   time.Duration
 )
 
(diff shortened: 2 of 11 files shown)


api/v1alpha1/base_types.go
```diff
@@ -266,6 +266,8 @@ type S3 struct {
 	// +optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	TLS *TLS `json:"tls,omitempty"`
+
+	Prefix string `json:"prefix"`
```
mmontes11, line 270: Could you documment and annotate this field like the other please?

cmd/backup/main.go
```diff
@@ -47,6 +48,7 @@ func init() {
 	RootCmd.PersistentFlags().BoolVar(&s3TLS, "s3-tls", false, "Enable S3 TLS connections.")
 	RootCmd.PersistentFlags().StringVar(&s3CACertPath, "s3-ca-cert-path", "s3/pki/tls.crt",
 		"Path to the CA to be trusted when connecting to S3.")
+	RootCmd.PersistentFlags().StringVar(&s3Prefix, "s3-prefix", "", "Set S3 prefix.")
```
mmontes11, line 51:
```suggestion
	RootCmd.PersistentFlags().StringVar(&s3Prefix, "s3-prefix", "", "S3 bucket prefix name to use.")
```

pkg/backup/storage.go
```diff
@@ -124,17 +132,17 @@ func (s *S3BackupStorage) List(ctx context.Context) ([]string, error) {
 
 func (s *S3BackupStorage) Push(ctx context.Context, fileName string) error {
 	filePath := filepath.Join(s.basePath, fileName)
-	_, err := s.client.FPutObject(ctx, s.bucket, fileName, filePath, minio.PutObjectOptions{})
+	_, err := s.client.FPutObject(ctx, s.bucket, s.Prefix+fileName, filePath, minio.PutObjectOptions{})
```
mmontes11, line 135: Needs to be added to the `List` operation as well.

mmontes11 (changes requested):
Thanks for the PR! 🙏🏻 You are on the right way

Left a couple of comments

api/v1alpha1/base_types.go
addreas: Sure thing!

pkg/backup/storage.go
addreas: Oh, missed that completely. Added it to the `ListObjectsOptions`.

mmontes11 (changes requested):
Changes LGTM! Thanks for your contribution.

It seems like you haven't re-generated the artifacts, could you run `make run` and commit?

addreas:
Artifacts re-generated now :+1:

mmontes11 (approved):
LGTM 🚀  Thanks a lot for your contribution, very much appreciated 🙏🏻 🦭

Issue #333, opened 2024-01-10
[Feature] Add service account name to MariaDB spec


hv036:
**Is your feature request related to a problem? Please describe.**
I believe we should allow users to define their own service accounts. Additionally, in clusters prior to version 1.25 with Pod Security Policies (PSPs) where running as root is restricted, defining a custom service account for MariaDB will be beneficial

**Describe the solution you'd like**
Add `serviceAccountName` in [MariaDBSpec](https://github.com/mariadb-operator/mariadb-operator/blob/main/docs/API_REFERENCE.md#mariadbspec)

**Environment details**:
- mariadb-operator version: 0.0.24
- Install method: helm

Hoega:
I think it should be completed now, PR #339 implements a serviceAccountName parameter.

mmontes11:
Supported in the following PR:
- https://github.com/mariadb-operator/mariadb-operator/pull/339

PR #339, opened 2024-01-11
Add an option to select a serviceAccount

### LINKED ISSUE #244: [Feature] Allow setting ServiceAccountName / don't use "default" ServiceAccount
When applying security policies like "no default service account name", currently the only way to do this is by creating a galera cluster. 

https://github.com/mariadb-operator/mariadb-operator/blob/b25c694a80d1064de120d623fede25824af9eda5/pkg/builder/statefulset_builder.go#L180-L187

I would like to either set the name manual, or just set a "random" name in `buildStsServiceAccountName` when not galera.

Alternative is deploying galera cluster. but you might not want to do that.

**Additional context**
-

**Environment details**:
- Kubernetes version: 1.26
- mariadb-operator version: latest (0.20.0)
- Install method: helm
- Install flavour: custom (i guess)

Hoega:
Closes #244  and #333 
Instead of using the default service account, a new service account with the mariadb name will be used:

`k get po mariadb-min-0 -o yaml`
```
     volumeMounts:
    - mountPath: /var/lib/mysql
      name: storage
    - mountPath: /etc/mysql/conf.d
      name: config
    - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
      name: kube-api-access-n2s4z
      readOnly: true
  automountServiceAccountToken: true
  serviceAccount: mariadb-min
  serviceAccountName: mariadb-min
```

You can also now specify a service account with a new option serviceAccountName:
```
kind: MariaDB
metadata:
  name: mariadb-min
spec:
  serviceAccountName: test123
```

If the service account does not exist, it will be created:
`  k get serviceaccounts `
```
NAME      SECRETS   AGE
default   0         49m
test123   0         114s
```


pkg/builder/statefulset_builder.go
```diff
@@ -157,12 +158,13 @@ func buildStsVolumeClaimTemplates(mariadb *mariadbv1alpha1.MariaDB) []corev1.Per
...
 func buildStsServiceAccountName(mariadb *mariadbv1alpha1.MariaDB) (autoMount *bool, serviceAccount string) {
-	if mariadb.Galera().Enabled {
-		mount := false
-		autoMount = &mount
-		serviceAccount = mariadb.Name
+	if mariadb.Spec.ServiceAccountName != nil {
+		if mariadb.Spec.Galera.Enabled {
+			return ptr.To(false), *mariadb.Spec.ServiceAccountName
+		}
+		return ptr.To(true), *mariadb.Spec.ServiceAccountName
```
Hoega, line 165: Do we need the automount = true? The serviceaccount token was mounted when the default SA was used, is it used?

mmontes11: We can set it to false in the non Galera clusters, as we are not currently using it. For Galera, we are explicitly mounting it here:
https://github.com/mariadb-operator/mariadb-operator/blob/main/pkg/builder/statefulset_builder.go#L197

Juding by https://github.com/mariadb-operator/mariadb-operator/issues/244 and https://github.com/mariadb-operator/mariadb-operator/issues/333, this issue is about be compliant with security policies, which in general are against the default `ServiceAccount`.

pkg/builder/statefulset_builder.go
```diff
@@ -157,12 +158,13 @@ func buildStsVolumeClaimTemplates(mariadb *mariadbv1alpha1.MariaDB) []corev1.Per
 }
 
 func buildStsServiceAccountName(mariadb *mariadbv1alpha1.MariaDB) (autoMount *bool, serviceAccount string) {
```
mmontes11, line 160:
```suggestion
func buildStsServiceAccountName(mariadb *mariadbv1alpha1.MariaDB) (serviceAccount string) {
```
We no longer need the `autoMount` flag, we can hardcode this to `false` in the builder.

mmontes11: For Galera, we are explicitly mounting it here:
https://github.com/mariadb-operator/mariadb-operator/blob/main/pkg/builder/statefulset_builder.go#L197

mmontes11 (changes requested):
Thanks for your PR @Hoega ! As always, very much appreciated 🙏🏻 

I've left a couple of suggestions, let me know what you think

mmontes11 (approved):
LGTM! 👍🏻 Thanks for your contribution! 🦭

PR #347, opened 2024-01-17
Add args option to restore jobs

### LINKED ISSUE #238: [Feature] Support extra arguments in `Restore` jobs
The same way we did for `Backup` jobs, support passing extra arguments by specifying `spec.args`. See:
- https://github.com/mariadb-operator/mariadb-operator/pull/162

**Additional context**
https://github.com/mariadb-operator/mariadb-operator/blob/88fe1fcfc06c8229bde8ba2eca82bad9467cf1b7/api/v1alpha1/backup_types.go#L46
https://github.com/mariadb-operator/mariadb-operator/blob/88fe1fcfc06c8229bde8ba2eca82bad9467cf1b7/pkg/builder/batch_builder.go#L147

Hoega:
Closes #238 

Add an 'args' option to the CRDs for restore jobs, similar to the existing option for backup jobs.

### DIFF
--- a/api/v1alpha1/restore_types.go
+++ b/api/v1alpha1/restore_types.go
@@ -16,6 +16,10 @@ type RestoreSpec struct {
 	// +kubebuilder:validation:Required
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	MariaDBRef MariaDBRef `json:"mariaDbRef" webhook:"inmutable"`
+	// Args to be used in the Restore container.
+	// +optional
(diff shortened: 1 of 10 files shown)


mmontes11 (changes requested):
Thanks again for this contribution @Hoega 🙏🏻 

Looks good, could you please add the following?
- Test that provides args to a `Restore`. You can reuse an existing one, no need to create a new test dedicated for this.
- Example that provides args to a `Restore`. Similarly, you can reuse an existing example.

mmontes11 (approved):
LGTM! 🚀 🙏🏻

Issue #343, opened 2024-01-15
[Feature] Add golangci-lint to ci


Hoega:
**Is your feature request related to a problem? Please describe.**

[golangci-lint](https://github.com/golangci/golangci-lint) highlights some errors depending on the set of linters enabled. For example some issues are:

```
pkg/controller/certificate/controller.go:97:15: ST1005: error strings should not be capitalized (stylecheck)
                return nil, fmt.Errorf("Error reconciling CA KeyPair: %v", err)
                            ^
pkg/controller/certificate/controller.go:104:16: ST1005: error strings should not be capitalized (stylecheck)
                        return nil, fmt.Errorf("Error reconciling CA KeyPair: %v", err)
                                    ^
pkg/controller/rbac/controller.go:164:11: ST1005: error strings should not end with punctuation or newlines (stylecheck)
                        return fmt.Errorf(
                               ^
controller/suite_test_data.go:19: File is not `gci`-ed with --skip-generated -s standard -s default (gci)
        //+kubebuilder:scaffold:imports
controller/suite_test.go:36: File is not `gci`-ed with --skip-generated -s standard -s default (gci)
        //+kubebuilder:scaffold:imports
controller/connection_controller.go:265:36: underef: could simplify (*conn.Spec.HealthCheck.RetryInterval).Duration to conn.Spec.HealthCheck.RetryInterval.Duration (gocritic)
                return ctrl.Result{RequeueAfter: (*conn.Spec.HealthCheck.RetryInterval).Duration}, nil
```

**Describe the solution you'd like**

First, fix the reported linters errors reported by `golang-lint run`.
TBD which set of linters to use, I propose this for starters but there are many more:
```
linters:
  enable:
    - dupword
    - errcheck
    - exportloopref
    - ginkgolinter
    - gci
    - gocritic
    - gosimple
    - govet
    - ineffassign
    - mirror
    - staticcheck
    - stylecheck
    - unused
    - whitespace
```
Then, add the golangci to github action for automatic CI validation.

**Environment details**:
- mariadb-operator version: main

**Additional context**
[golangci-lint](https://github.com/golangci/golangci-lint)
[golangci-lint-action](https://github.com/golangci/golangci-lint-action)

mmontes11:
Hey there @Hoega ! Thanks for reporting.

I'm a bit confused, I don't see any error poping up when running `make lint`:
```bash
 make lint
[...]

PR #361, opened 2024-01-28
Enable ginkgolinter

nunnatsa:
Bump golangci-lint to the latest version (v1.55.2)

Enable [ginkgolinter](https://github.com/nunnatsa/ginkgolinter) and fix findings.

related to #343


controller/mariadb_controller_test.go
```diff
@@ -130,33 +130,33 @@ var _ = Describe("MariaDB controller", func() {
 			}, testTimeout, testInterval).Should(BeTrue())
 
 			By("Expecting to create a exporter Deployment eventually")
-			Eventually(func() bool {
+			Eventually(func(g Gomega) bool {
 				var deploy appsv1.Deployment
 				if err := k8sClient.Get(testCtx, testMariaDb.MetricsKey(), &deploy); err != nil {
 					return false
 				}
 				expectedImage := os.Getenv("RELATED_IMAGE_EXPOTER")
```
mmontes11, line 138:
```suggestion
				expectedImage := os.Getenv("RELATED_IMAGE_EXPORTER")
```

There is a typo here that makes the test fail, as the env variable doesn't exist. Not sure why it was passing before, guessing that using the wrong Gomega instance has implications i.e `g.Expect` vs `Expect` ?

controller/mariadb_controller_test.go
```diff
@@ -130,33 +130,33 @@ var _ = Describe("MariaDB controller", func() {
...
+				g.Expect(expectedImage).ToNot(BeEmpty())
 				By("Expecting Deployment to have exporter image")
-				Expect(deploy.Spec.Template.Spec.Containers).To(ContainElement(MatchFields(IgnoreExtras,
+				g.Expect(deploy.Spec.Template.Spec.Containers).To(ContainElement(MatchFields(IgnoreExtras,
 					Fields{
 						"Image": Equal(expectedImage),
 					})))
 				return deploymentReady(&deploy)
-			})
+			}).Should(BeTrue())
```
mmontes11, line 146:
```suggestion
			}, testTimeout, testInterval).Should(BeTrue())
```
These `Eventually` block is wrong, missing timeouts. Again, not sure why it did work before.

controller/mariadb_controller_test.go
```diff
@@ -130,33 +130,33 @@ var _ = Describe("MariaDB controller", func() {
...
 				if err := k8sClient.Get(testCtx, testMariaDb.MetricsKey(), &svcMonitor); err != nil {
 					return false
 				}
-				Expect(svcMonitor.Spec.Selector).NotTo(BeNil())
-				Expect(svcMonitor.Spec.Selector).To(HaveKeyWithValue("app.kubernetes.io/name", "mariadb"))
-				Expect(svcMonitor.Spec.Selector).To(HaveKeyWithValue("app.kubernetes.io/instance", testMariaDbName))
-				Expect(svcMonitor.Spec.Endpoints).To(HaveLen(1))
+				g.Expect(svcMonitor.Spec.Selector).NotTo(BeNil())
+				g.Expect(svcMonitor.Spec.Selector).To(HaveKeyWithValue("app.kubernetes.io/name", "mariadb"))
+				g.Expect(svcMonitor.Spec.Selector).To(HaveKeyWithValue("app.kubernetes.io/instance", testMariaDbName))
```
mmontes11, line 156:
```suggestion
				g.Expect(svcMonitor.Spec.Selector.MatchLabels).NotTo(BeNil())
				g.Expect(svcMonitor.Spec.Selector.MatchLabels).To(HaveKeyWithValue("app.kubernetes.io/name", "exporter"))
				g.Expect(svcMonitor.Spec.Selector.MatchLabels).To(HaveKeyWithValue("app.kubernetes.io/instance", testMariaDbName))
```
These assertions are failing also. Wrong Gomega instance?

controller/mariadb_controller_test.go
```diff
@@ -130,33 +130,33 @@ var _ = Describe("MariaDB controller", func() {
...
+				g.Expect(svcMonitor.Spec.Endpoints).To(HaveLen(1))
 				return true
-			})
+			}).Should(BeTrue())
```
mmontes11, line 159:
```suggestion
			}, testTimeout, testInterval).Should(BeTrue())
```
Same as previous block.

mmontes11 (changes requested):
Thanks a lot for this PR @nunnatsa ! Very much appreciated 🙏🏻 

I think you uncovered two `Eventually` blocks that were misbehaving, I've added some commit suggestions that make the tests happy again.

controller/mariadb_controller_test.go
nunnatsa: Thanks for explaining that. The reason it was passing is the missing assertion of the `Eventually` container. Nothing actually checked this Eventually. 

The different between `Expect` and `g.Expect` is that the first exit Eventually/Constantly immediately, while the second only fails the current loop, letting Eventually to pole again and again until it's eventually working. In most cases, this is what we want.

controller/mariadb_controller_test.go
nunnatsa: Again, missing assertion (`Should`) for this `Eventually`. Anything could pass here.

For example:

`Expect(false)` will not fail the test, but `Expect(false).To(BeTrue())` will.

controller/mariadb_controller_test.go
nunnatsa: It's not totally wrong because there are default polling interval and timeout, but I fixed to match this specific test.

controller/mariadb_controller_test.go
mmontes11: Thanks a lot for the explanation, that makes sense now.

mmontes11 (approved):
LGTM 🚀  Thanks a lot for the contribution and the valuable insights about Gomega

PR #382, opened 2024-02-13
Add Glasskube to Adopters list

pmig:
Glasskube first introduced the `mariadb-operator` back in November 2022 (https://github.com/glasskube/operator/issues/20) for supporting Matomo via our apps operator and happily using (and migrating between versions) ever since.


mmontes11 (approved):
Thanks for using `mariadb-operator` 🦭 🧊 ! Looking forward to collaborate in the future!

PR #384, opened 2024-02-14
Add `init` and `agent` to glued CLI

mmontes11:


perfectra1n:
Does this mean that at some point in the near future that users should then no longer use the `agent` image, and should then use something else? Or what changes does this mean for users that have existing `spec.galera` entries? (Not sure if I'm missing something or if this is just part of a new version that will have updated documentation)

mmontes11:
Hey there @perfectra1n ! Yes, exactly you get it right. In further releases, you should be using the `mariadb-operator` image for both the `agent` and the `init`. 

I still need to write up documentation about this before the release as well as the migration guide, don't worry about that, just be sure to check the GitHub release before upgrading.

`v0.0.26` fixes critical crashes in Galera and improves the overall stability and specially, the cluster recovery process.It will be released next week. Stay tuned!

Issue #381, opened 2024-02-13
[Bug] 0.25.0 maxscale statefulset failing to start


pasztorl:
**Documentation**

- [ x ] I acknowledge that I have read the relevant [documentation](https://github.com/mariadb-operator/mariadb-operator/tree/main/docs).

**Describe the bug**

I'm testing the latest operator by creating a mariadb replication with maxscale. The maxscale statefulset pods failing to start because of a permission problem.
Here is the log output from the maxscale pod: 
```
2024-02-13 13:53:51   notice : MaxScale will be run in the terminal process.
2024-02-13 13:53:51   notice : /etc/config/..2024_02_13_13_53_43.4123823925/maxscale.cnf.d does not exist, not reading.
2024-02-13 13:53:51   notice : /var/lib/maxscale/maxscale.cnf.d does not exist, not reading.
2024-02-13 13:53:51   notice : Using up to 2.29GiB of memory for query classifier cache
2024-02-13 13:53:51   notice : syslog logging is disabled.
2024-02-13 13:53:51   notice : maxlog logging is enabled.
2024-02-13 13:53:51   error  : Failed to create directory '/var/lib/maxscale/maxscale.cnf.d': 13, Permission denied
2024-02-13 13:53:51   alert  : Can't access '/var/lib/maxscale/maxscale.cnf.d'.: No such file or directory
```
I see that the pod have a volume which mounted to this directory, the problem is that it tries to create a directory inside this mountpoint as maxscale user which is not allowed by the mountpoint filesystem permissions.

Here is the statefulset spec for the maxscale what the operator created:
```
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: test1-mariadb-maxscale
  namespace: test1-mariadb
  ...
spec:
  replicas: 1
  selector:
    matchLabels:
      app.kubernetes.io/instance: test1-mariadb-maxscale
      app.kubernetes.io/name: maxscale
  template:
    metadata:
      name: test1-mariadb-maxscale
      namespace: test1-mariadb
      creationTimestamp: null
      labels:
        app.kubernetes.io/instance: test1-mariadb-maxscale
        app.kubernetes.io/name: maxscale
    spec:
      volumes:
        - name: config
          secret:
            secretName: test1-mariadb-maxscale-config
            defaultMode: 420
      containers:
        - name: maxscale
          image: mariadb/maxscale:23.08
          command:
            - maxscale
          args:
            - '--config'
[...]

PR #403, opened 2024-02-21
Added MaxScale chown init container. Support for MaxScale init containers

mmontes11:
Related to https://github.com/mariadb-operator/mariadb-operator/issues/381

Kudos to @K-MeLeOn


lwj5:
not sure if it's there, but do add the resource limits from maxscale to the init container. If not in clusters that require limit will fail to start

Issue #442, opened 2024-03-10
[Bug] Anti Affinity


drag0n141:
**Documentation**
- [x] I acknowledge that I have read the relevant [documentation](https://github.com/mariadb-operator/mariadb-operator/tree/main/docs).

**Describe the bug**
When is set `enableAntiAffinity: true` and bootstrap the affinity rule is also set to the galera-cluster-restore pod. But i have only three nodes and because of that the restore can not start. I think it would be the best when the affinity rule is not set to the restore pod.

**Environment details**:
- Kubernetes version: 1.29.2
- Kubernetes distribution: k3s
- mariadb-operator version: 0.27.0
- Install method: helm
- Install flavor: recommended

drag0n141:
I have tested now a little bit more, the metrics pods have the same problem. It can not deployed when you have a galera cluster with 3 replicas and "only" three nodes.

sennerholm:
It's also impact the metric pod if you try to set it to not be used there, like
```
apiVersion: k8s.mariadb.com/v1alpha1
kind: MariaDB
metadata:
  name: mariadb
spec:
  affinity:
    enableAntiAffinity: true
  metrics:
    enabled: true
    exporter:
      affinity:
        enableAntiAffinity: false
```

sennerholm:
In the new 0.32.0 it's not working. The antiaffinity rules on the metric deployment is removed (step 1), but the generated rules on the Statefulset is using
`app.kubernetes.io/instance=mariadb`
and that is also set on the metric pod...

mmontes11:
Hey there! Thanks for reporting.

Metrics anti-affinity issue has been sorted our in the following PR:
- https://github.com/mariadb-operator/mariadb-operator/pull/446

Regarding the `Restore`,  the issue is that we are re-using the same `PodTemplate` as `MariaDB`, and therefore the same anti-affinity-rules as well:
https://github.com/mariadb-operator/mariadb-operator/blob/470218ee348a99705d8eb1b1592423ded97a4b97/pkg/builder/restore_builder.go#L22

Will raise a PR soon!

mmontes11:
In any case, you are still able to provide the anti-affinity rules yourself:

```yaml
kind: MariaDB
metadata:
  name: mariadb-galera
spec:
...
  affinity:
    enableAntiAffinity: true
    podAntiAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
      - labelSelector:
          matchExpressions:
[...]

PR #446, opened 2024-03-11
Fix metrics selector to avoid clashing with `MariaDB`. Sort out anti-affinity cases where there aren't enough nodes for the exporter `Pod`

mmontes11:
This PR fixes the metrics selector labels to have a different name than `MariaDB`. 

This allows to sort out anti-affinity situations where there aren't enough nodes to place the metrics exporter `Pod`.  Anti affinity needs to be disabled for metrics in such case:

```yaml
kind: MariaDB
metadata:
  name: mariadb-galera
spec:
...
  metrics:
    enabled: true
    exporter:
      affinity:
        enableAntiAffinity: false

  affinity:
    enableAntiAffinity: true
```

Related to:
- https://github.com/mariadb-operator/mariadb-operator/issues/442

### DIFF
--- a/controller/mariadb_controller_test.go
+++ b/controller/mariadb_controller_test.go
@@ -172,7 +172,7 @@ var _ = Describe("MariaDB controller", func() {
 				}
 				g.Expect(svcMonitor.Spec.Selector.MatchLabels).NotTo(BeEmpty())
 				g.Expect(svcMonitor.Spec.Selector.MatchLabels).To(HaveKeyWithValue("app.kubernetes.io/name", "exporter"))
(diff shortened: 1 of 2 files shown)


Trackhe:
sadly it seems not working for me. i can't get it running with the v0.27, the metrics pods say always 

`[Unschedulable] 0/3 nodes are available: 3 node(s) didn't satisfy existing pods anti-affinity rules. preemption: 0/3 nodes are available: 3 No preemption victims found for incoming pod.`

mmontes11:
That's expected @Trackhe ! This has been merged but not released yet. It will be available in `v0.0.28`, stay tuned 🚀

Issue #345, opened 2024-01-16
[bug] Missing support for private image repositories in Backup and Metrics


andreasgeisslerdt:
**Documentation**
- [x] I acknowledge that I have read the relevant [documentation](https://github.com/mariadb-operator/mariadb-operator/tree/main/docs).

**Describe the bug**
Installing mariadb-operator using private docker repositories requires the definition of registries and imagePullSecrets.
This generally works for the deployment of mariadb-operator via the values.yaml overrides: 

```
---
image:
  <private registry>/mariadb-operator/mariadb-operator
imagePullSecrets:
  - name: <registry secret>
webhook:
  image:
    <private registry>/mariadb-operator/mariadb-operator
  cert:
    certManager:
      enabled: true
  imagePullSecrets:
    - name: <registry secret>
metrics:
  enabled: true
```
Unfortunately for the Backup Job as well as for the Metrics Exporter the image information is taken from a predefined CM:

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: mariadb-operator-images
  namespace: mariadb-operator
data:
  MARIADB_OPERATOR_IMAGE: ghcr.io/mariadb-operator/mariadb-operator:v0.0.24
  RELATED_IMAGE_EXPORTER: prom/mysqld-exporter:v0.15.1
  RELATED_IMAGE_MARIADB: mariadb:11.2.2
```
Additionally for the JobTemplate used by the Backup CronJob an imagePullSecret cannot be added

**Expected behaviour**
The configuration options in the mariadb-operator to override images and imagePullSecrets should be taken into account in the jobTemplate for the Backup CronJob.
Additionally for "metrics" an "image" and "imagePullSecrets" config options could be added on operator level

**Steps to reproduce the bug**

1. deploy mariadb-operator including private registry and imagePullSecrets
2. create mariaDB and Backup resources
3. Check the create CronJob

mmontes11:
Hey there @addreas ! Thanks for reporting this

You are right, neither the `Backup` nor the `Restore` specs support providing a custom image at the moment, we are using the same as the `MariaDB` resource they point to:

- https://github.com/mariadb-operator/mariadb-operator/blob/main/docs/API_REFERENCE.md#backupspec
- https://github.com/mariadb-operator/mariadb-operator/blob/main/docs/API_REFERENCE.md#restorespec

We should include this fields here and default to the `MariaDB` images when not available. Contributions are welcome!

[...]

PR #459, opened 2024-03-14
Support for private registries in all CRDs via `ImagePullSecrets`

mmontes11:
Close https://github.com/mariadb-operator/mariadb-operator/issues/345


andreasgeisslerdt:
Great !!
Many thanks for spending the effort.
Andreas

Issue #484, opened 2024-03-19
[Bug] Galera config partially updated when `MariaDB` requests fails in Galera init container


mmontes11:
**Describe the bug**
When the `MariaDB` request fails in the Galera init container (i.e. control-plane down):
https://github.com/mariadb-operator/mariadb-operator/blob/7ba652aea1151c4ea10d46c45b84035130051bc2/cmd/init/main.go#L95
The Galera initContainer works in minimal mode without depending on the `MariaDB` object. In this mode, only the `Pod` IP is updated in the Galera config. file.

Recent changes have been introduced in this PR to support Galera IPv6:
https://github.com/mariadb-operator/mariadb-operator/pull/461

We need to port this changes to the following functions:
- https://github.com/mariadb-operator/mariadb-operator/blob/5031917a8dcc3e0a978f53031b48258d67c106cd/pkg/galera/config/config.go#L190
- https://github.com/mariadb-operator/mariadb-operator/blob/5031917a8dcc3e0a978f53031b48258d67c106cd/cmd/init/main.go#L155

This functions are used to update the `wsrep_node_address` currently, but we must also update the `wsrep_sst_receive_address` and `wsrep_provider_options` fields, like we do here:
https://github.com/mariadb-operator/mariadb-operator/blob/5031917a8dcc3e0a978f53031b48258d67c106cd/pkg/galera/config/config.go#L41

**Expected behaviour**
`wsrep_node_address`, `wsrep_sst_receive_address` and `wsrep_provider_options` are updated even if the `MariaDB` request in the Galera init container fails.

PR #487, opened 2024-03-21
Update Galera config when `MariaDB` request fails

mmontes11:
Closes https://github.com/mariadb-operator/mariadb-operator/issues/484


mmontes11:
FYI @djakobczak1. I've missed this case when reviewing the Galera IPv6 support last week (https://github.com/mariadb-operator/mariadb-operator/pull/461). This PR aims to fix the case described in https://github.com/mariadb-operator/mariadb-operator/issues/484, and also used the chance to do some minor refactors and improve the tests.

PR #524, opened 2024-03-29
fix: cleanup backup files not working with s3-prefix option

fvillain:
## Problem

When using S3 Storage option with a `Prefix` file the uploaded file are never matching the conditions required in shouldProcessBackupFile because it doesn't take into account the prefix.

As a consequence, the `maxRetention` is not respected and expired backup files are never deleted.

**Example with `prefix = foobar_`**
Backup filenames will be named:
```
foobar_backup.2024-03-29T11:20:02Z.sql
```
This file will never get cleaned because if this condition:
```
if !strings.HasPrefix(filename, "backup.")||!strings.HasSuffix(filename, ".sql") {
```

## Solution

> Please keep in mind this is a very "raw" proposal as I'm not a go developer, let's say it's more of a very-quick proposal for a fix that should work.

We can modify the method `shouldProcessBackupFile` and add a `prefix` parameter defaulting to `null` 
If the prefix value is null or an empty string then the behaviour of the method is not modified

If the prefix is not null nor empty then we can gently remove the prefix from the given `fileName` parameter to keep the underlying logic identical.

This should work imo but has not been tested.

We should also have a test for that I guess, but i don't know how to write that

## How I tested and validated the bug

It's dirty, and I know it :heart:

```go
package main

import (
	"fmt"
	"strings"
	"time"
)

func main() {
	filename := "foobar_backup.2024-03-29T10:00:03Z.sql"
	maxRetention := "1h"

	parts := strings.Split(filename, ".")
	timeRaw := parts[1]
	timeBackup, err := time.Parse(time.RFC3339, timeRaw)
	if err != nil {
		fmt.Println("unable to parse time")
	}

	now := time.Now()
	fmt.Println("Now: ")
	fmt.Println(now)

	retentionDuration, err := time.ParseDuration(maxRetention)
	if err != nil {
		fmt.Println(err)
	}

	nowMinusTimeBackup := now.Sub(timeBackup)

	fmt.Println("BackupTime:")
	fmt.Println(timeBackup)
	fmt.Println("Diff:")
	fmt.Println(nowMinusTimeBackup)
	if nowMinusTimeBackup > retentionDuration {
		fmt.Println("File should be deleted")
	} else {
		fmt.Println("File should NOT be deleted")
	}

	//if !strings.HasPrefix(filename, "backup.")||!strings.HasSuffix(filename, ".sql") {
	//	fmt.Println("false")
	//} else {
	//	fmt.Println("true")
	//}
	fmt.Println(strings.HasPrefix(filename, "foobar_"))
	fmt.Println(strings.Contains(filename, "backup."))
	fmt.Println(strings.HasSuffix(filename, ".sql"))
}
```

Thanks for your patience while reading this :pray:

### DIFF
--- a/pkg/backup/storage.go
+++ b/pkg/backup/storage.go
@@ -39,7 +39,7 @@ func (f *FileSystemBackupStorage) List(ctx context.Context) ([]string, error) {
 	var fileNames []string
 	for _, e := range entries {
 		fileName := e.Name()
-		if shouldProcessBackupFile(fileName, f.logger) {
+		if shouldProcessBackupFile(fileName, nil, f.logger) {
 			fileNames = append(fileNames, fileName)
 		}
 	}
@@ -130,7 +130,11 @@ func (s *S3BackupStorage) List(ctx context.Context) ([]string, error) {
 		Prefix: s.Prefix,
 	}) {
 		fileName := o.Key
-		if shouldProcessBackupFile(fileName, s.logger) {
+		prefix := nil
+		if s.Prefix != nil && s.Prefix != "" {
+			prefix := s.Prefix
+		}
+		if shouldProcessBackupFile(fileName, prefix, s.logger) {
 			fileNames = append(fileNames, fileName)
 		}
 	}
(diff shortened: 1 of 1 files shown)


mmontes11:
Hey @fvillain ! Thanks a lot for raising this PR and the effor documenting the issue.

Your changes look good, but I think the prefix implementation needed a propper implementation, some refactoring and automated tests, so I've raised another PR
- https://github.com/mariadb-operator/mariadb-operator/pull/554

The previous PR was based on your changes, so I've added you as a co-author. Thanks for your contribution 🙏🏻

Issue #549, opened 2024-04-10
[Bug] Reduce CRD size by introducing `JobContainerTemplate` and `JobPodTemplate` 


mmontes11:
**Describe the bug**
Our CRD size has notoriously increased when we embedded the `ContainerTemplate` and `PodTemplate` structs in the `Job` related CRs. We are almost hiting hard limits in regards to CRD size, for instance, helm keeps the charts in `Secrets` limited to 1Mb and the `ClusterServiceVersion` (which embeds CRDs) is shipped in a `ConfigMap`. 

**Expected behaviour**
Create a minimal version of  `ContainerTemplate` and `PodTemplate` for the `Job` related CRs: `JobContainerTemplate` and `JobPodTemplate`

mmontes11:
Currently our CRD bundle is `3,4M`:
```bash
❯ ls -h deploy/crds/
total 3,5M
drwxrwxr-x 2 martin martin 4,0K abr 15 16:40 .
drwxrwxr-x 5 martin martin 4,0K mar  8 13:15 ..
-rw-rw-r-- 1 martin martin 3,4M abr 15 16:40 crds.yaml
```

mmontes11:
Removing non used fields in the batch CRDs:
```bash
❯ ls -h deploy/crds/
total 2,8M
drwxrwxr-x 2 martin martin 4,0K abr 15 16:40 .
drwxrwxr-x 5 martin martin 4,0K mar  8 13:15 ..
-rw-rw-r-- 1 martin martin 2,8M abr 15 20:06 crds.yaml
````
0.6M 18% less

PR #561, opened 2024-04-15
Reduce CRD size

mmontes11:
Closes https://github.com/mariadb-operator/mariadb-operator/issues/549


mmontes11:
As described in https://github.com/mariadb-operator/mariadb-operator/issues/549, we have reduced a 18% the CRD size

Issue #490, opened 2024-03-22
Allow to skip TLS verification for backup storage[Feature]


volvera10:
**Is your feature request related to a problem? Please describe.**
Sometimes users run private S3-compatible storage with self-issued TLS certificate.

In that case Operators should be able to store backups there by either skipping TLS verification or by accepting self-issued CAs

**Describe the solution you'd like**
Allow for skipping TLS verification, such as:
```
tls:
  enabled: true
  caSecretKeyRef:
      name: minio-ca
       key: ca.crt
  insecureSkipTLSVerify: true
```
**Describe alternatives you've considered**

**Environment details**:
- Kubernetes version: 1.27.3
- Kubernetes distribution: kubeadm
- mariadb-operator version: 0.27.0
- Install method: helm
- Install flavor: recommended 

**Additional context**

volvera10:
@mmontes11 I am trying to create a PR for the above mods to add `insecureSkipTLSVerify`, however I keep getting access denied when attempting to create a new branch.  
I can post the proposed changes here? 
Please advise.

`C:\Users\d-veronica.olvera\mariadb-operator>git push origin tls-skip-verify
ERROR: Permission to mariadb-operator/mariadb-operator.git denied to volvera10.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.`

PR #690, opened 2024-06-17
Adding optional `insecureSkipTLSVerify` to TLS struct

volvera10:
Added optional config item `insecureSkipTLSVerify` to TLS struct. 

By default, when TLS is enabled, `insecureSkipTLSVerify` will be defaulted to false if not provided.

Linked issue:
https://github.com/mariadb-operator/mariadb-operator/issues/490


volvera10:
@mmontes11 what is the status of this PR getting reviewed?

mmontes11 (changes requested):
Thanks for your PR!

This would be the APi changes we need for implementing this, but we will need to propagate this further down to actually implement this. Please take a look at:
- https://github.com/mariadb-operator/mariadb-operator/blob/4788a7e43380b6fb2000bbb64224e60ca8c070b6/pkg/builder/batch_builder.go#L430
- https://github.com/mariadb-operator/mariadb-operator/blob/4788a7e43380b6fb2000bbb64224e60ca8c070b6/cmd/backup/main.go#L48
- https://github.com/mariadb-operator/mariadb-operator/blob/4788a7e43380b6fb2000bbb64224e60ca8c070b6/pkg/minio/minio.go#L21

volvera10:
@mmontes11 is this code ref correct:
https://github.com/mariadb-operator/mariadb-operator/blob/4788a7e43380b6fb2000bbb64224e60ca8c070b6/pkg/minio/minio.go#L21

Issue #605, opened 2024-05-07
[Bug] MARIADB_ROOT_HOST environment variable is not overridden.


gsjeon:
**Documentation**
- [ ] I acknowledge that I have read the relevant [documentation](https://github.com/mariadb-operator/mariadb-operator/tree/main/docs).

**Describe the bug**
MARIADB_ROOT_HOST environment variable is not overridden.
I have to restrict access to my root account for security reasons.
The MARIADB_ROOT_HOST variable has been redefined and is still defined as **root@%**

**Expected behaviour**

It should be defined as root@10.200.0.0/255.248.0.0 rather than root@%.

**Steps to reproduce the bug**

1. define mariadb yaml
```
---
apiVersion: k8s.mariadb.com/v1alpha1
kind: MariaDB
metadata:
  name: mariadb-galera
spec:
  rootPasswordSecretKeyRef:
    name: db-sc
    key: mariadb-root-password

  storage:
    size: 100Gi
    storageClassName: ceph

  replicas: 3

  maxScale:
    enabled: false

  galera:
    enabled: true
    primary:
      podIndex: 0
      automaticFailover: true
    sst: mariabackup
    availableWhenDonor: false
    galeraLibPath: /usr/lib/galera/libgalera_smm.so
    replicaThreads: 4
    providerOptions:
      gcs.fc_limit: '64'
    agent:
      image: ghcr.io/mariadb-operator/mariadb-operator:v0.0.28
      port: 5555
      kubernetesAuth:
        enabled: false
      gracefulShutdownTimeout: 5s
    recovery:
      enabled: true
      minClusterSize: 50%
      clusterMonitorInterval: 10s
      clusterHealthyTimeout: 30s
      clusterBootstrapTimeout: 10m
      podRecoveryTimeout: 3m
      podSyncTimeout: 3m
    initContainer:
      image: ghcr.io/mariadb-operator/mariadb-operator:v0.0.28
    config:
      reuseStorageVolume: true

  podSecurityContext:
    runAsUser: 0

  volumes:
    - name: localtime
      hostPath:
        path: /etc/localtime
    - name: tls
      secret:
          secretName: mariadb-galera-tls

  volumeMounts:
    - name: localtime
      mountPath: /etc/localtime
      readOnly: true
    - name: tls
      mountPath: /certs

  service:
    type: ClusterIP

  primaryService:
    type: ClusterIP

  secondaryService:
    type: ClusterIP

  tolerations:
[...]

PR #711, opened 2024-07-02
Prevent duplicated environment variable definitions

mbezhanov:
This is a PR to address the issue reported in #605.

I am intentionally indexing the positions of the `EnvVar` elements in the `env` slice by name, as this allows me to retain the order of elements when performing the overrides.

I'll be happy to tweak this further if needed.


mmontes11 (approved):
LGTM! Thanks a lot for your contribution and your in-depth investigation in https://github.com/mariadb-operator/mariadb-operator/issues/605 🙏🏻 🦭

Issue #685, opened 2024-06-13
[Feature] modifying users?


pasztorl:
If I create an User object the maxUserConnections become immutable. So If I want to adjust this setting I need to recreate the user. I think this is not so good in prod environment because when I delete the user the application can't connect to it, so it will be a downtime while the new user not created by the operator. So can you please consider to manage the users more dynamically?
Thanks!

mmontes11:
Hey there!

Fair enough! We have significantly improved the SQL object reconciliation in the last release, managing the updates of the `User` password, we could extend the `ALTER USER` statement used for reconciliation and also update the max connections.

For making this usable, we also need to make the `maxConnections` mutable, as you pointed out.

Contributions welcome!

AayushVinayDev:
Hi @mmontes11 @pasztorl can i contribute to this?

mmontes11:
Thanks for your [PR](https://github.com/mariadb-operator/mariadb-operator/pull/723) @AayushVinayDev ! I've left some comments

AayushVinayDev:
@mmontes11 Heyy!! Sorry about the delay, was busy with finals. I've made some changes as suggested, I'll commit them and write the tests when given the green signal. : )

PR #723, opened 2024-07-17
made changes to AlterUser to accept maxUserConnections

AayushVinayDev:
This is a PR to address the issue reported in #685.

To handle the users more dynamically, I've extended the AlterUser() to accept maxUserConnections. For this I've also made maxUserConnections mutable.

 Please let me know of my shortcomings :D

I'll be happy to tweak this further if needed.


api/v1alpha1/user_types.go
```diff
@@ -29,7 +29,7 @@ type UserSpec struct {
 	// +optional
 	// +kubebuilder:default=10
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:number"}
-	MaxUserConnections int32 `json:"maxUserConnections,omitempty" webhook:"inmutable"`
+	MaxUserConnections int32 `json:"maxUserConnections,omitempty"` //webhook:"inmutable"`
```
mmontes11, line 32:
```suggestion
	MaxUserConnections int32 `json:"maxUserConnections,omitempty"`
```

internal/controller/user_controller.go
```diff
@@ -132,8 +132,8 @@ func (wr *wrappedUserReconciler) Reconcile(ctx context.Context, mdbClient *sqlCl
 		if err := mdbClient.CreateUser(ctx, accountName, createUserOpts...); err != nil {
 			return fmt.Errorf("error creating User: %v", err)
 		}
-	} else if password != "" {
-		if err := mdbClient.AlterUser(ctx, username, password); err != nil {
+	} else if password != "" || wr.user.Spec.MaxUserConnections > 0 {
+		if err := mdbClient.AlterUser(ctx, username, password, wr.user.Spec.MaxUserConnections); err != nil {
```
mmontes11, line 136: What if the password is non empty but the connections are a zero value? My suggestion is that `AlterUser` takes this into account and avoids altering the `MAX_USER_CONNECTIONS` whenever the maxconnections param is 0.

pkg/sql/sql.go
```diff
@@ -264,8 +264,9 @@ func (c *Client) DropUser(ctx context.Context, accountName string) error {
 	return c.ExecFlushingPrivileges(ctx, query)
 }
 
-func (c *Client) AlterUser(ctx context.Context, accountName, password string) error {
-	query := fmt.Sprintf("ALTER USER '%s' IDENTIFIED BY '%s';", accountName, password)
+func (c *Client) AlterUser(ctx context.Context, accountName, password string, maxConnections int32) error {
```
mmontes11, line 267: I suggest the following signature:

```go
type UserOpts struct {
	IdentifiedBy       string
	MaxUserConnections int32
}

type UserOpt func(*UserOpts)

func WithIdentifiedBy(password string) UserOpt {
	return func(cuo *UserOpts) {
		cuo.IdentifiedBy = password
	}
}

func WithMaxUserConnections(maxConns int32) UserOpt {
	return func(cuo *UserOpts) {
		cuo.MaxUserConnections = maxConns
	}
}

func (c *Client) AlterUser(ctx context.Context, accountName string, userOpts ...UserOpt) error
```

Basically re-use the same options as [CreateUser](https://github.com/mariadb-operator/mariadb-operator/blob/4788a7e43380b6fb2000bbb64224e60ca8c070b6/pkg/sql/sql.go#L240) for consistency.

mmontes11: Options should only by set if non empty, like we do in [CreateUser](https://github.com/mariadb-operator/mariadb-operator/blob/4788a7e43380b6fb2000bbb64224e60ca8c070b6/pkg/sql/sql.go#L240)

pkg/controller/replication/config.go
```diff
@@ -207,7 +208,8 @@ func (r *ReplicationConfig) reconcileUserSql(ctx context.Context, mariadb *maria
 		return fmt.Errorf("error checking if replication user exists: %v", err)
 	}
 	if exists {
-		if err := client.AlterUser(ctx, opts.username, replPassword); err != nil {
+		maxUserConnections := user.Spec.MaxUserConnections
+		if err := client.AlterUser(ctx, opts.username, replPassword, maxUserConnections); err != nil {
```
mmontes11, line 212: See my previous comment. This should pass the options like we do for `CreateUser`.

mmontes11 (changes requested):
Thanks a lot for your contribution! 🙏🏻 

You are on the right track to implement this. I've left a couple of indications. Let me know if you need further help.

mmontes11:
@AayushVinayDev also, it would be great if you could add a test to prove that this functionallity works as expected.

internal/controller/user_controller.go
AayushVinayDev: Understood!! :D , I've configured `AlterUser` to take this into account and changed above code accordingly. 
```
else if password != "" || wr.user.Spec.MaxUserConnections > 0 {
		if err := mdbClient.AlterUser(ctx, username, sqlClient.WithIdentifiedBy(password), sqlClient.WithMaxUserConnections(wr.user.Spec.MaxUserConnections)); err != nil {
			return fmt.Errorf("error altering User: %v", err)
		}
	}
```

Please let me know if there is any change needed. :)

pkg/sql/sql.go
AayushVinayDev: Thanks for the insight! I've tried to re-use the same options as `CreateUserOpts{}`, it also takes into account to avoid altering the `MAX_USER_CONNECTIONS` whenever the `maxconnections` param is 0.
```
func (c *Client) AlterUser(ctx context.Context, accountName string, createUserOpts ...CreateUserOpt) error {
	opts := CreateUserOpts{}
	for _, setOpt := range createUserOpts {
		setOpt(&opts)
	}

	query := fmt.Sprintf("ALTER USER '%s' ", accountName)
	if opts.IdentifiedBy != "" {
		query += fmt.Sprintf("IDENTIFIED BY '%s' ", opts.IdentifiedBy)
	}
	if opts.MaxUserConnections > 0 {
		query += fmt.Sprintf("WITH MAX_USER_CONNECTIONS %d ", opts.MaxUserConnections)
	}
	query += ";"

	return c.ExecFlushingPrivileges(ctx, query)
}
```

pkg/controller/replication/config.go
AayushVinayDev: Let me know if any change is needed. :D
```
if exists {
		if err := client.AlterUser(ctx, opts.username, sqlClient.WithIdentifiedBy(replPassword)); err != nil {
			return fmt.Errorf("error altering replication user: %v", err)
		}
	}
```

PR #721, opened 2024-07-15
Support history limits in `Backup` and `SqlJob` resources

mbezhanov:
As requested in #641, this PR enables the `successfulJobsHistoryLimit` and `failedJobsHistoryLimit` fields for scheduled `Backup` and `SqlJob` resources.

Is the test coverage sufficient or should I add some more tests?

Close https://github.com/mariadb-operator/mariadb-operator/issues/641

Issue #383, opened 2024-02-13
[Feature] Add timezone value to Backup resources.


codestation:
**Is your feature request related to a problem? Please describe.**

When configuring a Backup resource the created CronJob will always be run in the UTC timezone.  One cannot choose their prefered timezone.

**Describe the solution you'd like**

Since Kubernetes 1.27 there is a `.spec.timeZone` value available to CronJobs. Make this available to all the resources with a `schedule` field (like `Backup`).

**Describe alternatives you've considered**

**Environment details**:
- Kubernetes version: 1.28
- Kubernetes distribution: k3s
- mariadb-operator version: 0.0.25
- Install method: helm
- Install flavor: recommended

**Additional context**

mmontes11:
Hey @codestation ! Thanks for bringing this up.

This will be a nice addition for the `Backup` and `SqlJob` resources, which are reconciled into `CronJobs as you correctly pointed out.

> Since Kubernetes 1.27 

I'm guessing that we will need to use the discovery client to check whether we are running in Kubernetes >= 1.27: 
https://pkg.go.dev/k8s.io/client-go/discovery 

Contributions are welcome!

mbezhanov:
> Since Kubernetes 1.27 there is a .spec.timeZone value available to CronJobs.

Kubernetes 1.27 is when the feature graduated from `Beta` to `GA`. However, the `timeZone` field has been available since Kubernetes 1.24, with 1.25 marking it as `Beta` and enabling its feature gate by default (meaning that since 1.25, administrators had to explicitly disable this feature if they didn't want their clusters to support it).

See: https://github.com/kubernetes/enhancements/issues/3140

Take the following `CronJob` manifest, for example: <details><summary><b>cronjob.yaml</b></summary>

```yaml
apiVersion: batch/v1
kind: CronJob
metadata:
  name: example-cronjob
spec:
  schedule: "0 12 * * *" # Every day at noon
  timeZone: "America/New_York" # Specify the time zone
  jobTemplate:
    spec:
      template:
        spec:
          containers:
          - name: example
            image: busybox
            args:
            - /bin/sh
            - -c
            - date; echo "Hello from the Kubernetes CronJob!"
          restartPolicy: OnFailure
```

</details>

Even with the `CronJobTimeZone` feature gate disabled, no errors will be thrown on Kubernetes versions >= 1.24 because of a `timeZone` field being present in the `CronJob` manifest, as the field already exists in the `CronJob` schema definition.

[...]

PR #724, opened 2024-07-18
Fix reconciliation of SqlJob history limit updates

mbezhanov:
While looking into #383, I found out that in #721 I failed to notice that for `SqlJob` resources, the `CronJob` reconciliation is handled in the `SqlJobReconciler` and not the `BatchReconciler`.

Because of this, once the initial history limits for an `SqlJob` are established, they cannot be subsequently updated. This is really inconvenient, and I am sorry about this!

This PR provides a fix and additional tests that cover the regression.

### DIFF
--- a/internal/controller/sqljob_controller.go
+++ b/internal/controller/sqljob_controller.go
@@ -235,6 +235,8 @@ func (r *SqlJobReconciler) reconcileCronJob(ctx context.Context, sqlJob *mariadb
 	}
 
 	patch := client.MergeFrom(existingCronJob.DeepCopy())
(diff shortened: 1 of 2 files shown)


mbezhanov:
Happy to hear I was able to fix it on time! Side question: should `Restore` support history limits as well? I can certainly add this to my backlog.

mmontes11:
> should Restore support history limits as well? I can certainly add this to my backlog.

No, the `Restore` CR never gets reconciled into a `CronJob`, it is always reconciled into a one-off `Job`. Therefore, this fields don't apply here AFAIK.

PR #741, opened 2024-07-23
Add suspend spec to mariadb kind

harunkucuk5:
This PR adds `suspend` spec to the mariadb kind which is useful during maintenance when you do not want the operator to touch mariadb clusters.

```bash
 kind: MariaDB        
...
   suspend: true                                                                                                                                                                                                 
   updateStrategy:                                                                                                                                                                                               
     type: ReplicasFirstPrimaryLast
...
```


api/v1alpha1/mariadb_types.go
```diff
@@ -461,6 +461,11 @@ type MariaDBSpec struct {
 	// +optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	SecondaryConnection *ConnectionTemplate `json:"secondaryConnection,omitempty" webhook:"inmutable"`
+	// Suspend defines whether the operator manages the mariadb or not.
```
mmontes11, line 464:
```suggestion
	// Suspend defines whether the MariaDB reconciliation loop is enabled or not. This can be useful for maintenance.
```

mmontes11 (changes requested):
Thanks for the contribution @harunkucuk5 ! This makes total sense to me. 

In the context of managing a fleet of databases in Kubernetes, where some of them might be suspended, it would be very useful to display the `Suspended` status in the `Status` custom column. This can be done by updating this function:

https://github.com/mariadb-operator/mariadb-operator/blob/2e776fcd623d87a032915d4b80ad857d91a0b7fe/pkg/condition/ready.go#L75

What do you think?

Other than that LGTM! Just added a minor code suggestions.

harunkucuk5:
> Thanks for the contribution @harunkucuk5 ! This makes total sense to me.
> 
> In the context of managing a fleet of databases in Kubernetes, where some of them might be suspended, it would be very useful to display the `Suspended` status in the `Status` custom column. This can be done by updating this function:
> 
> https://github.com/mariadb-operator/mariadb-operator/blob/2e776fcd623d87a032915d4b80ad857d91a0b7fe/pkg/condition/ready.go#L75
> 
> What do you think?
> 
> Other than that LGTM! Just added a minor code suggestions.

Thanks @mmontes11 for your review. I tried to show the suspended status with [965f24b](https://github.com/mariadb-operator/mariadb-operator/pull/741/commits/965f24b378cce85b3b93412cdeae56f645639969). However it did not work. Probably I am missing something. Even if I suspend a mariadb I still see the following status.

image

internal/controller/mariadb_controller.go
```diff
@@ -114,6 +114,12 @@ func (r *MariaDBReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
+	if mariadb.Spec.Suspend {
+		log.FromContext(ctx).V(1).Info("%s suspended. Skipping...", mariadb.Name)
+		condition.SetReadyWithMariaDB(&mariadb.Status, nil, &mariadb)
+		return ctrl.Result{}, nil
+	}
```
mmontes11, line 121:
```suggestion
	if mariadb.IsSuspended() {
		log.FromContext(ctx).V(1).Info("MariaDB is suspended. Skipping...")
		return ctrl.Result{}, r.patchStatus(ctx, &mariadb, func(status *mariadbv1alpha1.MariaDBStatus) error {
			condition.SetReadyWithMariaDB(&mariadb.Status, nil, &mariadb)
			return nil
		})
	}
```

api/v1alpha1/mariadb_types.go
mmontes11, line 464:
```suggestion
	// Suspend defines whether the MariaDB reconciliation loop is enabled. This can be useful for maintenance, as disabling the reconciliation loop prevents the operator from interfering with user operations during maintenance activities.
```

mmontes11 (changes requested):
Hey @harunkucuk5 ! 

I've added a code suggestion to show the `Suspended` status, please take a look.

I have a couple more asks if you don't mind:
- Take a look at code suggestion for the `suspend` field documentation
- Run `make gen` to regenerate the CRDs
- Add a test [here](https://github.com/mariadb-operator/mariadb-operator/blob/2e776fcd623d87a032915d4b80ad857d91a0b7fe/internal/controller/mariadb_controller_test.go#L57) to ensure that the functionality works. You have a couple of examples there. It could be something as simple as:
    - Create a `MariaDB` resource with `suspend=true`
    - Veriry that the the `Ready` condition is `false` with the reason `Suspended`

api/v1alpha1/mariadb_types.go
```diff
@@ -679,6 +684,11 @@ func (m *MariaDB) IsUpdating() bool {
 	return condition.Status == metav1.ConditionFalse && condition.Reason == ConditionReasonUpdating
 }
 
+// IsUpdating indicates that a MariaDB update is in progress.
```
mmontes11, line 687:
```suggestion
// IsSuspended indicates that a MariaDB update is suspended.
```

mmontes11 (changes requested):

iammuho:
Hey @harunkucuk5,

I love that idea. Just I wonder rather than we introduce a new type of field, should we use the status spec as other condition type/reasons ? 

```
// IsSuspended indicates that a MariaDB is suspended
func (m *MariaDB) IsSuspended() bool {
	condition := meta.FindStatusCondition(m.Status.Conditions, ConditionTypeSuspended)
	if condition == nil {
		return false
	}
	return condition.Status == metav1.ConditionTrue
}
```

Not a direct advice, just opening a discussion. What do you think @mmontes11 ?

mmontes11:
Hey there @iammuho ! Thanks for your input, much appreciated.

The question I normally ask myself before introducing a new status condition is: Do I need to abstract a specific state in the CR so other controllers can act based on this condition instead of inspecting the state themselves again?

> if we want to do some operations while "suspending in progress" .

Or as you said, there is a more complex logic, sometimes involving multiple states, that needs to be tracked?

In both of the cases above, the answer today would be no, since the process is pretty simple at the moment.

> block other activities while suspending.

We are already blocking other activities, as this is returning early on the reconciliation loop.

> That will solve your problem to show the "suspended" status 

Actually no, becuase the status column is currently mapping the `Ready` condition, not the `Suspended` condition.

Considering all this, I would advocate for the simplest approach for now, but we will keep in mind your pointers for the future @iammuho , thanks again for your input here!

iammuho:
Hey @mmontes11 ,

Thanks for clear message, I guess I understood your approach. Just to follow up, introducing the "suspended" as another condition status is not a clear way of solving that original request rather than a custom field ?

```
ConditionReasonInitializing        string = "Initializing"
	ConditionReasonInitialized         string = "Initialized"
	ConditionReasonPendingUpdate       string = "PendingUpdate"
	ConditionReasonUpdating            string = "Updating"
	ConditionReasonUpdated             string = "Updated"
	ConditionReasonSuspended           string = "Suspended"
```

As @harunkucuk5 already defined the condition "reason" suspended, wouldn't it be better to introduce the type too ? I still believe, introducing the conditionType == suspended is much simpler option that that many new tests + field . What do you think ?

harunkucuk5:
> Hey @harunkucuk5 !
> 
> I've added a code suggestion to show the `Suspended` status, please take a look.
> 
> I have a couple more asks if you don't mind:
> 
> * Take a look at code suggestion for the `suspend` field documentation
> * Run `make gen` to regenerate the CRDs
> * Add a test [here](https://github.com/mariadb-operator/mariadb-operator/blob/2e776fcd623d87a032915d4b80ad857d91a0b7fe/internal/controller/mariadb_controller_test.go#L57) to ensure that the functionality works. You have a couple of examples there. It could be something as simple as:
>   
>   * Create a `MariaDB` resource
>   * Ensure that it eventually becomes ready
>   * Update `spec.suspend=true`
>   * Verify that the the `Ready` condition is `false` with the reason `Suspended`

@mmontes11, thanks for your suggestions. it is working as expected now. I have added the integration test for the suspend case. Please have another look

mmontes11 (approved):
@harunkucuk5 thanks again for your contribution! This feature will definetly help managing large fleets of `MariaDBs`! 🙏🏻 🦭 

LGTM! 🚀

mmontes11:
@iammuho We really appreciate your appreciate your input and we will keep it in mind for future maintenance / refactors!

mmontes11:
This was very inspiring, I think it will be worth supporting it in `MaxScale`: 
- https://github.com/mariadb-operator/mariadb-operator/issues/769

Issue #749, opened 2024-07-26
[Feature] Support for `extraEnvFrom` in the helm chart


mmontes11:
**Describe the solution you'd like**
Currently, we support [`extraEnv`](https://github.com/mariadb-operator/mariadb-operator/blob/46ef148a2a8bb69b337dd93fccda50cffb6dddd7/deploy/charts/mariadb-operator/values.yaml#L63) for passing extra environment variables to the operator `Deployment`. 

Instead of passing each env variable individually, be able to pass multiple variables by referencing a `ConfigMap` or a `Secret`:
```yaml
extraEnvFrom:
   - configMapRef:
       name: mariadb-operator-env-extra
```

mmontes11:
@gprossliner that would be awesome! 🙏🏻 🙏🏻 🙏🏻 Please go ahead and let me know if you need any help!

mmontes11:
This has been addressed by @gprossliner via https://github.com/mariadb-operator/mariadb-operator/pull/766 ! Thanks for the contribution 🙏🏻 🦭

PR #766, opened 2024-08-01
HelmChart: extraEnvFrom for controller

gprossliner:
see #749 

This PR consists of two commits.

The first commit is the actual implementation. I tested it manually by using different values files
The second commit is a unit test for this feature. Please let me know if you are interested in unit tests to be included for the Helm chart. If not I can remove this commit or make a separate PR for this.

Some notes regarding the unit test:
* I didn't want to put ini directly in controller so I created a new directory `internal/helmtest`
* It uses the https://pkg.go.dev/github.com/gruntwork-io/terratest@v0.47.0/modules/helm package to render templates for testing. This runs the helm command, which needs to be installed. This may cause the CI to fail.
* I wanted to write the test using `ginko/v2` with `Descibe`... but because the `terratest` library needs the `*testing.T` instance I used go-testing directly. I think that there should be a way to get `t` with ginko, but I first want to know if testing should be included at all.


mmontes11 (approved):
Hey there @gprossliner ! Thanks a lot for this huge contribution! 🙏🏻 

LGTM! 💯 

I really like the approach you followed for doing helm chart unit testing, it is something definetly missing that will avoid regressions in helm templates. 

> It uses the https://pkg.go.dev/github.com/gruntwork-io/terratest@v0.47.0/modules/helm package to render templates for testing. This runs the helm command, which needs to be installed. This may cause the CI to fail.

I didn't knew `terratest` before, but it seems to be a commonly used as library for testing:
- https://medium.com/@zelldon91/advanced-test-practices-for-helm-charts-587caeeb4cb

I used to use [helm-unittest](https://github.com/helm-unittest/helm-unittest) in the past, where you basicaly declare tests and assertions in yaml files, but the ability to write them in Go is much better and easier to maintain. Thank you very much!

> I wanted to write the test using ginko/v2 with Descibe... but because the terratest library needs the *testing.T instance I used go-testing directly. I think that there should be a way to get t with ginko, but I first want to know if testing should be included at all.

I was looking into it, but there doesn't seem to be a straightforward way of achieving this, only some hack-arounds:
- https://github.com/onsi/ginkgo/issues/9

So I would suggest to leave it as it is unless you have something in mind, but we can tackle this in a separate PR.

P.S. I will create an issue to complete the helm unit tests.

mmontes11:
@gprossliner again, thanks for this contribution! 🙏🏻  I have created the following issue to complete the helm unit tests:
- https://github.com/mariadb-operator/mariadb-operator/issues/767

gprossliner:
OK. If I have some time to spent I can also add some of the unit tests for the chart.

PR #776, opened 2024-08-06
Add support for specifying password hashes for users

NHellFire:
This adds support for providing user passwords as a hash (for the default authentication plugin), or a named authentication plugin with optional arguments.

I have tested this with the following Secret:
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: mariadb-test
stringData:
  # Both passwords are "password"
  ed25519Hash: RBIiSdwSgBldhFZLFzRCTTc1Q0pRJ3pFZudlQ78k40o
  ed25519Plugin: ed25519
  passwordHash: "*2470C0C06DEE42FD1618BB99005ADCA2EC9D1E19"
```

And Users:
```yaml
kind: User
metadata:
  name: user
spec:
  mariaDbRef:
    name: mariadb
  passwordHashSecretKeyRef:
    name: mariadb-test
    key: passwordHash
---
kind: User
metadata:
  name: user2
spec:
  mariaDbRef:
    name: mariadb
  passwordViaSecretKeyRef:
    name: mariadb-test
    key: ed25519Plugin
  passwordViaUsingSecretKeyRef:
    name: mariadb-test
    key: ed25519Hash
```

Tests need adding, but I could use some help with that as I'm not sure how to add them to the existing user_controller tests.


mmontes11 (changes requested):
Hey there @NHellFire ! Thanks a lot for this contribution, I think you are on the right track. Here some comments that may help you:

I do agree with the new `passwordHashSecretKeyRef` field naming, but I would suggest creating more explicit naming for the plugin based authentication. Something that doesn't require previous SQL syntax knowledge (i.e. `VIA` and `USING`):

```yaml
kind: User
metadata:
  name: user2
spec:
  mariaDbRef:
    name: mariadb
  passwordPlugin:
       pluginNameSecretKeyRef:
           name: mariadb-test
           key: ed25519Plugin
       pluginArgSecretKeyRef:
            name: mariadb-test
            key: ed25519Hash
``` 
What do you think? I think grouping them togeter makes the API more clear in this case. Also we will need to add validation logic and tests in the webhook here:
- https://github.com/mariadb-operator/mariadb-operator/blob/main/api/v1alpha1/user_webhook.go

Basically, we will need to validate that `passwordSecretKeyRef`, `passwordHashSecretKeyRef` and `passwordPlugin` are mutually exclusive.

Regarding the tests, we will need to at least add a test case for `passwordHashSecretKeyRef` and `passwordPlugin`: Create a `User` resource and make sure it becomes `Ready`, validating that the query executed by the controller was successful. You have some examples here:
- https://github.com/mariadb-operator/mariadb-operator/blob/c1a463ec3da07715b634b9c1f825ebd75702b0be/internal/controller/user_controller_test.go#L125

Once the `User` has been created, we could attempt to connect using the hash and plugin credentials, but I think this is not supported in the go mysql driver:
- https://pkg.go.dev/github.com/go-sql-driver/mysql@v1.8.1#Config

Also, adding units tests for `pkg/sql` wouldn't hurt, specially now that we have multiple conditionals. 

Finally, it would be great to support the password hash authentication for the root an initial users. On the same line, we would need to extend the `MariaDB` CR and make use of the environment variables supported by the image:
- https://mariadb.com/kb/en/mariadb-server-docker-official-image-environment-variables/

Plugin authentication doesn't seem to be supported by the env vars, only via hash. I'm well aware that this is a lot of work. So feel free to open a different issue for the root and initial user.

Let me know if you need further help!

NHellFire:
Thanks for the feedback, Martin!
I think I've got everything done. Including making all the password methods mutually exclusive, and adding tests for password hash + plugin. All tests still pass, I should be able to push it later today.


> Finally, it would be great to support the password hash authentication for the root an initial users.

Would that actually work? Is the management done using the root password or is a separate privileged user created at bootstrap?
I have looked at the container and bootstrap runs with no root password until it's completed (and upgrades run with `skip-grant-tables`)

mmontes11:
Thanks for the effort @NHellFire ! I will take another look once you push your changes. Also make sure you rebase with `main`.

> Would that actually work? Is the management done using the root password or is a separate privileged user created at bootstrap?

The management done by the operator currently uses the `root` credentials provisioned by the image entrypoint via the `MARIAB_ROOT_PASSWORD` environment variable. Using separate privileged user would require in-depth analysis, we can discuss this in a different issue.  

For now, my suggestion is supporting the `passwordPlugin` for the initial user defined in the `MariaDB` CR via the `username` and `passwordSecretKeyRef` fields. We would need to introduce the `passwordPlugin` field as we described [here](https://github.com/mariadb-operator/mariadb-operator/pull/776#pullrequestreview-2225606118) in the `MariaDB` CR as well. This will just be propagated to the initial `User` resource created here:
- https://github.com/mariadb-operator/mariadb-operator/blob/f2a1672c33d3554a92550e9eb267b4ad2fbd439a/internal/controller/mariadb_controller.go#L773

I think it would be straightforward as it is just mapping the field from `MariaDB` from `User`, but I leave it up to you to tackle this in a different issue/PR. The only limitation I can think of is that we may not be able to check the `Connection` health here:
- https://github.com/mariadb-operator/mariadb-operator/blob/f2a1672c33d3554a92550e9eb267b4ad2fbd439a/internal/controller/connection_controller.go#L307

Since the go mysql driver does not support authentication via these plugins:
- https://pkg.go.dev/github.com/go-sql-driver/mysql@v1.8.1#Config

Or maybe I am missing something? 🤔 If that was the case, we could document this limitation properly and omit the healthcheck an auth plugin is used. Thoughts?

NHellFire:
Rebased and pushed now. Had less free time over the weekend than I expected.

> Since the go mysql driver does not support authentication via these plugins:
> 
>     * https://pkg.go.dev/github.com/go-sql-driver/mysql@v1.8.1#Config
> 
> 
> Or maybe I am missing something? 🤔 If that was the case, we could document this limitation properly and omit the healthcheck an auth plugin is used. Thoughts?

The Go driver does have support for [some auth plugins](https://github.com/go-sql-driver/mysql/blob/2f1527670cb7207fd213f92c7120f9387fe256cf/auth.go#L277), so maybe an option to disable the healthcheck instead?

mmontes11:
No prob, thanks for the great effort @NHellFire ! Very much appreciated ! 🙏🏻 

> The Go driver does have support for [some auth plugins](https://github.com/go-sql-driver/mysql/blob/2f1527670cb7207fd213f92c7120f9387fe256cf/auth.go#L277), so maybe an option to disable the healthcheck instead?

We have support to disable the healthcheck already,  just by setting the `Connections`'s field `healthCheck` to `nil`: 

https://github.com/mariadb-operator/mariadb-operator/blob/b5836f17930f3010bd9313d25a949cdd574f6559/internal/controller/connection_controller.go#L303

This should be the case whenever we think the go driver might not manage to use the plugins.

docs/API_REFERENCE.md
```diff
@@ -1158,6 +1158,23 @@ _Appears in:_
 
 
 
+#### PasswordPluginSpec
```
mmontes11, line 1161: Could we undo this? Just run `make gen`. I prefer not to publish docs with unreleased fields to avoid confussion.

I will take care of generating the docs at the end of the release.

mmontes11 (approved):
LGTM! 🚀 Thanks for this massive contribution @NHellFire! 👍🏻 💯  

I will take care of the following items:
- Support for password hashes and plugins in the initial users specified in `MariaDB`.  Basically add `passwordHashSecretKeyRef` and `passwordPlugin` fields that will be mapped to the initial `User` resource
- Regarding the `Connection` `healthCheck`. I will document that only [some plugins ](https://github.com/go-sql-driver/mysql/blob/2f1527670cb7207fd213f92c7120f9387fe256cf/auth.go#L277) are supported by the Go driver. The `healthCheck` needs to explicitly enabled by the end user, so I don't think this would be a big deal if properly documented.
- Add documentation for the password hashes and plugins, linking the [MariaDB docs](https://mariadb.com/kb/en/authentication-plugins/) for further detail.

Again, thank you very much for this contribution! This will be released in `v0.0.30`, stay tuned!

Issue #507, opened 2024-03-26
[Feature] Point In Time Recovery (PITR)


mmontes11:
**Describe the solution you'd like**
Ability to restore the state of `MariaDB` in a particular point in time, minimizing the [RPO](https://aws.amazon.com/blogs/mt/establishing-rpo-and-rto-targets-for-cloud-applications/#:~:text=As%20a%20quick%20refresher%2C%20RTO,loss%20your%20application%20can%20tolerate.) (data loss) and  [RTO](https://aws.amazon.com/blogs/mt/establishing-rpo-and-rto-targets-for-cloud-applications/#:~:text=As%20a%20quick%20refresher%2C%20RTO,loss%20your%20application%20can%20tolerate.) (time to recover) as much as possible.

For achieving this we will need to combine:
- **Full physical backups**: It is a snapshot of the current database state files kept in an object store (S3 compatible). Periodically taken by a `CronJob` that mounts the `MariaDB` PVC. 
- **Binlog archival**: Binary log files that contain the database events that happened after the full physical backup was taken. Making use of parallelism, our [sidecar agent](https://github.com/mariadb-operator/mariadb-operator/blob/main/cmd/agent/main.go) will periodically archive the binlog files in an object store (S3 compatible).

It is important to note that the initial restoration/bootstrapping process happens offline before the `MariaDB` instance is provisioned, so there is no impact in the database performance nor its availability. This will be done by either an init `Job` (Galera) or an init container on each instance (replication).

Initially, the plan is to use this disaster recovery strategy for Galera, as we are already making use of the agent, required for log archival. However, the replication architecture could catch up with Galera and make use of the agent so the PITR can also be performed. Not only this would be beneficial for backing up and restore replication clusters, but also for spining up new replicas or restoring existing ones.  We currently have multiple issues opened where either the replica is in a bad state and needs to be restored or the master has purged binary logs and it is not possible to perform the full replication:
- https://github.com/mariadb-operator/mariadb-operator/issues/406
- https://github.com/mariadb-operator/mariadb-operator/issues/246
- https://github.com/mariadb-operator/mariadb-operator/issues/141

Regarding the physical backups with `maria-backup`, this has been attempted already and there is a WIP PR:
- https://github.com/mariadb-operator/mariadb-operator/pull/273

Also, this PITR feature would superseed the initial strategy we considered for physical backups with `maria-backup`: 
- https://github.com/mariadb-operator/mariadb-operator/issues/11

[...]

PR #779, opened 2024-08-09
Add TimeZone spec to `Backup` and `SqlJob` resources

mbezhanov:
Closes https://github.com/mariadb-operator/mariadb-operator/issues/383.

This PR implements the request from #383.

It also introduces a couple of changes that we agreed upon in previous PRs:

1. Addition of a `CronJobTemplate` type to encapsulate the `TimeZone`, `SuccessfulJobsHistoryLimit`, and `FailedJobsHistoryLimit` fields.
2. Heavy refactoring of `backup_controller_test.go` and `sqljob_controller_test.go` to cover additional test cases and reduce code duplication.

I was also thinking about refactoring `SqlJob` to use `BatchReconciler` in this PR, but the PR became kind of large, so I'll leave it for some other time if that's okay.

A few things to point out:

- I haven't run the tests on anything other than Kubernetes 1.30 yet, but plan on doing so in the following days.
- I intentionally left the individual commits to (hopefully) make the review a bit easier, but I'm ready to squash them before merging.
- I can further split the refactoring changes into a separate PR if preferred.

I remain available if any additional changes are necessary.


mbezhanov:
By the way, I had to make some manual tweaks to the KIND cluster to get the integration tests running locally (had to manually provision a MinIO access key and then create the respective Kubernetes secret for use with the S3 bucket tests).

Is there a procedure that provisions this automatically?

mbezhanov:
Tested on all versions between (and including) `1.23` and `1.30`. `TimeZone` field is silently dropped on `1.23` and `1.24` and successfully set on the rest as expected.

The refactored integration tests run successfully on everything >= `1.25`. They fail on <= `1.24`, but only on the `TimeZone` assertions, as the field is ignored, so they are never met (this is expected).

Everything looks good on my end. The PR is ready for review.

internal/controller/backup_controller_test.go
```diff
@@ -18,81 +20,135 @@ var _ = Describe("Backup", func() {
...
+		),
+		Entry("should reconcile a CronJob with PVC storage and history limits",
+			"backup-pvc-scheduled-with-limits-test",
+			applyDecoratorChain[*mariadbv1alpha1.Backup](
+				getBackupWithPVCStorage,
+				decorateBackupWithSchedule,
+				decorateBackupWithHistoryLimits,
+			),
+			testBackup,
+		),
```
mmontes11, line 85: This is brilliant! ✨

internal/controller/backup_controller_test.go
```diff
@@ -18,81 +20,135 @@ var _ = Describe("Backup", func() {
...
-		key := types.NamespacedName{
-			Name:      "backup-s3-scheduled-test-prefix",
-			Namespace: testNamespace,
-		}
-		backup := getBackupWithScheduleAndHistoryLimits(
-			getBackupWithS3Storage(key, "test-backup", "mariadb"),
-		)
-		testS3Backup(backup)
-	})
+	DescribeTable("Creating a Backup",
```
mmontes11, line 23: Much clear and readable 🌟

mmontes11 (approved):
Thank you very much for this brilliant contribution! ✨ 🌟 

LGTM! 🚀 

The refactor is massively helpful, not only because it provides re-usability and more clarity, but also because we will be having more `CronJob` objects in the future to implement scheduled physical backups for [PITR](https://github.com/mariadb-operator/mariadb-operator/issues/507)! Thank you very much for taking the time, it is very much appreciated!

I will be documenting this together with the [`MariaDB` timeZone support](https://github.com/mariadb-operator/mariadb-operator/pull/726) and release it in `v0.0.30`. Stay tuned! 

> The refactored integration tests run successfully on everything >= 1.25. They fail on <= 1.24, but only on the TimeZone assertions, as the field is ignored, so they are never met (this is expected).

Thank you for the effort in testing with multiple Kubernetes versions! At some point, we will implement end-to-end (e2e) matrix testing to be executed before the release to ensure we don't miss any version-related regressions. We will focus on non-EOL versions of Kubernetes, as listed here:
- https://endoflife.date/kubernetes
Therefore, the timeZone tests not working on versions <= 1.24 shouldn't be a concern, as these versions are quite old and have been EOL for a while.

PR #783, opened 2024-08-12
validation rule added for an odd mariadb.spec.replicas value

businessbean:
- [CEL](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation-rules) based validation rule used
- to lower the risk of split brain situations for a Galera cluster
- `ReplicasAllowEvenNumber` option added to disable the validation
- ~~manifest example extended~~
- I have used these steps to generate the `k8s.mariadb.com_mariadbs.yaml` file
```shell
make gen
```
This is the message returned by kubectl if the then the validation fails (for instance if replicas: 2 has been defined):
`Error: UPGRADE FAILED: failed to create resource: MariaDB.k8s.mariadb.com "mariadb" is invalid: spec: Invalid value: "object": An odd number of MariaDB instances (mariadb.spec.replicas) is required to avoid split brain situations. Use 'mariadb.spec.replicasAllowEvenNumber: true' to disable this validation.`

### DIFF
--- a/api/v1alpha1/mariadb_types.go
+++ b/api/v1alpha1/mariadb_types.go
@@ -428,6 +428,10 @@ type MariaDBSpec struct {
 	// +kubebuilder:default=1
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:podCount"}
 	Replicas int32 `json:"replicas,omitempty"`
+	// disables the validation check for an odd number of replicas.
+	// +kubebuilder:default=false
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	ReplicasAllowEvenNumber bool `json:"replicasAllowEvenNumber,omitempty"`
 	// Port where the instances will be listening for connections.
 	// +optional
 	// +kubebuilder:default=3306
@@ -524,6 +528,7 @@ type MariaDB struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
+	// +kubebuilder:validation:XValidation:rule="self.replicas %2 == 1 || self.replicasAllowEvenNumber", message="An odd number of MariaDB instances (mariadb.spec.replicas) is required to avoid split brain situations. Use 'mariadb.spec.replicasAllowEvenNumber: true' to disable this validation."
 	Spec   MariaDBSpec   `json:"spec"`
 	Status MariaDBStatus `json:"status,omitempty"`
 }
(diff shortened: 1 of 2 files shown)


mmontes11 (changes requested):
Thanks for your contribution, this is great! LGTM! 👍🏻 

I can confirm it works:

```bash
❯ kubectl apply -f - <<EOF
kind: MariaDB
metadata:
  name: mariadb-galera
spec:
  rootPasswordSecretKeyRef:
    name: mariadb
    key: root-password

  storage:
    size: 1Gi

  replicas: 2

  galera:
    enabled: true

  metrics:
    enabled: true
EOF
The MariaDB "mariadb-galera" is invalid: spec: Invalid value: "object": An odd number of MariaDB instances (mariadb.spec.replicas) is required to avoid split brain situations. Use 'mariadb.spec.replicasAllowEvenNumber: true' to disable this validation.

❯ kubectl apply -f - <<EOF
kind: MariaDB
metadata:
  name: mariadb-galera
spec:
  rootPasswordSecretKeyRef:
    name: mariadb
    key: root-password

  storage:
    size: 1Gi

  replicas: 2
  replicasAllowEvenNumber: true

  galera:
    enabled: true

  metrics:
    enabled: true
EOF

mariadb.k8s.mariadb.com/mariadb-galera created
```

> make docker-build
make docs-api
make manifests-bundle
make helm-gen

Yes, this is the only way to verify this is actually working (confirmed). We try to avoid updating the docs, helm chart, manifests and examples in the `main` branch and we do this in the [release branch
](https://github.com/mariadb-operator/mariadb-operator/pull/722) instead. People got confused in the past when we were generating everything in `main`, attempting to try new fields that were not released.

Could we discard the changes in `docs`, `deploy` (except the CRDs) and `examples`? Basically, only run `make gen`. I will take care of generating the rest myself after merging.

Thanks anyway for providing an example and make sure everything is up to date!

mmontes11 (approved):
LGTM! 🙏🏻 

Thanks for your contribution!

There is probably a lot of stuff that we can migrate from the validation webhooks to CEL, for instance:
https://github.com/mariadb-operator/mariadb-operator/blob/1ba711991deb4119b9a6e900e261e9caa63d7c3c/api/v1alpha1/mariadb_webhook.go#L43

This PR is the first step! Very much appreciated!

businessbean:
> LGTM! 🙏🏻
> 
> Thanks for your contribution!
> 
> There is probably a lot of stuff that we can migrate from the validation webhooks to CEL, for instance:
> 
> https://github.com/mariadb-operator/mariadb-operator/blob/1ba711991deb4119b9a6e900e261e9caa63d7c3c/api/v1alpha1/mariadb_webhook.go#L43
> 
> This PR is the first step! Very much appreciated!

Let me check which checks are the low hanging fruits...

PR #788, opened 2024-08-15
multiple webhook validations migrated CRD validation checks using the CEL syntax

businessbean:
To simplify the webhook code i have migrated several checks into CRD validations using the CEL syntax including support for potentially undefined YAML keys.

## validateHA function
- duplicate HA method check
- replica count check
- replica count check for disabled HA
## validateGalera function
- galera.replicaThreads check
## validateRootPassword function
- mariadb.rootEmptyPassword check

I have kept the empty functions to be able to add other checks that are not possible with the CEL feature set.

### DIFF
--- a/api/v1alpha1/mariadb_webhook.go
+++ b/api/v1alpha1/mariadb_webhook.go
@@ -250,12 +250,5 @@ func (r *MariaDB) validateUpdateStorage(old *MariaDB) error {
 }
 
 func (r *MariaDB) validateRootPassword() error {
-	if r.IsRootPasswordEmpty() && r.IsRootPasswordDefined() {
-		return field.Invalid(
-			field.NewPath("spec").Child("rootEmptyPassword"),
-			r.Spec.RootEmptyPassword,
-			"'spec.rootEmptyPassword' must be disabled when 'spec.rootPasswordSecretKeyRef' is specified",
-		)
-	}
 	return nil
 }
--- a/api/v1alpha1/mariadb_webhook.go
+++ b/api/v1alpha1/mariadb_webhook.go
@@ -93,13 +93,6 @@ func (r *MariaDB) ValidateDelete() (admission.Warnings, error) {
 }
 
 func (r *MariaDB) validateHA() error {
-	if r.IsHAEnabled() && r.Spec.Replicas <= 1 {
-		return field.Invalid(
-			field.NewPath("spec").Child("replicas"),
-			r.Spec.Replicas,
-			"Multiple replicas must be specified when 'spec.replication' or 'spec.galera' are configured",
-		)
-	}
 	return nil
 }
 
--- a/api/v1alpha1/mariadb_webhook.go
+++ b/api/v1alpha1/mariadb_webhook.go
@@ -93,13 +93,6 @@ func (r *MariaDB) ValidateDelete() (admission.Warnings, error) {
 }
 
 func (r *MariaDB) validateHA() error {
-	if !r.IsHAEnabled() && r.Spec.Replicas > 1 {
-		return field.Invalid(
-			field.NewPath("spec").Child("replicas"),
-			r.Spec.Replicas,
-			"Multiple replicas can only be specified when 'spec.replication' or 'spec.galera' are configured",
-		)
-	}
 	if r.IsHAEnabled() && r.Spec.Replicas <= 1 {
 		return field.Invalid(
 			field.NewPath("spec").Child("replicas"),
--- a/api/v1alpha1/mariadb_webhook.go
+++ b/api/v1alpha1/mariadb_webhook.go
@@ -130,13 +130,6 @@ func (r *MariaDB) validateGalera() error {
 			)
 		}
 	}
-	if galera.ReplicaThreads < 0 {
-		return field.Invalid(
-			field.NewPath("spec").Child("galera").Child("replicaThreads"),
-			galera.ReplicaThreads,
-			"'spec.galera.replicaThreads' must be at least 1",
-		)
-	}
 	_, exists := galera.ProviderOptions[galerakeys.WsrepOptISTRecvAddr]
 	if exists {
 		return field.Invalid(
--- a/api/v1alpha1/mariadb_webhook.go
+++ b/api/v1alpha1/mariadb_webhook.go
@@ -1,7 +1,6 @@
 package v1alpha1
 
 import (
-	"errors"
 	"reflect"
 
 	galerakeys "github.com/mariadb-operator/mariadb-operator/pkg/galera/config/keys"
@@ -94,9 +93,6 @@ func (r *MariaDB) ValidateDelete() (admission.Warnings, error) {
 }
 
 func (r *MariaDB) validateHA() error {
-	if r.Replication().Enabled && r.IsGaleraEnabled() {
-		return errors.New("You may only enable one HA method at a time, either 'spec.replication' or 'spec.galera'")
-	}
 	if !r.IsHAEnabled() && r.Spec.Replicas > 1 {
 		return field.Invalid(
 			field.NewPath("spec").Child("replicas"),
--- a/api/v1alpha1/mariadb_webhook_test.go
+++ b/api/v1alpha1/mariadb_webhook_test.go
@@ -478,20 +478,6 @@ var _ = Describe("MariaDB webhook", func() {
 				},
 				false,
 			),
-			Entry(
-				"Valid rootEmptyPassword",
-				&MariaDB{
-					ObjectMeta: meta,
-					Spec: MariaDBSpec{
-						RootPasswordSecretKeyRef: GeneratedSecretKeyRef{},
-						RootEmptyPassword:        ptr.To(true),
-						Storage: Storage{
-							Size: ptr.To(resource.MustParse("100Mi")),
-						},
-					},
-				},
-				false,
-			),
 		)
 
 		It("Should default replication", func() {
--- a/internal/controller/mariadb_controller_galera_test.go
+++ b/internal/controller/mariadb_controller_galera_test.go
@@ -93,6 +93,7 @@ var _ = Describe("MariaDB Galera", Ordered, func() {
 								},
 							},
 						},
+						ReplicaThreads: 1,
 					},
 				},
 				Replicas: 3,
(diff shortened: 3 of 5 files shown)


api/v1alpha1/mariadb_webhook.go
```diff
@@ -94,23 +93,6 @@ func (r *MariaDB) ValidateDelete() (admission.Warnings, error) {
 }
 
 func (r *MariaDB) validateHA() error {
-	if r.Replication().Enabled && r.IsGaleraEnabled() {
```
mmontes11, line 97: Could we leave the current validations as they are and make them configurable?

Although [CEL was introduced as alpha in 1.26](https://kubernetes.io/blog/2022/12/20/validating-admission-policies-alpha/), I would like to do a progressive migration instead of getting rid of our current validation straight away.

We would announce CEL support in our release notes for those who are in Kubernetes 1.26 or greater, and at the same time urge people in older versions to migrate, saying that the current webhook validations will be deprecated in further releases.

In any case, having duplicate validations is not ideal, I would make them configurable via the `WEBHOOK_LEGACY_VALIDATIONS_DISABLED` environment variable to be injected in the webhook deployment. Basically, whenever this env variable is present we will be skipping elements in the following slice:
- https://github.com/mariadb-operator/mariadb-operator/blob/b5836f17930f3010bd9313d25a949cdd574f6559/api/v1alpha1/mariadb_webhook.go#L45

What do you think?

The rationale behind using and env variable to control this instead of checking the Kubernetes versions is that the webhook is quite limited in regards to extensibility:
- https://github.com/mariadb-operator/mariadb-operator/blob/main/api/v1alpha1/mariadb_webhook.go

Since the `ValidateCreate` and `ValidateUpdate` methods need to be defined in the CR, there is no way we can have a single discovery client, instead we would need to instanciate a discovery client per each request, leading to increased memory allocations. We could use a singleton available through a shared variable, but I am not a big fan of this approach, it is more complex and I think it doesn't really worth it for this case.

mmontes11 (changes requested):
Hey there @businessbean ! Thanks a lot for your work here paving the road towards the future of validations.

Although I do think this is the way forward, I think we should roll this out progressively, giving time to our users to migrate to Kubernetes 1.26 in case they haven't already. Please see my comments for details

I also should express that I am a bit concerned about the readability of CEL, but this might just be a personal opinion, as I feel more comfortable reading and maintaining Go code rather than conditions that tend to be complicated. In any case, I wanted to double check with you as I haven't really used CEL before, since the validations are done by the API server, our webhook tests still apply right? We need to have a safety net in case that regressions are introduced in the CEL conditions.

I guess this might be a readability vs maintenance trade-off, as in the future we might be able to get rid of the webhook (?) Or maybe not, since we will need  to have conversion webhooks anyway to automatically migrate from one version to another `v1alpha1` -> `v1beta1` or to potentially perform complex validations.

Again thanks a lot for your hard work on this, please see my comments.

internal/controller/mariadb_controller_galera_test.go
```diff
@@ -93,6 +93,7 @@ var _ = Describe("MariaDB Galera", Ordered, func() {
 								},
 							},
 						},
+						ReplicaThreads: 1,
```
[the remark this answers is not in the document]
businessbean: At validation time `ReplicaThreads` seems to be not set. But you set it in different areas of the code. It's not clear to me why this is happening.

api/v1alpha1/mariadb_webhook_test.go
```diff
@@ -478,20 +478,6 @@ var _ = Describe("MariaDB webhook", func() {
 				},
 				false,
 			),
-			Entry(
```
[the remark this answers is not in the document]
businessbean: The unit test triggered by the PR action had been failed and i thought it is not required anymore because the CRD validation will do that...

api/v1alpha1/mariadb_webhook.go
businessbean: Because of the operator is on `v1alpha1` i thought i can be more "aggressive", but providing both checks with a deprecation phase is also a good idea. It means we need a Kubernetes version check [for the CEL rules](https://github.com/mariadb-operator/mariadb-operator/blob/8e5bbbc9eea4f5be96954796ed39a58dd052726a/api/v1alpha1/mariadb_types.go#L531-L536) to be able to still support older Kubernetes releases. On the other hand `1.26` is already [EOL](https://kubernetes.io/releases/)...

businessbean: Kubernetes `1.22` fails to import the CRD definitions if the validation code is included.
```
error: error validating "crds.yaml": error validating data: ValidationError(CustomResourceDefinition.spec.versions[0].schema.openAPIV3Schema.properties.spec): unknown field "x-kubernetes-validations" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.JSONSchemaProps; if you choose to ignore these errors, turn validation off with --validate=false
```
This means we would need to generate two `crds.yaml` files (with and without the validation code) and we cannot use Helm to apply the correct one for a certain cluster, because Helm does not support the [templating of CRD definitions](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#some-caveats-and-explanations)...

mmontes11: I'm planing to release `v0.0.30` shortly and the idea is to provide some visibility about our initial CEL support introduced in https://github.com/mariadb-operator/mariadb-operator/pull/783, also referring to the current PR to highlight that we are still working on it. I would prefer to roll this out progressively to see if people is actually facing other problems we haven't considered yet.

At the same time, I will be using this opportunity to say that our minimum supported version is .126, encouraging people to migrate. 

> Kubernetes 1.22 fails to import the CRD definitions if the validation code is included.

Thank you for testing this! I was testing the `release-v0.0.30` branch with Kubernetes 1.22, which has changes introduced in https://github.com/mariadb-operator/mariadb-operator/pull/783, and therefore also the `x-kubernetes-validations` field but I couldn't reproduce the problem. Maybe it has to do with something in the current branch?

```bash
❯ export KIND_IMAGE=kindest/node:v1.22.0

❯ make cluster-delete && make cluster
/home/martin/go/src/github.com/mariadb-operator/mariadb-operator/bin/kind delete cluster --name mdb
Deleting cluster "mdb" ...
Deleted nodes: ["mdb-control-plane"]
/home/martin/go/src/github.com/mariadb-operator/mariadb-operator/bin/kind create cluster --name mdb --image kindest/node:v1.22.0
Creating cluster "mdb" ...
 ✓ Ensuring node image (kindest/node:v1.22.0) 🖼
 ✓ Preparing nodes 📦
 ✓ Writing configuration 📜
 ✓ Starting control-plane 🕹️
 ✓ Installing CNI 🔌
 ✓ Installing StorageClass 💾
Set kubectl context to "kind-mdb"
You can now use your cluster with:

kubectl cluster-info --context kind-mdb

Not sure what to do next? 😅  Check out https://kind.sigs.k8s.io/docs/user/quick-start/

❯ make docker-build && make docker-load
docker buildx build -t docker-registry3.mariadb.com/mariadb-operator/mariadb-operator:v0.0.30 . --load
/home/martin/go/src/github.com/mariadb-operator/mariadb-operator/bin/kind load docker-image --name mdb docker-registry3.mariadb.com/mariadb-operator/mariadb-operator:v0.0.30
Image: "docker-registry3.mariadb.com/mariadb-operator/mariadb-operator:v0.0.30" with ID "sha256:3418ab5d5122df88db981808bba8bac6425571ddd773e09df75627bf0f994937" not yet present on node "mdb-control-plane", loading...

❯ helm install mariadb-operator deploy/charts/mariadb-operator
NAME: mariadb-operator
LAST DEPLOYED: Thu Aug 22 11:21:41 2024
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
mariadb-operator has been successfully deployed! 🦭

Not sure what to do next? 😅 Check out:
https://github.com/mariadb-operator/mariadb-operator#quickstart

❯ kubectl get pods
NAME                                                READY   STATUS    RESTARTS   AGE
mariadb-operator-558785774-clll9                    1/1     Running   0          49s
mariadb-operator-cert-controller-7c4b7c5dd5-48z9n   1/1     Running   0          49s
mariadb-operator-webhook-98765c9c-jmrcz             1/1     Running   0          49s
``` 

> On the other hand 1.26 is already [EOL](https://kubernetes.io/releases/)

Absolutely, it's EOL I get your point. However, in my experience, Kubernetes version pace is faster than people migrating. My suggestion is to include a notice in the release notes and offer users a little more time to make the transition.

mmontes11: REAMDE.md updated:
- https://github.com/mariadb-operator/mariadb-operator/tree/release-v0.0.30?tab=readme-ov-file#kubernetes

Kubernetes v1.26 will be the minimum supported version starting from mariadb-operator v0.0.30. This is a step towards getting everyone migrated to a version with CEL, so we can continue working on this PR and roll it out in further versions.

businessbean: > v0.0.30 REAMDE.md updated:
> 
>     * https://github.com/mariadb-operator/mariadb-operator/tree/release-v0.0.30?tab=readme-ov-file#kubernetes
> 
> 
> Kubernetes v1.26 will be the minimum supported version starting from mariadb-operator v0.0.30. This is a step towards getting everyone migrated to a version with CEL, so we can continue working on this PR and roll it out in further versions.

That sounds good.

businessbean:
> Thank you for testing this! I was testing the release-v0.0.30 branch with Kubernetes 1.22, which has changes introduced in https://github.com/mariadb-operator/mariadb-operator/pull/783, and therefore also the x-kubernetes-validations field but I couldn't reproduce the problem. Maybe it has to do with something in the current branch?

I have tested it again. It works if i install the CRD's with `helm install mariadb-operator deploy/charts/mariadb-operator`, but it fails with `kubectl apply --server-side=true --filename=deploy/charts/mariadb-operator/crds/crds.yaml`. Helm seems to disable the validation [during installation](https://github.com/helm/helm/blob/668acec7f6ea3d459a9c0dd3e7958fa4d9791c75/pkg/action/install.go#L160) and a `kubectl get crd/mariadbs.k8s.mariadb.com -o yaml| grep validation` does not find any `x-kubernetes-validations` entries.

Btw. Thanks for describing all the required steps for a local build and installation.

mmontes11:
> I have tested it again. It works if i install the CRD's with helm install mariadb-operator deploy/charts/mariadb-operator, but it fails with kubectl apply --server-side=true --filename=deploy/charts/mariadb-operator/crds/crds.yaml. Helm seems to disable the validation [during installation](https://github.com/helm/helm/blob/668acec7f6ea3d459a9c0dd3e7958fa4d9791c75/pkg/action/install.go#L160) and a kubectl get crd/mariadbs.k8s.mariadb.com -o yaml| grep validation does not find any x-kubernetes-validations entries.

Hey @businessbean! Thanks a lot for giving it a second try, this is very good to know! 

This basically means that when installing the new `v0.0.30` release in Kubernetes < v1.26, it will complain if installing the [raw manifests ](https://github.com/mariadb-operator/mariadb-operator/tree/main/deploy/manifests) on the server side. This is more than acceptable I would say, besides, v1.26 will be the minimal supported Kubernetes version, so we are good!

> Btw. Thanks for describing all the required steps for a local build and installation.

No prob. There are more details [here](https://github.com/mariadb-operator/mariadb-operator/blob/main/docs/DEVELOPMENT.md).

PR #826, opened 2024-09-04
Update mariadb-operator-enterprise.clusterserviceversion.yaml

mariadb-pieterhumphrey:
updating description with a few links.


mmontes11 (approved):
LGTM! I will run `make bundle` to regenerate the final bundle in:
- https://github.com/mariadb-operator/mariadb-operator/blob/main/deploy/olm/manifests/mariadb-operator-enterprise.clusterserviceversion.yaml

Issue #642, opened 2024-05-24
[Bug] Scale to 0 replicas


MMEe111:
I need to stop the galera cluster, but I can't stop it with scale replicas=0, and I can't run the galera cluster properly after I delete the yaml of the galera cluster

Kubernetes v1.25.0
mariadb-operator v0.0.28

log
```makefile
Defaulted container "mariadb" out of: mariadb, agent, init (init)
2024-05-24 08:02:55+00:00 [Note] [Entrypoint]: Entrypoint script for MariaDB Server 1:10.11.7+maria~ubu2204 started.
2024-05-24 08:02:56+00:00 [Warn] [Entrypoint]: /sys/fs/cgroup///memory.pressure not writable, functionality unavailable to MariaDB
2024-05-24 08:02:56+00:00 [Note] [Entrypoint]: Switching to dedicated user 'mysql'
2024-05-24 08:02:56+00:00 [Note] [Entrypoint]: Entrypoint script for MariaDB Server 1:10.11.7+maria~ubu2204 started.
2024-05-24 08:02:58+00:00 [Note] [Entrypoint]: MariaDB upgrade not required
2024-05-24  8:02:58 0 [Note] Starting MariaDB 10.11.7-MariaDB-1:10.11.7+maria~ubu2204 source revision 87e13722a95af5d9378d990caf48cb6874439347 as process 1
2024-05-24  8:02:58 0 [Note] WSREP: Loading provider /usr/lib/galera/libgalera_smm.so initial position: 00000000-0000-0000-0000-000000000000:-1
2024-05-24  8:02:58 0 [Note] WSREP: wsrep_load(): loading provider library '/usr/lib/galera/libgalera_smm.so'
2024-05-24  8:02:58 0 [Note] WSREP: wsrep_load(): Galera 26.4.16(r7dce5149) by Codership Oy <info@codership.com> loaded successfully.
2024-05-24  8:02:58 0 [Note] WSREP: Initializing allowlist service v1
2024-05-24  8:02:58 0 [Note] WSREP: CRC-32C: using 64-bit x86 acceleration.
2024-05-24  8:02:58 0 [Note] WSREP: Found saved state: 40bc1019-1913-11ef-a4af-b3463c57d946:-1, safe_to_bootstrap: 0
2024-05-24  8:02:58 0 [Note] WSREP: GCache DEBUG: opened preamble:
Version: 2
UUID: 40bc1019-1913-11ef-a4af-b3463c57d946
[...]

Issue #638, opened 2024-05-22
[Bug] Unable to scale to 0


kimutir:
**Documentation**
- [x] I acknowledge that I have read the relevant [documentation](https://github.com/mariadb-operator/mariadb-operator/tree/main/docs).

**Describe the bug**

Hello!
I created MariaDB with 3 replicas and replication enabled from example manifest.
When i scale  by setting replicas in manifest to 1, i have to disable replication. It is ok, but after scaling down 2 pvcs remain untouched.
During scaling  by setting replicas in manifest to 0 with disabled replication kubectl output says that MariaDB has been configured, but there is still one pod left, replication status shows last replicas and all pvcs still exist.

```
kubectl apply -f mariadb-replication.yaml
> mariadb.k8s.mariadb.com/mariadb-repl configured

kubectl get pod
NAME                                                              READY   STATUS    RESTARTS         AGE
mariadb-repl-0                                                    1/1     Running   0                107s

kubectl get mariadb mariadb-repl -o yaml
apiVersion: k8s.mariadb.com/v1alpha1
kind: MariaDB
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"k8s.mariadb.com/v1alpha1","kind":"MariaDB","metadata":{"annotations":{},"name":"mariadb-repl","namespace":"default"},"spec":{"metrics":{"enabled":true},"primaryService":{"metadata":{"annotations":{"metallb.universe.tf/loadBalancerIPs":"172.18.0.130"}},"type":"LoadBalancer"},"replicas":0,"replication":{"enabled":false},"rootPasswordSecretKeyRef":{"generate":true,"key":"root-password","name":"mariadb"},"secondaryService":{"metadata":{"annotations":{"metallb.universe.tf/loadBalancerIPs":"172.18.0.131"}},"type":"LoadBalancer"},"service":{"metadata":{"annotations":{"metallb.universe.tf/loadBalancerIPs":"172.18.0.120"}},"type":"LoadBalancer"},"storage":{"size":"1Gi"}}}
  creationTimestamp: "2024-05-22T09:39:19Z"
  generation: 5
  name: mariadb-repl
  namespace: default
  resourceVersion: "27884856"
  uid: 6b7b4aa5-6eaf-45f8-bbd4-bb72704c0f3b
spec:
  image: mariadb:10.11.7
  metrics:
    enabled: true
    exporter:
[...]

PR #829, opened 2024-09-04
Allow scaling to 0 to stop workloads

Jancis:
Close https://github.com/mariadb-operator/mariadb-operator/issues/642
Close https://github.com/mariadb-operator/mariadb-operator/issues/638

Allows scaling to 0 to stop workloads. Works both ways, by scaling mariadb resource (`kubectl scale mdb`) and statefulset (`kubectl scale sts`) - changes are reflected to the related resource.

### DIFF
--- a/api/v1alpha1/condition_types.go
+++ b/api/v1alpha1/condition_types.go
@@ -34,6 +34,7 @@ const (
 	ConditionReasonUpdating            string = "Updating"
 	ConditionReasonUpdated             string = "Updated"
 	ConditionReasonSuspended           string = "Suspended"
+	ConditionReasonStopped             string = "Stopped"
 
 	ConditionReasonMaxScaleNotReady string = "MaxScaleNotReady"
 	ConditionReasonMaxScaleReady    string = "MaxScaleReady"
--- a/api/v1alpha1/mariadb_types.go
+++ b/api/v1alpha1/mariadb_types.go
@@ -429,9 +429,9 @@ type MariaDBSpec struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	MaxScale *MariaDBMaxScaleSpec `json:"maxScale,omitempty"`
 	// Replicas indicates the number of desired instances.
-	// +kubebuilder:default=1
+	// +optional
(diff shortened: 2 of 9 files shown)


mmontes11 (changes requested):
Hey there @Jancis ! Thanks a lot, this is a fantastic contribution, Seems like you are understanding the codebase very well despite this one being your first contribution, well done 🎖️ 

I've left some detailed comments that hopefully will serve you as guidance to cross the finish line, it looks promising. 

I also have some more asks for you if you don't mind:
- The same way we did with the `Suspended` state, please add an integration tests [here](https://github.com/mariadb-operator/mariadb-operator/blob/343405d3099fbf846862f4933ceccd12ee220cd8/internal/controller/mariadb_controller_test.go#L90) to prove that this feature works as intended.
- Whenever we upscale a Galera cluster, we could set the `GaleraNotReady` condition like we do [here](https://github.com/mariadb-operator/mariadb-operator/blob/343405d3099fbf846862f4933ceccd12ee220cd8/pkg/controller/galera/init.go#L51) to trigger the recovery process. I can do this myself if you feel like it is too complicated.

Thanks!

Jancis:
I implemented most of the changes (except tests, did not check them out yet), but I ran into issue that it tries runs reconciliation cycle on Downscaled (non-Ready) mariadb resources, causing frequent API calls.

pinpins:
no plans to get it merged?

Jancis:
I lacked ability to make this work properly, it was making quite a bit of API queries while downscaled, to query state and I could not figure out the best approach; then I got pulled into other stuff and I dropped the mariadb-operator for a while.

I made the requested changes (Use "Downscaled" instead of "Stopped"; Don't reconcile statefulset) but never pushed it. Here they are, in case you are interested - https://github.com/mariadb-operator/mariadb-operator/compare/main...wunderio:mariadb-operator:feature/downscaling?expand=1

I made a custom operator that does the reconciliation of mariadb operator owned statefulsets back to mdb resource (that I had to [remove from the PR](https://github.com/mariadb-operator/mariadb-operator/commit/e4877980909333145f559c0986f00f1e101b877e#diff-7ca6f0c1644a79617a2178f139538d91af71f183a30e80a947dc4f5a04e1d185L57-L67)), but then i left that idea too, had to switch over to other stuff.

Issue #341, opened 2024-01-15
[Feature] Compress backup files


TheNozomi:
**Is your feature request related to a problem? Please describe.**
When using mariadb-dump/mysqldump, you can pipe its output to a compression program, for instance, gzip or 7z, and compress the resulting dump file on a single step.
As far as I know, this is currently not possible with the mariadb-operator `Backup` resource, and if you want to compress the dump you have to do so manually.

**Describe the solution you'd like**
Add a `spec.compress` field to the `Backup` resource, which can be a boolean, or a string that indicates the compression type (gzip, 7z, zstd, etc.). If compression is enabled, the backup job would compress the file using the selected compression program, and then save it.

**Describe alternatives you've considered**
N/A

**Environment details**:
- Kubernetes version: 1.27.5
- Kubernetes distribution: k3s
- mariadb-operator version: 0.24.0
- Install method: Helm
- Install flavor: Recommended

**Additional context**
As for now, what we're doing to save compressed dumps is using the Backup resource to generate a dump, save it to an NFS volume mapped in Kubernetes, and then we have a cron job on the server that reads the file, compresses it using gzip, and uploads it to S3 using the AWS CLI.
Now that there's support for directly uploading dumps to S3 in mariadb-operator (which I've been testing and works well for our use case), it would be very useful to have the option of compressing the dumps before uploading them to S3, so we can eliminate the manual steps.

mmontes11:
Hola @TheNozomi ! 100% this makes total sense to me.

Not only we should support it in the `Backups`, but also we should make the `Restore` resource compatible with it. As you said we should introduce a new field in the `Backup` resource indicating the compression algorithm:

```yaml
apiVersion: mariadb.mmontes.io/v1alpha1
kind: Backup
metadata:
  name: backup
spec:
   compression: gzip
   ....
```
We don't need to change anything in the `Restore` API, as we already have a reference to the `Backup`:
```yaml
apiVersion: mariadb.mmontes.io/v1alpha1
kind: Restore
metadata:
  name: restore
spec:
  backupRef:
    name: backup
  ...
```

Regarding the implementation, I would support this in our new [backup](https://github.com/mariadb-operator/mariadb-operator/blob/main/cmd/backup/main.go) and [restore](https://github.com/mariadb-operator/mariadb-operator/blob/main/cmd/backup/restore.go) subcommands:
[...]

PR #836, opened 2024-09-08
backup compression

vixns:
Close https://github.com/mariadb-operator/mariadb-operator/issues/341

I'm trying to implement #341 based mostly on https://github.com/mariadb-operator/mariadb-operator/issues/341#issuecomment-1913116742

Feel free to edit or reuse any part of this contribution in another PR


cmd/backup/restore.go
```diff
@@ -81,3 +93,64 @@ func getTargetTime() (time.Time, error) {
...
+		return (err)
+	}
+	defer originalFile.Close()
+
+	bReader := bufio.NewReader(compressedFile)
+	testBytes, err := bReader.Peek(2)
+	if err != nil {
+		return (err)
+	}
+	if testBytes[0] == 31 && testBytes[1] == 139 {
```
mmontes11, line 120: Although I really appreciated the effort here, I think this will be difficult to maintain and scale to have more compression algorithms. 

I suggest that, instead of checking the contents of the file, we could add the compression algorithm to the backup filename: `backup.2024-08-26T12:24:34Z.{gzip,zlib}.sq`l. We are handling this here:
- https://github.com/mariadb-operator/mariadb-operator/blob/6171b061f944e10e29197fbdc7e4d1144b08352d/pkg/backup/backup.go#L69
- https://github.com/mariadb-operator/mariadb-operator/blob/6171b061f944e10e29197fbdc7e4d1144b08352d/pkg/command/backup.go#L226

This way, the name of the file has all the metadata explicitly to operate with it, it is human readable and there is no need to check its contents.

It is important to note that, this should be backwards compatible, the process should allow file names without compression.

mmontes11 (changes requested):
Hey there @vixns ! Thank you very much for working on this.

I've left some notes, but I think you are on the right track!

Also, it would be great if we could add some integration tests, to ensure that the functionality is working as expected:
- https://github.com/mariadb-operator/mariadb-operator/blob/main/internal/controller/backup_controller_test.go
- https://github.com/mariadb-operator/mariadb-operator/blob/main/internal/controller/restore_controller_test.go

vixns:
Thanks @mmontes11 for the detailed review, this PR improved a lot! 
I'll try to add some integration tests soon.

api/v1alpha1/backup_types.go
```diff
@@ -61,7 +82,7 @@ type BackupSpec struct {
 	// Compression algorithm to be used in the Backup.
 	// +optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
```
mmontes11, line 84: I suggest we add an enum annotation here, to it maps to an OpenAPI enum. It would be something like:

```bash
	// +kubebuilder:validation:Enum=none;zlib;gzip
```

mmontes11:
@vixns wow, this is looking great! congratulations, it is exactly what I had in mind! 🚀 

Left another suggestion regarding adding an enum to the OpenAPI spec.

This way, it will be validated by the Kubernetes apiserver and we don't need to add anything extra to the validation webhook.

api/v1alpha1/backup_types.go
mmontes11: See an example here:
https://github.com/mariadb-operator/mariadb-operator/blob/6171b061f944e10e29197fbdc7e4d1144b08352d/api/v1alpha1/mariadb_galera_types.go#L360

mmontes11 (changes requested):
@vixns wow, this is looking great! congratulations, it is exactly what I had in mind! 🚀 

Left another suggestion regarding adding an enum to the OpenAPI spec.

This way, it will be validated by the Kubernetes apiserver and we don't need to add anything extra to the validation webhook.

api/v1alpha1/backup_types.go
vixns: done, thx for the tip

vixns:
Sadly i can't run the integration tests on my local computers, they are sloooow ( 20 minutes for 30 tests before i stop them) and many timeouts makes random tests failing.

I can't develop and test any integration test in these conditions, is there a way to run a solo test ? @mmontes11

mmontes11:
> I can't develop and test any integration test in these conditions, is there a way to run a solo test ? @mmontes11

Hey @vixns ! Sorry to see you struggling. Yes, you can add a ginkgo `Focus` decorator:
- https://onsi.github.io/ginkgo/#focused-specs

I hope this helps, let me know otherwise so I can figure out something else.

mmontes11:
Hey @vixns ! I will have some spare time soon, planning to port this code to another PR and add you as a contributor in the commit to give you credit. I hope this is fine with you.

I have plans to add:
- The missing integration tests
- Replace `zlib` with `bzip2` leveraging https://github.com/dsnet/compress/tree/master/bzip2. `zlib` is not really used for backups and `bzip2` offers a better compression ratio than `gzip`

vixns:
@mmontes11 just added bzip2, still no tests

mmontes11:
Thanks @vixns ! `bzip2` changes LGTM as well! Marked as [done](https://github.com/mariadb-operator/mariadb-operator/pull/836#issuecomment-2346913183)

vixns:
> LGTM! 🚀 Thanks for the great work @vixns !
> 
> We can merge once we have integration tests!

Just added some tests, I could not run them on my machine :crossed_fingers:

mmontes11 (approved):
LGTM! 💯 🚀

It works like a charm! Thank you very much for this massive contribution @vixns ! 🙇🏻 🙏🏻 

Will document this feature and release it in the next version! Stay tuned!

PR #909, opened 2024-10-04
Add replication role labels

### LINKED ISSUE #823: [Feature] Add primary / secondary labels to pods
Please, consider labeling pods in the SemiSync Replication cluster as `primary` /  `secondary` to allow k8s admins to use label selectors when listing pods, or when there is a need to create a custom service with a custom port (for a sidecar) that should target only either master or replica.

Alternatively, allow specifying custom ports in https://github.com/mariadb-operator/mariadb-operator/blob/main/docs/API_REFERENCE.md#servicetemplate

**Environment details**:
- Kubernetes version: [1.25]
- Kubernetes distribution: [Rancher]
- mariadb-operator version: [0.29.0]
- Install method: [helm]
- Install flavor: [recommended]

nocturo:
This PR closes #823.

The assumption made here is that we only want this to be available when replication is enabled, since on galera it's all primary and standalone is not a cluster. We could perhaps tag it as such, i.e `k8s.mariadb.com/role=galera` or `k8s.mariadb.com/role=standalone`, but I think there is little to no value having it.

### DIFF
--- a/internal/controller/mariadb_controller.go
+++ b/internal/controller/mariadb_controller.go
@@ -349,6 +349,10 @@ func (r *MariaDBReconciler) reconcileStatefulSet(ctx context.Context, mariadb *m
 }
 
 func (r *MariaDBReconciler) reconcilePodLabels(ctx context.Context, mariadb *mariadbv1alpha1.MariaDB) (ctrl.Result, error) {
+	if mariadb.Status.CurrentPrimaryPodIndex == nil {
+		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
+	}
+
 	podList := corev1.PodList{}
 	listOpts := &client.ListOptions{
 		LabelSelector: klabels.SelectorFromSet(
@@ -373,7 +377,7 @@ func (r *MariaDBReconciler) reconcilePodLabels(ctx context.Context, mariadb *mar
 			return ctrl.Result{}, fmt.Errorf("error getting Pod '%s' index: %v", pod.Name, err)
 		}
 
-		if mariadb.Status.CurrentPrimaryPodIndex != nil && *podIndex == *mariadb.Status.CurrentPrimaryPodIndex {
+		if *podIndex == *mariadb.Status.CurrentPrimaryPodIndex {
 			role = "primary"
 		}
 
--- a/internal/controller/mariadb_controller.go
+++ b/internal/controller/mariadb_controller.go
@@ -32,14 +32,17 @@ import (
 	"github.com/mariadb-operator/mariadb-operator/pkg/environment"
 	"github.com/mariadb-operator/mariadb-operator/pkg/health"
 	"github.com/mariadb-operator/mariadb-operator/pkg/metadata"
+	mdbpod "github.com/mariadb-operator/mariadb-operator/pkg/pod"
 	"github.com/mariadb-operator/mariadb-operator/pkg/predicate"
 	"github.com/mariadb-operator/mariadb-operator/pkg/refresolver"
+	sts "github.com/mariadb-operator/mariadb-operator/pkg/statefulset"
 	"github.com/mariadb-operator/mariadb-operator/pkg/watch"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	policyv1 "k8s.io/api/policy/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	klabels "k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
@@ -164,6 +167,10 @@ func (r *MariaDBReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		{
 			Name:      "Replication",
 			Reconcile: r.ReplicationReconciler.Reconcile,
+		},
+		{
+			Name:      "Labels",
+			Reconcile: r.reconcilePodLabels,
 		},
 		{
 			Name:      "Galera",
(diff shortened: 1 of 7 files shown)


internal/controller/mariadb_controller.go
```diff
@@ -341,6 +348,55 @@ func (r *MariaDBReconciler) reconcileStatefulSet(ctx context.Context, mariadb *m
...
+
+		if pod.Status.PodIP == "" || pod.Spec.NodeName == "" {
+			continue
+		}
+		podIndex, err := sts.PodIndex(pod.Name)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("error getting Pod '%s' index: %v", pod.Name, err)
+		}
+
+		if *podIndex == *mariadb.Status.CurrentPrimaryPodIndex {
```
mmontes11, line 380: `mariadb.Status.CurrentPrimaryPodIndex` may not be set at this point under some conditions, potentially leading to dereferencing a nil pointer. Could we add a guard at the begining of this function to ensure we have `mariadb.Status.CurrentPrimaryPodIndex`  set ?

internal/controller/mariadb_controller.go
```diff
@@ -341,6 +348,55 @@ func (r *MariaDBReconciler) reconcileStatefulSet(ctx context.Context, mariadb *m
...
+			role = "primary"
+		}
+		pod.Labels = labels.NewLabelsBuilder().WithLabels(pod.Labels).WithPodRole(role).Build()
+
+		if mdbpod.PodReady(&pod) {
+			if err := r.Update(ctx, &pod); err != nil {
```
mmontes11, line 386: The operator does not have permission for updating `Pods`. Would you mind updating this to automatically generate them when running `make gen`?

https://github.com/mariadb-operator/mariadb-operator/blob/1d6dac00e10ae0fbfbf0fa8cbced5bf0c97266aa/internal/controller/mariadb_controller.go#L98C1-L99C1

Also, since the `Pod` spec is quite big and to follow the codebase conventions, I think it will be better to use a `patch` operation instead, see an example here:
https://github.com/mariadb-operator/mariadb-operator/blob/1d6dac00e10ae0fbfbf0fa8cbced5bf0c97266aa/pkg/controller/galera/recovery.go#L475

We will need to update the helm chart RBAC as well, as it is not automatically generated by `make gen`:
https://github.com/mariadb-operator/mariadb-operator/blob/1d6dac00e10ae0fbfbf0fa8cbced5bf0c97266aa/deploy/charts/mariadb-operator/templates/rbac.yaml#L92
https://github.com/mariadb-operator/mariadb-operator/blob/1d6dac00e10ae0fbfbf0fa8cbced5bf0c97266aa/deploy/charts/mariadb-operator/templates/rbac-namespace.yaml#L75

mmontes11 (changes requested):
Thanks for your contribution! Seems like you easily get acquainted with the codebase! 💯 🙏🏻 

You are on the right track, just added some suggestions! 

> We could perhaps tag it as such, i.e k8s.mariadb.com/role=galera or k8s.mariadb.com/role=standalone, but I think there is little to no value having it.

Although Galera topology allows multi-master, we are performing the writes in a single node to prevent deadlock. So basically primary and replica concepts still apply to our case, in fact, the `*-primary` and `*-secondary` `Services` may be used with Galera as well. Therefore, I think we could reuse the labels for Galera as well, see the comment.

internal/controller/mariadb_controller_replication_test.go
```diff
@@ -155,6 +155,20 @@ var _ = Describe("MariaDB replication", Ordered, func() {
 		By("Expecting to create a secondary Service")
 		Expect(k8sClient.Get(testCtx, mdb.SecondaryServiceKey(), &svc)).To(Succeed())
 
+		By("Expecting role label to be set to primary")
```
mmontes11, line 158: It will be great to add this checks to the Galera test as well, since this labels are not tied to replication.

nocturo: Added test for galera too.

internal/controller/mariadb_controller.go
nocturo: Thanks good catch, I've changed it to use `patch` instead of `update` and updated Helm charts to include patch verb.

nocturo:
Thanks for the lightning fast review! ⚡ 

I've addressed the comments, let me know if there is anything else you'd need updated.

internal/controller/mariadb_controller.go
```diff
@@ -341,6 +348,53 @@ func (r *MariaDBReconciler) reconcileStatefulSet(ctx context.Context, mariadb *m
...
+
+		if pod.Status.PodIP == "" || pod.Spec.NodeName == "" {
+			continue
+		}
+		podIndex, err := sts.PodIndex(pod.Name)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("error getting Pod '%s' index: %v", pod.Name, err)
+		}
+
+		if mariadb.Status.CurrentPrimaryPodIndex != nil && *podIndex == *mariadb.Status.CurrentPrimaryPodIndex {
```
mmontes11, line 376: Instead of doing this check here could we do this at the begining of `reconcilePodLabels` the function:

```go
if mariadb.Status.CurrentPrimaryPodIndex == nil {
    return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
}
``` 

`CurrentPrimaryPodIndex` will be set eventually, but without this early return we will be marking all pods as `replica`, which is not very precise. Then we can update the current statement to just:

```suggestion
		if  *podIndex == *mariadb.Status.CurrentPrimaryPodIndex {
```

Doing as suggested should do the trick.

mmontes11 (changes requested):
Thanks for the changes @nocturo ! 

All looks good except one minor detail, see my comment.

mmontes11:
@nocturo  Also, run `make gen` to generate the artifacts:
- https://github.com/mariadb-operator/mariadb-operator/actions/runs/11182807574/job/31090532013?pr=909

internal/controller/mariadb_controller.go
nocturo: Thanks, I haven't seen that happen during testing, primary was always coming first before I saw replica labels being applied but doesn't hurt to be more precise :)

mmontes11 (approved):
LGTM! 🚀 

Thank you very much for you contribution @nocturo ! 🙇🏻

PR #912, opened 2024-10-05
Allow adding custom ports in services

hedgieinsocks:
This PR follows up with https://github.com/mariadb-operator/mariadb-operator/issues/823#issuecomment-2340194499 suggestion of adding `ServicePorts` to MariaDB.

This will allow us to do something like below, so we could easily create a `ServiceMonitor` that will target specifically replica and leave more air for the busy primary. But that's just an example to prop the validity of the proposed change.

```yaml
spec:
  sidecarContainers:
    - name: disk-usage-exporter
      image: cool_exporter_image:1.0.1
     
  servicePorts:
    - name: disk-usage-exporter
      port: 9995

  secondaryService:
    metadata:
      labels:
        app.kubernetes.io/component: disk-usage-exporter
```

I initially tried to do validation via `+kubebuilder:validation` rules, but that did not feel right (or I just don't have experience with such things to do the job properly). 

Anyway, hope to hear your feedback next week. Have a good weekend for now :relaxed:


mmontes11 (changes requested):
Hey there @hedgieinsocks ! Thanks for another contribution. 🙏🏻 

Your suggestion makes sense to me and the approach followed looks correct except for one reason.

In the last release, we decided to not use the Kubernetes type directly in our CRD to minimize its size, as there is a hard limit of 1MB. See the PR: https://github.com/mariadb-operator/mariadb-operator/pull/869

`ServicePort` upstream type has fields we are not directly using and comments that would enlarge the CRD considerably. For this reason, I suggest we create a minimal version of the `ServicePort` type to use it in our CRDs:

```go
// Refer to the Kubernetes docs: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.31/#serviceport-v1-core
type ServicePort struct {
	// +optional
	Name string `json:"name,omitempty" protobuf:"bytes,1,opt,name=name"`
	Port int32 `json:"port" protobuf:"varint,3,opt,name=port"`
}

func (r ServicePort) ToKubernetesType() corev1.ServicePort {
	return corev1.ServicePort{
		Name: r.Name,
		Port: r.Port,
	}
}
```

See other types for further reference:
- https://github.com/mariadb-operator/mariadb-operator/blob/main/api/v1alpha1/kubernetes_types.go

We also have an utility that may be of your interest:
https://github.com/mariadb-operator/mariadb-operator/blob/2de2838bc08af01e2b41451fe1d236b2e9be800b/pkg/kubernetes/adapter/adapter.go#L7

hedgieinsocks:
Once again, thank you for such prompt and detailed guidance! Pushed changes for round 2 review :parachute:

mmontes11 (approved):
LGTM! 👍🏻 

Thanks again for another contribution. This will definetly help enhancing the `MariaDB` extendibility, much appreaciated! 🙏🏻

PR #916, opened 2024-10-06
Use printf to add hosts to /etc/hosts

hedgieinsocks:
`echo` behaves differently across various shells and systems.
The current approach is reported to work in the devcontainer and Ubuntu where `sh` is a symlink to `dash`.
 Unfortunately, this is different in Fedora where special symbols like `\n` and `\t` do not get expanded in `echo` command without `-e` option.
This leads to invalid entries in `/etc/hosts` after running `make net`.
I suggest that we use `printf` in this specific case to make the hack portable and predictable.


mmontes11 (approved):
LGTM! 👍🏻 

Good catch! Thanks for the contribution.

PR #918, opened 2024-10-06
Make `maxUserConnections` mutable

hedgieinsocks:
This PR closes https://github.com/mariadb-operator/mariadb-operator/issues/685

Similar PR made recently by another contributor https://github.com/mariadb-operator/mariadb-operator/pull/723

According to MySQL [documentation](https://dev.mysql.com/doc/refman/8.4/en/user-resources.html), when `MAX_USER_CONNECTIONS` is set to `0`, it falls back to the global `max_user_connections` value. I believe this applies to MariaDB as well.

So I suggest that we simplify the logic by always setting the limit implicitly. 

P.S.: I've left the current default limit of 10 connections. But personally, I think it should be 0 by default. This way, out of the box everyone will be limited to the server setting, and only specific (ab)users could be tweaked if needed.

### DIFF
--- a/api/v1alpha1/user_types.go
+++ b/api/v1alpha1/user_types.go
@@ -32,11 +32,11 @@ type UserSpec struct {
 	// +optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:advanced"}
 	PasswordPlugin PasswordPlugin `json:"passwordPlugin,omitempty"`
-	// MaxUserConnections defines the maximum number of connections that the User can establish.
+	// MaxUserConnections defines the maximum number of simultaneous connections that the User can establish.
 	// +optional
 	// +kubebuilder:default=10
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:number"}
-	MaxUserConnections int32 `json:"maxUserConnections,omitempty" webhook:"inmutable"`
+	MaxUserConnections int32 `json:"maxUserConnections,omitempty"`
 	// Name overrides the default name provided by metadata.name.
 	// +optional
 	// +kubebuilder:validation:MaxLength=80
--- a/api/v1alpha1/user_webhook_test.go
+++ b/api/v1alpha1/user_webhook_test.go
@@ -162,7 +162,7 @@ var _ = Describe("User webhook", func() {
 				func(umdb *User) {
 					umdb.Spec.MaxUserConnections = 20
 				},
-				true,
+				false,
 			),
 			Entry(
 				"Duplicate authentication methods",
--- a/config/crd/bases/k8s.mariadb.com_users.yaml
+++ b/config/crd/bases/k8s.mariadb.com_users.yaml
@@ -84,8 +84,8 @@ spec:
                 type: object
               maxUserConnections:
                 default: 10
-                description: MaxUserConnections defines the maximum number of connections
-                  that the User can establish.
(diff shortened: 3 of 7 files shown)


mmontes11 (commented):
Thanks for one more contribution @hedgieinsocks ! You are on fire! 🦔 🔥 

I think MariaDB works a little different than MySQL in this case, see the docs:
- https://mariadb.com/kb/en/create-user/#resource-limit-options

To prove this, I've created a test user:

```sql
MariaDB [(none)]> CREATE user foo IDENTIFIED by 'MariaDB11!' WITH MAX_USER_CONNECTIONS 0;
Query OK, 0 rows affected (0.019 sec)

MariaDB [(none)]> select User, max_connections, max_user_connections from mysql.user where User='foo';
+------+-----------------+----------------------+
| User | max_connections | max_user_connections |
+------+-----------------+----------------------+
| foo  |               0 |                    0 |
+------+-----------------+----------------------+
1 row in set (0.000 sec)
```

According to the docs, this user will have unlimited connections. Therefore, I think it makes more sense to keep the existing defaults and limit the connections by default. We making an extensive usage of the `User` CR  internally in the operator and I wouldn't like people to see the operator as the reason to waste connections in their `MariaDB`.

The current changes not only make the field mutable but also enable the user to set it to `0` if they want to, this way they can opt-in for unlimited connections. So I think we are good!

What do you think?

hedgieinsocks:
```

**Comment by hedgieinsocks** [comment]
I've played a bit, and the results show that we don't do anything different from oracle. Any new user is created with  `max_user_connections` set to `0` if not implicitly specified otherwise. And it does fallback to the global `max_user_connections` value.

```sql
MariaDB [(none)]> SELECT @@GLOBAL.max_user_connections;
+-------------------------------+
| @@GLOBAL.max_user_connections |
+-------------------------------+
|                             5 |
+-------------------------------+
1 row in set (0.001 sec)

MariaDB [(none)]> CREATE USER 'fox'@'%' IDENTIFIED BY 'qwerty';
Query OK, 0 rows affected (0.005 sec)

MariaDB [(none)]> SELECT user, max_user_connections FROM mysql.user WHERE User = 'fox';
+------+----------------------+
| User | max_user_connections |
+------+----------------------+
| fox  |                    0 |
+------+----------------------+
1 row in set (0.002 sec)
```

```sh
mysql@mariadb-0:/$ mariadb-slap --skip-ssl -u fox -pqwerty --concurrency=10 --number-of-queries=1000 --auto-generate-sql --iterations=1
mariadb-slap: Error when connecting to server: 1203 User fox already has more than 'max_user_connections' active connections
mariadb-slap: Error when connecting to server: 1203 User fox already has more than 'max_user_connections' active connections
Benchmark
        Average number of seconds to run all queries: 1.043 seconds
        Minimum number of seconds to run all queries: 1.043 seconds
        Maximum number of seconds to run all queries: 1.043 seconds
        Number of clients running queries: 10
        Average number of queries per client: 100
```

_Off-topic: I was surprised to learn that by default, the operator starts MariaDB with `--max-user-connections=0` and it is impossible to change this value globally afterward without a restart (ERROR 1290). Only if we set `max_user_connections=5` into `my.conf` it becomes dynamic._

But I digress, I'm perfectly OK with keeping the default limit as 10. If you are fine with my PR as it is, I'm triple fine with that :grinning:

mmontes11:
Hey there @hedgieinsocks ! 

Thanks for your input. I suggest we merge it as it is, ship it in this week's release, and tackle the defaults in a separate issue:
- https://github.com/mariadb-operator/mariadb-operator/issues/922

I've linked the relevant comments in the issue above as context.

Thanks!

PR #932, opened 2024-10-10
fix: exclude # symbol from password generator

### LINKED ISSUE #716: [Bug] Some generated metrics passwords do not work
**Documentation**

**Describe the bug**

It seems that mariadb-operator 0.0.29 can generate metrics passwords that the mysqld-exporter (v0.15.1) does not accept / cannot handle. I had a wordpress deployment where metrics did not work. All I could see in the logs were:

```
...
ts=2024-07-12T08:18:44.442Z caller=exporter.go:152 level=error msg="Error pinging mysqld" err="Error 1045 (28000): Access denied for user 'mariadb-metrics'@'10.42.0.232' (using password: YES)"
ts=2024-07-12T08:18:46.462Z caller=exporter.go:152 level=error msg="Error pinging mysqld" err="Error 1045 (28000): Access denied for user 'mariadb-metrics'@'10.42.0.232' (using password: YES)"
...
```

On the other hand, I had an almost identical mariadb deployment where metrics was working. After some digging it seems that the wordpress deployment had a metrics password containing a # character. I deleted the password secret to generate a new one and when I got a password without any # character, metrics scraping seems to work again.

I'm not sure which characters are problematic. But it seems that most "odd" characters are OK. But I never got it working when a # was included in the password.

**Expected behaviour**

Generated metrics passwords should not contain illegal(?) / problematic characters.

1. Deploy a mariadb with metrics enabled.
2. Check what characters the generated metrics secret contain. If it does not contain a #, delete it and let a new one be generated.
3. Check if metrics is working or not.

**Debug information**

**Environment details**:
- Kubernetes version: 1.29.6
- Kubernetes distribution: k3s
- mariadb-operator version: 0.0.29
- Install method: helm (mariadb-operator chart 0.29.0)

am6737:
- \# is treated as a comment in  configuration files, leading to failed parsing when the password contains this character.

The # symbol has been excluded from the global password generator to prevent these issues with mysqld-exporter.

Closes #716

### DIFF
--- a/pkg/controller/secret/controller.go
+++ b/pkg/controller/secret/controller.go
@@ -22,7 +22,7 @@ type SecretReconciler struct {
 
 func NewSecretReconciler(client client.Client, builder *builder.Builder) (*SecretReconciler, error) {
 	generator, err := password.NewGenerator(&password.GeneratorInput{
-		Symbols: "~!@#$%^&*()_+-={}|[]:<>/",
+		Symbols: "~!@$%^&*()_+-={}|[]:<>/",
 	})
 	if err != nil {
 		return nil, fmt.Errorf("error creating password generator: %v", err)


mmontes11 (approved):
LGTM!  Thanks for the in-depth analysis of the issue and this contribution @am6737, very much appreciated! 🙏🏻

am6737:
> LGTM! Thanks for the in-depth analysis of the issue and this contribution @am6737, very much appreciated! 🙏🏻

Thanks a lot! I’m glad I could help.  🙌🏻

PR #1009, opened 2024-10-29
Populate secondary service endpoints with custom service ports

hedgieinsocks:
Hi, this is a follow-up fix for https://github.com/mariadb-operator/mariadb-operator/pull/912
It turned out that custom ports are not automatically injected to the endpoints for the secondary service, and we must do it manually.


mmontes11 (approved):
LGTM! 🚀

This slipped through my review of https://github.com/mariadb-operator/mariadb-operator/pull/912, thanks for fixing it 🙏🏻

PR #1029, opened 2024-11-05
ci: test-image{,-ent} display mariadb version

grooverdan:
because I pass sha hashes in CI sometimes and knowing the version is a good thing to know.


mmontes11 (approved):
LGTM! 🚀 👍🏻 

This would be super useful, thank you very much!

PR #1048, opened 2024-11-13
ci github workflows - optional mariadb_version

grooverdan:
This enables workflows to be triggered by sha hash, but provides the version so the operator knows what version it is working with.

Only required if the mariadb_image doesn't include a tag.


mmontes11 (approved):
LGTM! 

FYI: `MARIADB_ENTRYPOINT_VERSION` will be renamed to `MARIADB_DEFAULT_VERSION` at some point, as we are using the variable for purposes other than the entrypoint now. Will take into account this changes.

PR #1059, opened 2024-11-26
Bump entrypoints to include 11.7 (2024Q4p2)

grooverdan:


hack/get_entrypoint.sh
```diff
@@ -46,6 +46,17 @@ for VERSION_DIR in $(find mariadb-docker/ -maxdepth 1 -type d -regex 'mariadb-do
   fi
 done
 
+ENTRYPOINT_PATH=mariadb-docker/main-ubi
```
mmontes11, line 49: @grooverdan could you provide some context about this folder? Why getting the entrypoint from `main-ubi` and not from `main`? Are these development versions i.e. 11.8 currently?

Leaving a small comment would be appreciated.

mmontes11 (commented):
@grooverdan Thank you for the PR, LGTM! Before merging, could you please provide some context on the question I've left?

hack/get_entrypoint.sh
grooverdan: was mainly because the version was easier to parse out of the Dockerfile in the ubi dir vs the ubuntu version in its dockerfile.

mmontes11: Fair enough, thanks for the context.

Issue #1005, opened 2024-10-29
[Bug] Error 1130: Host '10.131.0.28' is not allowed to connect to this MariaDB server


djjudas21:
**Describe the bug**

I have deployed a MariaDB Galera Cluster with MaxScale on OpenShift and used the CRs to create some Databases, Users and Grants. The Operator log is full of repeating error messages like this (`10.131.0.28` is the operator pod IP) 

```
Error 1130: Host '10.131.0.28' is not allowed to connect to this MariaDB server
``` 

I've included examples of Database, User and Grant but I think it's the same fundamental error in each case. This also occurs for every namespace on the cluster that has CRs for these resources (about 40 of them).

### Database

```
[...]

PR #1083, opened 2024-12-16
FLUSH PRIVILEGES unnecessary for user/grant manipulation

grooverdan:
FLUSHING PRIVILEGES isn't required after GRANT/CREATE/DROP/ALTER USER or GRANT/REVOKE.

Removing this pattern.

Closes: #1005


mmontes11 (approved):
LGTM! 🙏🏻  Thanks for looking into this.

[MDEV-35663](https://jira.mariadb.org/browse/MDEV-35663) for context

PR #1109, opened 2025-01-02
Add the activeDeadlineSeconds attribute to the backup pod

Funk66:
The backup pod gets stuck while dumping the data from one of our clusters. It happens irregularly for about 1/3 of the hourly jobs. I have yet to figure out the issue - possibly a deadlock situation? In the meantime, I thought it'd be helpful if the pod gets removed automatically when the job hasn't succeeded in a reasonable amount of time. That way, the stuck pod isn't blocking the cronjob from scheduling new pods.
I'm opening this PR as a way of illustrating my request. I do not know if these code changes are complete and functional. I guess including a test for this newly added field would be desirable.


mmontes11:
> I have yet to figure out the issue - possibly a deadlock situation?

Just to let you know, we have merge a new change to prevent deadlocks while backups are being taken:
- https://github.com/mariadb-operator/mariadb-operator/pull/1127

This will be shortly released in `0.37.0`. 

Regarding next steps in your PR, you need to set the `activeDeadlineSeconds` in the backup `Job here:

https://github.com/mariadb-operator/mariadb-operator/blob/3625487e863ccd43d3b3abb05f293b290d2172a1/pkg/builder/batch_builder.go#L107

PR #1127, opened 2025-01-16
Prevent deadlocks on `Database` reconcile while transactions are running

### LINKED ISSUE #1126: [Bug] deadlocks on operator database creation attempts
**Documentation**
- [ x] I acknowledge that I have read the relevant [documentation](https://github.com/mariadb-operator/mariadb-operator/tree/main/docs).

**Describe the bug**

During database contents import, "CREATE DATABASE IF NOT EXISTS..." raise a deadlock

**Expected behaviour**

No Deadlocks!

1. run a big database sql import
2. wait for the operator to reconcile Database resource
3. deadlock, import fails

Tested on MariaDB 10.11 Galera clusters using a migration script from a spring + hibernate java app. The mariadb user does not have LOCK privilege, that's not an issue about using locks with galera clusters.

vixns:
fix #1126 

@mmontes11 this bug is really annoying, hope this pr can land in release 0.37

### DIFF
--- a/pkg/sql/sql.go
+++ b/pkg/sql/sql.go
@@ -390,7 +390,16 @@ type DatabaseOpts struct {
 }
 
 func (c *Client) CreateDatabase(ctx context.Context, database string, opts DatabaseOpts) error {
-	query := fmt.Sprintf("CREATE DATABASE IF NOT EXISTS `%s` ", database)
+	sql := fmt.Sprintf("SELECT EXISTS (SELECT 1 FROM INFORMATION_SCHEMA.SCHEMATA WHERE SCHEMA_NAME = '%s')", database)
+	row := c.db.QueryRowContext(ctx, sql)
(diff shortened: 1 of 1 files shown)


mmontes11 (approved):
LGTM! Thanks for raising this @vixns, this will be released in `0.37.0`

Some context about this: The `CreateDatabase` function is exclusively executed by the controller here:

https://github.com/mariadb-operator/mariadb-operator/blob/ef20d329551620d7b07ef6e0b2b2cd3846c12cfe/pkg/sql/sql.go#L392

With the new version of `CreateDatabase` we are no longer locking the entire database, instead only one row of the `INFORMATION_SCHEMA.SCHEMATA` table will be locked, something that will definetly help importing backups. Locking the entire database is kindof overkill for our case, since the `Database` reconciliation happens periodically.

Issue #1123, opened 2025-01-14
[Bug] MariaDB CR object deletion with ArgoCD


ifalex:
**Documentation**
- [ x] I acknowledge that I have read the relevant [documentation](https://github.com/mariadb-operator/mariadb-operator/tree/main/docs).

**Describe the bug**

When installing mariadb-operator through ArgoCD (therefore Helm in behind), the installation completes successfully, new MariaDB's CRs objects are created correctly and pods spawn without any issue. (Tried both ways with or without generated rootpassword)

```yaml
        apiVersion: k8s.mariadb.com/v1alpha1
        kind: MariaDB 
        metadata:
          name: mariadb-standalone
        spec:
          rootPasswordSecretKeyRef:
            name: mariadb-root
            key: password
            generate: true
          imagePullPolicy: Always
          port: 3306
          storage:
            size: 1Gi
          myCnf: |
            [mariadb]
            bind-address=0.0.0.0
            default_storage_engine=InnoDB
            binlog_format=row
            innodb_autoinc_lock_mode=2
            innodb_buffer_pool_size=1024M
            max_allowed_packet=256M
          resources:
            requests:
              cpu: 100m
              memory: 128Mi
```

The problem arises when I delete the Argo application holding the mariadb-operator. Due to the finalizers added by subsequent objects created by mariadb-operator (databases, users, grants) the object remains in Deleting state in ArgoCD for ever.

One workaround was to add:

```yaml
inheritMetadata:
    annotations:
        argocd.argoproj.io/sync-wave: \"1\"
```
on the MariaDB CR created and this way after 10 minutes or more the application gets deleted. But nevertheless this should not stay so long.

**Expected behaviour**

When deleted from ArgoCD mariadb-operator together with resources created should be deleted. Without any manual intervention in deleting finalizers.

**Steps to reproduce the bug**

1. Install MariaDb Operator with ArgoCD application together with one/multiple MariaDB resource.
2. Delete MariaDb application from ArgoCD
3. Observe that MariaDb will not be deleted from ArgoCd

**Debug information**
- Related object events:
```bash
kubectl get events --field-selector involvedObject.name=<mariadb-resource-name>

LAST SEEN   TYPE     REASON             OBJECT                             MESSAGE
16m         Normal   SuccessfulCreate   statefulset/mariadb-repl           create Claim storage-mariadb-repl-0 Pod mariadb-repl-0 in StatefulSet mariadb-repl success
[...]

PR #1146, opened 2025-01-31
fix: don't reconcile mariadb cr if it's marked for deletion

lsoica:
MariaDB CR should not be reconciled when marked for deletion.

Fixes https://github.com/mariadb-operator/mariadb-operator/issues/1123

Related to: https://mariadb-community.slack.com/archives/C06AWJBAC1H/p1736934983515949


mmontes11 (approved):
LGTM! Thanks for your contribution! 🙏🏻

PR #1172, opened 2025-02-19
Fix some funny mistake with 'supersedes'.

Uburro:


mmontes11 (approved):
Oops, sorry about that. LGTM! Thanks for the fix.

PR #1202, opened 2025-03-21
Fixed so webhook secrets are not deployed when using cert-manager

Xartos:
When upgrading from version 0.0.31 to 0.0.32 I noticed that the helm chart tried to add some secrets that was suppose to be managed by cert-manager.

So I found that in [this commit](https://github.com/mariadb-operator/mariadb-operator/commit/a4c2d1de5ee59a34d91eb07138bdae2ff7ef0e64#diff-0facff5d85ae2370e1c04b961141e7e6cf64bcdc329908205e9c35df1711004f) the `not` didn't follow to apply to the `.Values.webhook.cert.certManager.enabled` as it was previously.
So this should fix this issue.


mmontes11:
Superseded by https://github.com/mariadb-operator/mariadb-operator/pull/1203

I have added you as co-author, thanks for your contribution!

PR #1203, opened 2025-03-21
Remove `Secret` webhook template

mmontes11:
The webhook `Secret` template is currently created when cert-manager is enabled in the helm chart, as pointed out by @Xartos in  https://github.com/mariadb-operator/mariadb-operator/pull/1202. This is a bug, as the `Secret` is not needed, it is fully managed by cert-manager. 

There was a regression in the condition, and the `Secret` should be rendered when cert-controller is used to manage certificates, serving as template and importing it into the helm release. This implies that the `Secret` is owned by helm and therefore uninstalled with the helm release.

At this point, in the existing installations of the operator using cert-controller (default values), will not have the `Secret` imported into the release. If we change this behaviour, it will require the users to manually import it. Otherwise, the installation will result in the following error:

```bash
Error: UPGRADE FAILED: Unable to continue with update: Secret "mariadb-operator-webhook-ca" in namespace "default" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"
``` 

We have opted for not importing the `Secret`, remove the template, and implement the same behaviour as cert-manager i.e. not delete the `Secret` when uninstalling.


trunet:
I just found this bug upgrading from 0.31.0 to 0.32.0 as well. I believe this fix is on main, how should we be able to upgrade to 0.0.32? Shouldn't you generate a 0.0.32.1 or something to allow us to go to that version before proceeding to next version upgrades?

mmontes11:
Hey @trunet ! We are not backporting this fix for maintenance overhad reasons. We will be considering maintaining multiple versions, and backporting fixes, when reaching `v1.0.0`.

This will be released in `v0.38.0`

Xartos:
> I just found this bug upgrading from 0.31.0 to 0.32.0 as well. I believe this fix is on main, how should we be able to upgrade to 0.0.32? Shouldn't you generate a 0.0.32.1 or something to allow us to go to that version before proceeding to next version upgrades?

@trunet I downloaded the chart locally

```
helm pull mariadb-operator/mariadb-operator --untar --version 0.32.0
```
Applied the patch, installed that. Rinse and repeat for each version until you are up to date.
Not an optimal solution but it works.

Issue #1132, opened 2025-01-20
[Feature] Add command Parameter to Exporter Object


iammuho:
It would be helpful to allow the configuration of a custom command for the Exporter container in the MariaDB Operator. This would enable users to override default parameters for the mysqld-exporter, providing greater flexibility in customizing its behavior.

**Describe the solution you'd like**
The current implementation of the Exporter object does not provide a way to override the command of the mysqld-exporter container. This feature would be beneficial in scenarios where users need to:

- Adjust the metrics collection parameters for the exporter.
- Specify custom flags for the mysqld-exporter.
- Fine-tune the behavior of the exporter container for specific monitoring needs.

**Example command:**

```
 - /bin/mysqld_exporter
        - --collect.binlog_size
        - --collect.engine_innodb_status
        - --collect.global_status
        - --collect.global_variables
        - --collect.info_schema.clientstats
        - --collect.info_schema.innodb_metrics
        - --collect.info_schema.innodb_tablespaces
        - --collect.info_schema.processlist
        - --collect.info_schema.tables
        - --collect.info_schema.tablestats
        - --collect.info_schema.schemastats
        - --collect.info_schema.userstats
        - --collect.perf_schema.eventsstatementssum
        - --collect.perf_schema.file_events
        - --collect.perf_schema.file_instances
        - --collect.perf_schema.indexiowaits
        - --collect.perf_schema.memory_events
        - --collect.perf_schema.tableiowaits
        - --collect.perf_schema.tablelocks
        - --collect.slave_status
   ```

pschiffe:
I was just looking for this option.

pschiffe:
Related feature to this would be option to modify `relabelings` of `serviceMonitor`s, for example, I would like to add `node_name` label to mysql metrics.

PR #1217, opened 2025-03-26
Add Args to Exporter

hedgieinsocks:
Closes https://github.com/mariadb-operator/mariadb-operator/issues/1132

### DIFF
--- a/api/v1alpha1/base_types.go
+++ b/api/v1alpha1/base_types.go
@@ -795,6 +795,11 @@ type Exporter struct {
 	// +optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	ImagePullSecrets []LocalObjectReference `json:"imagePullSecrets,omitempty" webhook:"inmutable"`
+	// CollectorFlags is the list of mysqld-exporter collector flags.
+	// See available flags: https://github.com/prometheus/mysqld_exporter?tab=readme-ov-file#collector-flags
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:advanced"}
+	CollectorFlags []string `json:"collectorFlags,omitempty"`
 	// Port where the exporter will be listening for connections.
 	// +optional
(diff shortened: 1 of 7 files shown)


api/v1alpha1/base_types.go
```diff
@@ -795,6 +795,11 @@ type Exporter struct {
 	// +optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	ImagePullSecrets []LocalObjectReference `json:"imagePullSecrets,omitempty" webhook:"inmutable"`
+	// CollectorFlags is the list of mysqld-exporter collector flags.
+	// See available flags: https://github.com/prometheus/mysqld_exporter?tab=readme-ov-file#collector-flags
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:advanced"}
+	CollectorFlags []string `json:"collectorFlags,omitempty"`
```
mmontes11, line 802: Could we rename this to `Args`? Altough the most common use case will be defining [collector flags](https://github.com/prometheus/mysqld_exporter?tab=readme-ov-file#collector-flags) there are also [general flags](https://github.com/prometheus/mysqld_exporter?tab=readme-ov-file#general-flags) that the user could provide using this functionality.

mmontes11 (changes requested):
Hey @hedgieinsocks ! LGTM, this is something we are missing and that we are willing to support.

I've left a comment that aims to generalize this approach to pass any kind of flag. Could you take a look?

Also, do you mind adding a test here to provide that the args are propagated correctly?:
https://github.com/mariadb-operator/mariadb-operator/blob/main/pkg/builder/deployment_builder_test.go

Thanks for your contribution!

api/v1alpha1/base_types.go
hedgieinsocks: Sure, I've copied the new generic version from `Container` object.

mmontes11 (approved):
LGTM! Thanks for your contribution 🚀

Issue #1223, opened 2025-03-29
[Bug] Cannot delete MariaDB CR with propagation policy foreground


lsoica:
This is a follow up on #1146  and #1123 

#1146 fixed part of the issue, but there's more. When a MariaDB CR is removed with foreground propagation, the deletion gets stuck sometimes. This happens when the StatefulSet is delete first. In this case the "sql" CRs like user and database cannot be removed as they have a finalizer, and the MariaDB CR cannot be removed because of blockOwnerDeletion: true on the "sql" CRs.

This PR removes the finalizers from the "sql" CRs in case the MariaDB CR is marked for deletion.

PR #1224, opened 2025-03-29
fix: Add finalizer removal logic for MariaDB deletion in SqlFinalizer

lsoica:
When MariaDB CR is marked for deletion, the "sql" CRs should not keep finalizer.

Fixes https://github.com/mariadb-operator/mariadb-operator/issues/1223


mmontes11 (approved):
LGTM! This could potentially unlock deletion scenarios. One side note for anyone checking this in the future: if the `MariaDB` gets recreated, the SQL resources will get its finalizer back:
https://github.com/mariadb-operator/mariadb-operator/blob/6b65ec8c7cbfa6db70e0948551663e803a612532/pkg/controller/sql/finalizer.go#L43

We have plans to release a `0.38.1` patch version in ~1 week time

PR #1192, opened 2025-03-10
feat: support HostPath in VolumeSource to fix #1101

kevinvalk:
Kept it really simple to be inline with `CSIVolumeSource` implementation.

PR #1225, opened 2025-03-30
fix: HostPath not properly copied into StorageVolumeSource struct

kevinvalk:
I screwed up in #1192 as I forgot to add `HostPath` to `func (v StorageVolumeSource) ToKubernetesType()`...

This PR fixes that issue and also adds a test to avoid these kind of mistakes in the future. The test ensures that any field implemented in `StorageVolumeSource` is copied into the `corev1.VolumeSource` in the (backup) job and fails if this is not done. This is done using reflection to ensure the test fails when the `StorageVolumeSource` is modified without understanding anything else (like I did in #1192).


mmontes11 (approved):
LGTM! Sorry, this slipped through my initial review. With the new test in place, It will be unlekily that we don't detect this kindof situations in the future. Thank you very much!

We have plans to release a `0.38.1` patch version in ~1 week time

PR #1244, opened 2025-04-10
Add requeue-sql-max-offset option

hedgieinsocks:
If one hosts thousands of databases/users/grants, they would probably set `--requeue-sql=24h` but this would still cause a short but huge spike in reconcile queue depth. They could set `--requeue-sql=0` to completely disable requeue feature. But this PR would allow one to retain requeue without experiencing a hard hit for concurrent reconciles once in a while. It should look like this:

```sh
--requeue-sql=24h
--requeue-sql-max-offset=1h
```

### DIFF
--- a/pkg/controller/sql/controller.go
+++ b/pkg/controller/sql/controller.go
@@ -163,13 +163,14 @@ func (r *SqlReconciler) requeueResult(ctx context.Context, resource Resource, er
 		return ctrl.Result{Requeue: true}, nil
 	}
 	if resource.RequeueInterval() != nil {
+		requeueInterval := r.addRequeueIntervalOffset(resource.RequeueInterval().Duration)
 		if r.LogSql {
 			log.FromContext(ctx).V(1).Info("Requeuing SQL resource")
 		}
-		return ctrl.Result{RequeueAfter: resource.RequeueInterval().Duration}, nil
+		return ctrl.Result{RequeueAfter: requeueInterval}, nil
 	}
 	if r.RequeueInterval > 0 {
-		requeueInterval := r.requeueIntervalWithOffset()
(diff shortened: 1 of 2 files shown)


pkg/controller/sql/controller.go
```diff
@@ -160,14 +169,22 @@ func (r *SqlReconciler) requeueResult(ctx context.Context, resource Resource, er
 		return ctrl.Result{RequeueAfter: resource.RequeueInterval().Duration}, nil
 	}
 	if r.RequeueInterval > 0 {
+		requeueInterval := r.requeueIntervalWithOffset()
 		if r.LogSql {
 			log.FromContext(ctx).V(1).Info("Requeuing SQL resource")
 		}
-		return ctrl.Result{RequeueAfter: r.RequeueInterval}, nil
+		return ctrl.Result{RequeueAfter: requeueInterval}, nil
```
mmontes11, line 176: For consistency, I think we should probably add the offset here as well:
- https://github.com/mariadb-operator/mariadb-operator/blob/b63954b40fd38d7913c46417dfb0a0b67b8b9e4f/pkg/controller/sql/controller.go#L160

This applies to `requeueInterval` provided in the CR, whereas the one here is the default set by the controller. One could always leave the offset to 0 (default) without affecting the current behaviour.

mmontes11 (changes requested):
Hey, thanks for bringing this up @hedgieinsocks . The idea looks legit, this could potentially prevent spikes when reconciling a large number of resources.

Left a suggestion.

mmontes11 (approved):
LGTM! Thanks a lot for this contribution @hedgieinsocks 🙏🏻 

I'm planning to release a patch version soon with some fixes, this will be included.

PR #1253, opened 2025-04-17
Fix special replica probe

hedgieinsocks:
Fix bug in replication probe introduced in 0.38.0 when we dropped custom replication connection name.

### DIFF
--- a/pkg/builder/container_builder.go
+++ b/pkg/builder/container_builder.go
@@ -34,7 +34,7 @@ var (
 				Command: []string{
 					"bash",
 					"-c",
-					"mariadb -u root -p\"${MARIADB_ROOT_PASSWORD}\" -e \"SELECT 1;\"",
+					"mariadb-admin -u root -p\"${MARIADB_ROOT_PASSWORD}\" ping",
 				},
 			},
 		},
--- a/pkg/builder/container_builder.go
+++ b/pkg/builder/container_builder.go
@@ -34,7 +34,7 @@ var (
 				Command: []string{
 					"bash",
 					"-c",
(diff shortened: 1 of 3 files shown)


mmontes11 (changes requested):
Hey @hedgieinsocks ! Thanks for the PR.

> fix bug in replication probe introduced in 0.38.0 when we dropped custom replication connection name

LGTM. Happy to merge

> use mariadb-admin ping for probes

Although this seems like an improvement, the probes aren't as accurate as a would like. Could we discard this and discuss it in a separate issue?

I have something in mind: introduce a sidecar container to resolve probes based on the Pod role i.e. primary or replica:
- If primary, the probe would do a `SELECT 1;`  or check any other more accurate internal
- If replica, the probe would check the replication lag and compare it with the threshold available in the CR.

See the galera probes as an example:
https://github.com/mariadb-operator/mariadb-operator/blob/main/pkg/galera/agent/handler/probe.go

Ideally, network traffic should not be forwarded to lagged replicas, something not currently managed in the existing replication probes.

Also, there are other advantages:
- Being written in Go allows to introduce unit tests of the handlers
- We could reuse our [autoUpdateDataPlane feature ](https://github.com/mariadb-operator/mariadb-operator/blob/main/docs/UPDATES.md#data-plane-updates)to upgrade sidecar containers. Not expecting to introduce many changes in the sidecar once we get the right probes though, that was the case with galera.

Let me know what you think!

hedgieinsocks:
Thank you for such a detailed reply. I have left only the fix for the current replica status check. As for your suggestion to refactor probes, it sounds cool, but unfortunately I have nothing more to contribute to it for now.

mmontes11 (approved):
LGTM! Thanks for your contribution @hedgieinsocks !

PR #1256, opened 2025-04-18
fix(doc): incorrect Maxscale default usernames

harrytang:


mmontes11 (approved):
Thanks for raising this @harrytang !

Issue #1252, opened 2025-04-17
[Feature] Please make "api" its own Go module


dtantsur:
**Is your feature request related to a problem? Please describe.**
I'm trying to use the api module in my code. Since it's a part of the entire mariadb-operator:

1. It's easy to hit conflicting versions of controller-runtime
2. I get a lot of transient dependencies that I don't really need
3. My dependency tree is now so large that golangci-lint goes OOM on a 16G laptop.

**Describe the solution you'd like**
Make the `api` directory a module of its own ([example](https://github.com/metal3-io/ironic-standalone-operator/commit/0b1a1542f4b6efd3cf843058bf6ec7244c46517a))

**Describe alternatives you've considered**
I guess I could use an untyped client but that's significantly less developer-friendly.
You could also provide a separate client but that's more work for you.

**Environment details**:
Not relevant

**Additional context**

mmontes11:
Hey @dtantsur ! 

That sounds like a legit request, and it seems like a common practise that projects like prometheus-operator are following:

main go.mod:
https://github.com/prometheus-operator/prometheus-operator/blob/b8e2df1c092b464deee69b178451a7f068a9ffdb/go.mod#L24

api go.mod: 
https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/apis/monitoring/go.mod

We are aware of multiple projects, including yours, that are creating CRs programatically, so it completely makes sense to split. That said, `controller-runtime` is still required as dependency in the api module, which will still create some friction with potential conflicts, but bumping it in the api pkg will be significantly easier compared to main pkg, as it sometimes requires adaptations due to breaking changes, like [the one recently introduced in the webhooks](https://github.com/mariadb-operator/mariadb-operator/pull/1204).

Contributions are welcomed! Will try to book some time on the upcoming release if no PRs are raised. Also, will create some documentation around this, as it will imply changes when consuming the api as library.

PR #1271, opened 2025-05-05
Add typed clientset and associated code generators

### LINKED ISSUE #1270: [Feature] Provide go clientset for operator resources
Many projects, such as [Knative](https://github.com/knative/serving/tree/main/pkg/client) and [cert-manager](https://github.com/cert-manager/cert-manager/tree/master/pkg/client) provide typed go clients to make it easy to integrate them into other projects. I'm currently working on a project where I would like to be able to use the same with the MariaDB Operator types so that I can easily manage mariadb resources in my program.

Use the https://github.com/kubernetes/code-generator package to automatically generate a clientset for the mariadb operator resources.

**Additional context**
I'm happy to contribute a PR for this

Cali0707:
Fixes #1270 

This PR uses kube_codegen from https://github.com/kubernetes/code-generator to generated a typed clientset.

In order for the codegen to run properly, I had to update the package path in `api` from `api/v1alpha1` to `api/mariadb/v1alpha1`, which led to lots of changes in import paths elsewhere. I _think_ I got those all correct.

cc @mmontes11


mmontes11:
Hey there @Cali0707 ! 

Looking good, I think you are on the right track. Could you take a look at my comments? Also, it seems like there are some unused imports which is making the CI fail.

As requested in https://github.com/mariadb-operator/mariadb-operator/issues/1252, I think it would be great to have a dedicated go package for the client. Otherwise, someone importing the client will need to import all our dependencies and potentially end up with conflicts. This is a very common practise, please take a look at the typed client from prometheus operator:
- https://github.com/prometheus-operator/prometheus-operator/tree/main/pkg/client

> where should I include a sample code snippet on how to use the typed clientset?

We can create a `examples/client/*.go` which I will include in the Go packages documentation in the upcoming release.

internal/webhook/v1alpha1/backup_webhook_test.go
```diff
@@ -1,9 +1,10 @@
 package v1alpha1
 
 import (
+	v1alpha2 "github.com/mariadb-operator/mariadb-operator/api/mariadb/v1alpha1"
```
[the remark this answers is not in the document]
Cali0707: For sure, I'm not even totally sure where this import came from - I think IntelliJ may have tried this when I refactored the file paths 🤷

pkg/client/go.mod
```diff
@@ -0,0 +1,71 @@
+module github.com/mariadb-operator/mariadb-operator/pkg/client
+
+go 1.24.2
+
+replace github.com/mariadb-operator/mariadb-operator => ../../
+
+require (
+	github.com/mariadb-operator/mariadb-operator v0.0.0-00010101000000-000000000000
```
mmontes11, line 8: Why this import? I believe this would bring all the dependencies from the root project here, something that defeats the purpose of having a dedicated module for the client. Prometheus operator client is not making this import:
https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/client/go.mod

examples/client/main.go
```diff
@@ -0,0 +1,64 @@
...
+	"context"
+	"fmt"
+
+	mariadbv1alpha1 "github.com/mariadb-operator/mariadb-operator/api/mariadb/v1alpha1"
+	"github.com/mariadb-operator/mariadb-operator/pkg/client/clientset/versioned"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/rest"
+)
+
+func main() {
```
mmontes11, line 13: Thank you for this, looks great! 🙏🏻

go.mod
```diff
@@ -26,9 +27,9 @@ require (
 	github.com/stretchr/testify v1.10.0
 	go.uber.org/zap v1.27.0
 	golang.org/x/sync v0.12.0
-	k8s.io/api v0.32.3
-	k8s.io/apimachinery v0.32.3
-	k8s.io/client-go v0.32.3
+	k8s.io/api v0.33.0
```
mmontes11, line 30: Can we bump the Kubernetes libraries in a different PR? This requires more changes, like updating the KIND versions for the tests, among other things

mmontes11 (changes requested):
Thanks for the changes @Cali0707, looking promising.

Left a couple of comments. It also seems like the CI is failing. Could take a look? 🙏🏻

PR #1272, opened 2025-05-10
Fix most outstanding typos

hedgieinsocks:
While these are purely cosmetic changes, eye-catching typos can misrepresent the quality of the project, so we could try to slowly get rid of them. Also, we could even add `typos`  [action](https://github.com/crate-ci/typos/blob/master/docs/github-action.md) for PRs.


mmontes11:
Alright @hedgieinsocks that makes sense. With the `.typos.toml`  and the GitHub action in place, I think we will be good.

hedgieinsocks:
Nice! Let's see how the action behaves on prs, and tune/fix if anything makes it stumble 🙏

mmontes11:
cc @hedgieinsocks FYI just rebased this branch, works like a charm:
https://github.com/mariadb-operator/mariadb-operator/actions/runs/15062979029/job/42341550217

PR #1279, opened 2025-05-12
feat: overwrite tolerations and nodeSelector in the restore job when using bootstrapFrom

kevinvalk:
Builds on #1192, specifically, when using the `bootstrapFrom` mechanism when using `hostPath` backups. In that case, you want to ensure the restore job is ran from the correct node that contains the backups.

P.S. In the meantime you could set the `nodeSelector` on the mariadb spec as the recovery job "inherits" the node selector (and tolerations).


mmontes11 (changes requested):
LGTM!

After merging https://github.com/mariadb-operator/mariadb-operator/pull/1272, some conflicts arised, sorry for the inconvenience. Could you take a look?

mmontes11 (approved):
LGTM! Thanks for the contribution 🚀

PR #1287, opened 2025-05-28
[Replication] Add optional wait before automatic failover

hedgieinsocks:
The idea is to somehow make replication more resilient to node flapping events e.g. when rke2-agent restarts and kubelet flaps the mariadb master pod into not ready state for several seconds due to volume remount. 

So perhaps with this we could optionally wait [here](https://github.com/mariadb-operator/mariadb-operator/blob/8671ab54e953e2611e377c0dab42ab5e9d9741b6/internal/controller/pod_controller.go#L66) and recheck whether the pod is still not ready. And if it is ready, then return.


internal/controller/pod_controller.go
```diff
@@ -64,6 +65,15 @@ func (r *PodController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 			return ctrl.Result{Requeue: true}, nil
 		}
 	} else {
+		if automaticFailoverDeferral := mariadb.GetAutomaticFailoverDeferral(); automaticFailoverDeferral > 0 {
```
mmontes11: This logic is specific to the replication topology. Please move it here:
https://github.com/mariadb-operator/mariadb-operator/blob/e286069b1c104a914b391d8bc3f96c740b80df97/internal/controller/pod_replication_controller.go#L64

You will need to adapt the `PodReadinessController` signature to return a tuple:
https://github.com/mariadb-operator/mariadb-operator/blob/e286069b1c104a914b391d8bc3f96c740b80df97/internal/controller/pod_controller.go#L18

internal/controller/pod_controller.go
```diff
@@ -64,6 +65,15 @@ func (r *PodController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
...
+			if podReadyLastTransitionTime := mariadbpod.PodReadyLastTransitionTime(&pod); !podReadyLastTransitionTime.IsZero() {
```
mmontes11: I'm not sure if we can rely on the `Ready` `lastTransitionTimeStamp`.  Tying such a critical feature to something outside our control might not be a good idea.

Instead, I suggest we create a new  `status` field named `currentPrimaryFailingSince`, which can be set here when the primary `Pod` transtitions from ready to not ready state and `currentPrimaryFailingSince` is not set:
https://github.com/mariadb-operator/mariadb-operator/blob/e286069b1c104a914b391d8bc3f96c740b80df97/internal/controller/pod_replication_controller.go#L64

Using `status.currentPrimaryFailingSince` and the new `spec.replication.automaticFailoverDeferral`, we can decide if we should update the `spec.replication.primary.podIndex` here, or alternatively requeuing:
https://github.com/mariadb-operator/mariadb-operator/blob/e286069b1c104a914b391d8bc3f96c740b80df97/internal/controller/pod_replication_controller.go#L73

Finally, we also need to make sure that we reset the `status.currentPrimaryFailingSince`  when te primary Pod transitions from not ready to ready state here:
https://github.com/mariadb-operator/mariadb-operator/blob/e286069b1c104a914b391d8bc3f96c740b80df97/internal/controller/pod_replication_controller.go#L43

internal/controller/pod_controller.go
hedgieinsocks: If we move the logic to pod_repliction_controller.go, then we will have to return an error in `ReconcilePodNotReady()` when deferral time has not yet passed?

mmontes11: If the deferal time is set, and it has not yet passed, we can requeue and log a message.
If deferal time is not set, it should work as it is currently.

hedgieinsocks: ah, now I see what you meant by tuple for PodReadinessController, thanks!

mmontes11: Sorry, I should have been more explicit: tuple = `(ctrl.Result, error)`

hedgieinsocks: Alternatively, we could return some special error, e.g. `ErrDeferAutomaticFailover` and requeue on it to avoid refactoring of the underlying functions and deviation from galera analogs. Dunno, which is better, wdyt?

mmontes11: That would work as well. `PodReplicationController` would return `ErrDeferAutomaticFailover` and `PodController` would handle it, keeping the `PodReadinessController` as it is.

hedgieinsocks:
> I think we are on the right track, left some suggestions to make the implementation more robust and maintainable.

Preliminary, it seems that the suggested custom status approach from your last comment works as expected in my test env. But it required touching a few important things here and there, so quite some tests will be required once we are settled with the code itself. 

```sh
# trace related logs with debug level logging while performing manual or automatic failovers and updates with ReplicasFirstPrimaryLast.
❯ stern MARIADB_OPERATOR_POD -e ca-bundle -e cert -i replication -i switchover -i update
```

api/v1alpha1/mariadb_replication_types.go
```diff
@@ -88,6 +88,10 @@ type PrimaryReplication struct {
 	// +optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:booleanSwitch"}
 	AutomaticFailover *bool `json:"automaticFailover,omitempty"`
+	// AutomaticFailoverDeferral indicates the duration before re-evaluating the need for an automatic primary failover.
```
mmontes11:
```suggestion
	// AutomaticFailoverDelay indicates the duration before performing an automatic primary failover. By default, no extra delay is added.
```

internal/controller/mariadb_controller_update.go
```diff
@@ -79,17 +79,19 @@ func (r *MariaDBReconciler) reconcileUpdates(ctx context.Context, mdb *mariadbv1
...
-		return result, err
-	}
-
 	primaryPod := podsByRole.primary
 	if podpkg.PodUpdated(&primaryPod, stsUpdateRevision) {
 		logger.V(1).Info("Primary Pod up to date", "pod", primaryPod.Name)
 		return ctrl.Result{}, nil
 	}
 
+	if err := r.waitForConfiguredReplica(mdb, logger); err != nil {
```
mmontes11, line 88: What motivated moving this block after `PodUpdated`? I'm unable to see the reason, maybe missing something

internal/controller/pod_controller.go
```diff
@@ -64,6 +65,10 @@ func (r *PodController) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 		}
 	} else {
 		if err := r.podReadinessController.ReconcilePodNotReady(ctx, pod, mariadb); err != nil {
+			if errors.Is(err, ErrDeferAutomaticFailover) {
+				log.FromContext(ctx).V(1).Info("Defer reconciling Pod in non Ready state", "pod", pod.Name)
```
mmontes11:
```suggestion
				log.FromContext(ctx).V(1).Info("Delaying primary switchover. Skipping reconciliation of Pod in non Ready state", "pod", pod.Name)
```

internal/controller/pod_replication_controller.go
```diff
@@ -13,12 +15,17 @@ import (
...
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+var (
+	ErrDeferAutomaticFailover = errors.New("deferring automatic failover")
```
mmontes11:
```suggestion
	ErrDelayAutomaticFailover = errors.New("delaying automatic failover")
```

mmontes11 (changes requested):
I have thoroughly tested this, it works pretty well. Congratulations 🎖️ 

I've requested some cosmetic changes, but other than that LGTM!

Thank you for this contribution, another step closer to GA replication.

internal/controller/mariadb_controller_update.go
hedgieinsocks: When I was changing the naming and logic of  `waitForConfiguredReplication()` => `waitForConfiguredReplica()` I had a bit of a hard time to recall and understand what was going on there. So I added extra one-line comments to the invocations of `r.waitForConfiguredReplica()` and `r.triggerSwitchover()` as a helping hand for whoever reads it in the future. And then I noticed that primary block sitting between them, and realized that since we should always strive to return early, we should first check if primary is up to date, and only if not, start waiting for a configured replica.

mmontes11: Return early, always a good idea. That makes more sense now, thanks for the explanation!

PR #1288, opened 2025-05-29
[Replication] Add verbosity to skipped switchover steps

hedgieinsocks:
This should make it a bit easier to investigate botched failovers. 

For example, when my node is flapping for a few seconds (rke2 agent restart), with 2 replica cluster with automatic failover, I see that in 50% of cases the cluster ends up with 2 masters: the new one and the old one because `changePrimaryToReplica()` sees the master is not ready and skips this step. And a few seconds later, the restarted kubelet gets hold of things, and the old master pod becomes ready. 

I figured it out only by noticing the absence of `Change primary to be a replica` log entry. With the suggested change, it would be much easier to understand what happened.


mmontes11 (approved):
LGTM! Hopefully this sheds some light during the switchover troubleshooting, something that will help maturing the replication feature to GA.

PR #1300, opened 2025-06-12
make embed-entrypoint for 2025q2

grooverdan:
I've removed out the 10.5 instance.

10.6 left as is, though noting the next 10.6 instance will be jammy based, mariadb:10.6-jammy is there now if that needs to be tested.


mmontes11 (approved):
Thanks for raising this @grooverdan ! LGTM

I don't see any outstanding changes in the 10.6 entrypoint, it should be fine. FYI: However, there seems to be something in `mariadb:10.6-jammy` consistently making the replication tests unhappy:
https://github.com/mariadb-operator/mariadb-operator/actions/runs/15604929192/job/43954139159

grooverdan:
> Thanks for raising this @grooverdan ! LGTM
> 
> I don't see any outstanding changes in the 10.6 entrypoint, it should be fine. FYI: However, there seems to be something in `mariadb:10.6-jammy` consistently making the replication tests unhappy: https://github.com/mariadb-operator/mariadb-operator/actions/runs/15604929192

quite weird. I think I saw the focal based 10.6 of the same server version pass, entrypoint is the same.

PR #1306, opened 2025-06-25
Release 25.08.0

mmontes11:


docs/releases/RELEASE_25.08.0_HEADER.md
```diff
@@ -0,0 +1,147 @@
...
+
+We are introducing `mariadb-cluster`, a new helm chart that simplifies the deployment of a `MariaDB` cluster and its associated CRs managed by the operator. It allows you to manage all CRs in a single helm release, handling their relationships automatically so you don't need to configure the references manually.
+
+Refer to the [helm documentation](https://github.com/mariadb-operator/mariadb-operator/blob/main/docs/HELM.md) for further details.
+
+Kudos to @hedgieinsocks for this initiative!
+
+### Replication improvements
+
+We have made some progress towards making our asynchronous replication feature GA. Refer to the PRs below for more details:
```
hedgieinsocks: Perhaps we should say `semi-sync`, it has already caused a tiny confusion :smile:  https://mariadb-community.slack.com/archives/C06AWJBAC1H/p1751016401204439

mmontes11: We should be supporting both semi-sync and fully async in further releases, so I think I will rename this to just "replication" to avoid causing confusion.

Issue #199, opened 2023-08-16
[Feature] Manage existing Database


aschaber1:
Hello :)

**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
I have a MariaDB provided by a DBaaS. I'd like to be able to manage Users, Grants and logical Databases with the operator. I might just be missing the right pieces from the documentation.

**Describe alternatives you've considered**
Managing the Users Grants and DBs as well in Terraform.

**Additional context**
We're running EKS on AWS and the DBaaS is a RDS Database. But I believe this issue would apply in any environment.

**Environment details**:
- Kubernetes version: 1.26
- mariadb-operator version:  v0.0.19
- Install method: helm
- Install flavour: recommended

Thank you in advance for any pointers.

mmontes11:
Hey @aschaber1 ! Thanks for this suggestion. This is currently not supported but it is definetly interesting 👍🏻 

We will need to adapt our SQL resources to add an `externalMariaDb` field. This will basically replace the current `mariaDbRef`  that refers to local `MariaDB` instances provisioned by the `mariadb-operator` in the same cluster:

```yaml
apiVersion: mariadb.mmontes.io/v1alpha1
kind: User
metadata:
  name: user
spec:
  externalMariaDb:
    host: mydb.123456789012.us-east-1.rds.amazonaws.com.
    user: admin
    passwordSecretKeyRef:
      name: admin
      key: password
    caSecretKeyRef:
      name: aws-rds-ca
      key: ca.crt
  passwordSecretKeyRef:
    name: user
    key: password
  maxUserConnections: 20
```

```yaml
apiVersion: mariadb.mmontes.io/v1alpha1
kind: Grant
metadata:
  name: grant
spec:
  externalMariaDb:
    host: mydb.123456789012.us-east-1.rds.amazonaws.com.
    user: admin
    passwordSecretKeyRef:
      name: admin
      key: password
    caSecretKeyRef:
      name: aws-rds-ca
      key: ca.crt
  privileges:
    - "SELECT"
    - "INSERT"
    - "UPDATE"
  database: "*"
  table: "*"
  username: user
  grantOption: true
```

```yaml
apiVersion: mariadb.mmontes.io/v1alpha1
kind: Database
metadata:
  name: data-test
spec:
  externalMariaDb:
    host: mydb.123456789012.us-east-1.rds.amazonaws.com.
    user: admin
    passwordSecretKeyRef:
      name: admin
[...]

PR #1311, opened 2025-06-30
Adding support to external MariaDB server

wfelipew:
Adding support to external MariaDB server as requested on https://github.com/mariadb-operator/mariadb-operator/issues/199.

It supports:
* Connections
* Users
* Grants
* Databases
* Backups
* SQLJobs


mmontes11:
Thanks for raising this @wfelipew . Will take a closer look in the following weeks and let you know. Do you mind rebasing with main?

mmontes11:
Hey @wfelipew ! As we get close to the [25.08 release](https://github.com/mariadb-operator/mariadb-operator/pull/1306), I'm finally finding some time to review this. Would you mind resolving the conflicts and rebasing?

wfelipew:
> Hey @wfelipew ! As we get close to the [25.08 release](https://github.com/mariadb-operator/mariadb-operator/pull/1306), I'm finally finding some time to review this. Would you mind resolving the conflicts and rebasing?

done

mmontes11:
Thanks for the effort @wfelipew ! Looks promising, will take a look tomorrow.

api/v1alpha1/mariadb_types.go
```diff
@@ -1036,6 +1155,125 @@ func (m *MariaDBList) ListItems() []client.Object {
...
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:shortName=emdb
+// +kubebuilder:subresource:status
+// +kubebuilder:subresource:scale:specpath=.spec.replicas,statuspath=.status.replicas
+// +kubebuilder:printcolumn:name="Ready",type="string",JSONPath=".status.conditions[?(@.type==\"Ready\")].status"
+// +kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.conditions[?(@.type==\"Ready\")].message"
+// +kubebuilder:printcolumn:name="Primary",type="string",JSONPath=".status.currentPrimary"
+// +kubebuilder:printcolumn:name="Updates",type="string",JSONPath=".spec.updateStrategy.type"
```
mmontes11: These two annotations can be omitted for `ExternalMariaDB`: We don't know what the primary is nor we are handling updates.

api/v1alpha1/mariadb_types.go
```diff
@@ -1036,6 +1155,125 @@ func (m *MariaDBList) ListItems() []client.Object {
...
+
+// IsReady indicates whether the External MariaDB instance is ready
+func (m *ExternalMariaDB) IsReady() bool {
+	return meta.IsStatusConditionTrue(m.Status.Conditions, ConditionTypeReady)
+}
+
+// Get image pull policy
+func (m *ExternalMariaDB) GetImagePullPolicy() corev1.PullPolicy {
+	pp := corev1.PullIfNotPresent
+	return pp
```
mmontes11:
```suggestion
	return corev1.PullIfNotPresent
```
We are returning a copy, it is safe to return `corev1.PullIfNotPresent` directly unless I am missing something

api/v1alpha1/mariadb_types.go
```diff
@@ -1036,6 +1155,125 @@ func (m *MariaDBList) ListItems() []client.Object {
...
+}
+
+// Get image pull secrets
+func (m *ExternalMariaDB) GetImagePullSecrets() []LocalObjectReference {
+	return nil
```
mmontes11:
```suggestion
	return nil // ExternalMariaDB uses official MariaDB images (publicly available) for the Backups
```

api/v1alpha1/mariadb_types.go
```diff
@@ -1036,6 +1155,125 @@ func (m *MariaDBList) ListItems() []client.Object {
...
+}
+
+// Get image
+func (m *ExternalMariaDB) GetImage() string {
+	return fmt.Sprintf("mariadb:%s", m.Status.Version)
```
mmontes11:
```suggestion
	return fmt.Sprintf("mariadb:%s", m.Status.Version) // // ExternalMariaDB uses official MariaDB images (publicly available) for the Backups
```

api/v1alpha1/mariadb_types.go
```diff
@@ -1036,6 +1155,125 @@ func (m *MariaDBList) ListItems() []client.Object {
...
+}
+
+// IsTLSRequired indicates whether TLS is enabled and must be enforced for all connections.
+func (m *ExternalMariaDB) IsTLSRequired() bool {
+	// if !m.IsTLSEnabled() {
+	// 	return false
+	// }
+	// tls := ptr.Deref(m.Spec.TLS, TLS{})
+	// return ptr.Deref(tls.Required, false)
+	return false
```
mmontes11:
```suggestion
	return false // ExternalMariaDB does not make use of this, as it is a internal server setting
```

docs/EXTERNAL_MARIADB.md
```diff
@@ -0,0 +1,99 @@
+# TLS
```
mmontes11:
```suggestion
# External MariaDB
```

api/v1alpha1/mariadb_types.go
```diff
@@ -745,6 +786,37 @@ type MariaDBStatus struct {
...
+	// Conditions for the ExternalMariadb object.
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=status,xDescriptors={"urn:alm:descriptor:io.kubernetes.conditions"}
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+	// Version of the external MariaDB server
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=status
+	Version string `json:"version,omitempty"`
+
+	// // ReplicationStatus is the replication current state for each Pod.
```
mmontes11: I don't think we are handling this. Could we remove?

api/v1alpha1/mariadb_types.go
```diff
@@ -1036,6 +1155,125 @@ func (m *MariaDBList) ListItems() []client.Object {
 	return items
 }
 
+// IsHAEnabled indicates whether the MariaDB instance has Galera enabled
+func (m *ExternalMariaDB) IsGaleraEnabled() bool {
+	return false
```
mmontes11: `ExternalMariaDB` must be aware of Galera. Since we support `Backup`, not indicating that Galera is enabled, will cause `mariadb-dump` to use `LOCK TABLES`, something incompatible with Galera. See:
https://github.com/mariadb-operator/mariadb-operator/blob/3da8aa4d9cd306026293cb3a353c89b3b17c1de0/pkg/command/backup.go#L406

I suggest we add a `spec.galera.enabled=<bool>` flag to `ExternalMariaDB` to indicate that the external MariaDB has Galera enabled.

pkg/controller/sql/controller.go
```diff
@@ -185,14 +186,21 @@ func (r *SqlReconciler) addRequeueIntervalOffset(duration time.Duration) time.Du
 	return duration
 }
 
-func waitForMariaDB(ctx context.Context, client ctrlclient.Client, mdb *mariadbv1alpha1.MariaDB,
+func waitForMariaDB(ctx context.Context, client ctrlclient.Client, mdb interfaces.MariaDBGenericInterface,
```
mmontes11, line 189: I think we can leave the signature of this function as it is, with `*mariadbv1alpha1.MariaDB`:

```go
func waitForMariaDB(ctx context.Context, client ctrlclient.Client, mdb *mariadbv1alpha1.MariaDB,
	logSql bool) (ctrl.Result, error) {

	if mdb.GetObjectKind().GroupVersionKind().Kind == mariadbv1alpha1.ExternalMariaDBKind {
		return ctrl.Result{}, nil
	}

	healthy, err := health.IsStatefulSetHealthy(
		ctx,
		client,
		ctrlclient.ObjectKeyFromObject(mdb),
		health.WithDesiredReplicas(mdb.Spec.Replicas),
		health.WithPort(mdb.Spec.Port),
		health.WithEndpointPolicy(health.EndpointPolicyAll),
	)
	if err != nil {
		return ctrl.Result{}, err
	}
	if !healthy {
		if logSql {
			log.FromContext(ctx).V(1).Info("MariaDB unhealthy. Requeuing SQL resource")
		}
		return ctrl.Result{RequeueAfter: 1 * time.Second}, nil
	}
	return ctrl.Result{}, nil
}
``` 
`MariaDB` has already the `GetObjectKind()` method, so you could return early when it is a `ExternalMariaDBKind`. This implies that we can remove `GetReplicas` from the `ReplicationAwareInterface`, as this is the only place where it is used.

api/v1alpha1/mariadb_keys.go
```diff
@@ -65,6 +65,25 @@ func (m *MariaDB) TLSCABundleSecretKeyRef() SecretKeySelector {
 	}
 }
 
+// TLSCABundleSecretKeyRef defines the key selector for the TLS Secret trust bundle
+func (m *ExternalMariaDB) TLSCABundleSecretKeyRef() SecretKeySelector {
```
mmontes11: Would you mind movind these new keys to `external_mariadb_keys.go`? `mariadb_keys.go` is becoming lengthy

api/v1alpha1/mariadb_types.go
```diff
@@ -1018,6 +1090,53 @@ func (m *MariaDB) TLSClientNames() []string {
...
+}
+
+// Get image
+func (m *MariaDB) GetImage() string {
+	return m.Spec.Image
+}
+
+// Get MariaDB hostname
+func (m *MariaDB) GetHost() string {
+	if m.Replication().Enabled {
```
mmontes11:
```suggestion
	if m.IsHAEnabled() {
```
When any HA topology is enabled i.e. Galera or replication, we should target the primary `Pod` via the primary `Service`

pkg/sql/sql.go
```diff
@@ -136,37 +138,49 @@ func NewClient(clientOpts ...Opt) (*Client, error) {
...
-			if mariadb.IsHAEnabled() {
-				return statefulset.ServiceFQDNWithService(
-					mariadb.ObjectMeta,
-					mariadb.PrimaryServiceKey().Name,
-				)
-			}
-			return statefulset.ServiceFQDN(mariadb.ObjectMeta)
-		}()),
-		WithPort(mariadb.Spec.Port),
+	if mariadb.GetObjectKind().GroupVersionKind().Kind == mariadbv1alpha1.ExternalMariaDBKind {
```
mmontes11: With my previous suggestions in place, I think both implementations are homogeneus. Could we just do this and remove the conditionals?



```go
              opts = []Opt{
			WithUsername(mariadb.GetSUName()),
			WithPassword(password),
			WitHost(mariadb.GetHost()),
			WithPort(mariadb.GetPort()),
		}
```

In my opinion, this is why interfaces exists, to create more abstraction and avoid type conditions

api/v1alpha1/mariadb_types.go
```diff
@@ -1036,6 +1155,125 @@ func (m *MariaDBList) ListItems() []client.Object {
...
+	return m.Spec.Host
+}
+
+// Get MariaDB port
+func (m *ExternalMariaDB) GetPort() int32 {
+	return m.Spec.Port
+}
+
+// Get MariaDB replicas
+func (m *ExternalMariaDB) GetReplicas() int32 {
```
mmontes11: I think we can get rid of `GetReplicas`, both here and in the interface, see my other comment. `waitForMariaDB` seems the only reference and it can be avoided.

api/v1alpha1/mariadb_types.go
```diff
@@ -1036,6 +1155,125 @@ func (m *MariaDBList) ListItems() []client.Object {
...
+	return 1
+}
+
+// IsHAEnabled indicates whether the MariaDB instance has HA enabled (Always false for external MariaDB)
+func (m *ExternalMariaDB) IsHAEnabled() bool {
+	return false
+}
+
+// Get MariaDB Superuser name
+func (m *ExternalMariaDB) GetSUName() *string {
```
mmontes11: How about changing the interface to
```suggestion
func (m *ExternalMariaDB) GetSUName() string {
```
and default to the root user?
```bash
return ptr.Deref(m.Spec.Username, "root")
```

We are dereferencing this in an unsafe way in other places, see my other commend. With this in place, not only we have a reasonable default, but we avoid risk of panicking.

pkg/sql/sql.go
```diff
@@ -136,37 +138,49 @@ func NewClient(clientOpts ...Opt) (*Client, error) {
...
+		opts = []Opt{
+			WithUsername(*mariadb.GetSUName()),
```
mmontes11: This is might cause a panic, I suggest we default to "root` and return a string, see my other comment.

api/v1alpha1/mariadb_types.go
mmontes11: By changing this, we will ge able to use `GetHost` in `NewClientWithMariaDB`. The implementation is equivalent.

api/v1alpha1/mariadb_types.go
```diff
@@ -1018,6 +1090,53 @@ func (m *MariaDB) TLSClientNames() []string {
...
+	return statefulset.ServiceFQDN(m.ObjectMeta)
+}
+
+// Get MariaDB port
+func (m *MariaDB) GetPort() int32 {
+	return m.Spec.Port
+}
+
+// Get MariaDB replicas
+func (m *MariaDB) GetReplicas() int32 {
```
mmontes11, line 1053: I think we can get rid of `GetReplicas`, both here and in the interface, see my other comment,

api/v1alpha1/mariadb_types.go
```diff
@@ -688,6 +688,47 @@ type MariaDBSpec struct {
 	SecondaryConnection *ConnectionTemplate `json:"secondaryConnection,omitempty" webhook:"inmutable"`
 }
 
+const (
+	// ExternalMariaDBKind
+	ExternalMariaDBKind = "ExternalMariaDB"
+)
+
+// ExternalMariaDBSpec defines the desired state of an External MariaDB
+type ExternalMariaDBSpec struct {
```
mmontes11: Could we move these types to `external_mariadb_types.go` along with their tests to `external_mariadb_types_test,go`? It is getting lengthy and hard to follow.

api/v1alpha1/mariadb_types.go
```diff
@@ -1036,6 +1155,125 @@ func (m *MariaDBList) ListItems() []client.Object {
...
+	return m.Spec.Port
+}
+
+// Get MariaDB replicas
+func (m *ExternalMariaDB) GetReplicas() int32 {
+	return 1
+}
+
+// IsHAEnabled indicates whether the MariaDB instance has HA enabled (Always false for external MariaDB)
+func (m *ExternalMariaDB) IsHAEnabled() bool {
```
mmontes11: I think we can get rid of `IsHAEnabled`, both here and in te interface, see my other comment. `NewClientWithMariaDB` seems the only reference and we could avoid it by using `GetHost`

api/v1alpha1/mariadb_types.go
```diff
@@ -1036,6 +1155,125 @@ func (m *MariaDBList) ListItems() []client.Object {
...
+}
+
+// Get MariaDB Superuser name
+func (m *ExternalMariaDB) GetSUName() *string {
+	return m.Spec.Username
+}
+
+// Get MariaDB Superuser credentials
+func (m *ExternalMariaDB) GetSUCredential() SecretKeySelector {
+	return *m.Spec.PasswordSecretKeyRef
```
mmontes11: I think dereferencing here is risky. There is no default, and the user might have not set it.

Could we change the signature to?
```go
func (m *ExternalMariaDB) GetSUCredential() *SecretKeySelector
```
We will need to handle nullability where invoked.

docs/EXTERNAL_MARIADB.md
mmontes11: Thank you very much for taking the time documenting this!

docs/EXTERNAL_MARIADB.md
```diff
@@ -0,0 +1,99 @@
...
+
+
+## Supported objects
+
+Currently the `ExternalMariaDB` is supported by the following objects:
+* Connection
+* User
+* Grant
+* Backup
+* SQLJob
```
mmontes11:
```suggestion
* SqlJob
```

internal/controller/backup_controller.go
```diff
@@ -95,9 +95,16 @@ func (r *BackupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+	mariadb interfaces.MariaDBGenericInterface) error {
 	return r.patch(ctx, backup, func(b *mariadbv1alpha1.Backup) {
-		backup.SetDefaults(mariadb)
+		if backup.Spec.MariaDBRef.Kind == mariadbv1alpha1.ExternalMariaDBKind {
+			mariadb_obj := mariadb.(*mariadbv1alpha1.ExternalMariaDB)
+			backup.SetExternalDefaults(mariadb_obj)
+		} else {
+			mariadb_obj := mariadb.(*mariadbv1alpha1.MariaDB)
+			backup.SetDefaults(mariadb_obj)
+		}
```
mmontes11:
```suggestion
		if externalMariaDB, ok :=  mariadb.(*mariadbv1alpha1.ExternalMariaDB); ok {
			backup.SetExternalDefaults(externalMariaDB)
		} else if mariadbObj, ok :=  mariadb.(*mariadbv1alpha1.MariaDB); ok  {
			backup.SetDefaults(mariadbObj)
		}
```

internal/controller/backup_controller_test.go
```diff
@@ -201,6 +201,94 @@ var _ = Describe("Backup", Label("basic"), func() {
 	)
 })
 
+var _ = Describe("Backup from external MariaDB", func() {
```
mmontes11, line 204: Thank you for taking the time testing this, it is crucial for making this sustainable for the maintenance perspective

internal/controller/connection_controller_test.go
```diff
@@ -436,3 +437,284 @@ var _ = Describe("Connection", Label("basic"), func() {
...
+				var err error
+				testCidrPrefix, err = docker.GetKindCidrPrefix()
+				Expect(err).NotTo(HaveOccurred())
+				Expect(testCidrPrefix).NotTo(BeEmpty())
+				return "test:MariaDB11!@tcp(" + testEmulateExternalMdbHost + ":3306)/test" +
+					"?timeout=5s&tls=mariadb-emdb-test-default-client-mdb-emulate-external-test-client-cert&parseTime=true"
+			}(),
+		),
+
+		// Entry(
```
mmontes11: I think this is something we want to test. Ensuring that we are able to establish TLS connections to external databases seems like a must have.

internal/controller/external_mariadb_controller.go
```diff
@@ -0,0 +1,253 @@
...
+		return ctrl.Result{RequeueAfter: 3 * time.Second}, fmt.Errorf("error connecting to MariaDB: %v", err)
+	}
+	defer client.Close()
+
+	raw_version, err := client.SystemVariable(ctx, "version")
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("unable to get MariaDB version: %v", err)
+	}
+
+	split_version := strings.Split(raw_version, "-")
```
mmontes11, line 189: I like the approach of inspecting the external `MariaDB` to figure out the version. However this seems prone to errors. Currently, it works because our container images have the version with this format `11.4.5-MariaDB-ubu2404-log`, but maybe other external databases i.e. RDS may not follow the same format, and we will end up pointing to a non existing image :
https://hub.docker.com/_/mariadb/tags

I suggest that we allow the user to optionally specify the version via a `spec.version` in the `ExternalMariaDB`. If nil, we can fallback to this approach, but we should return an error and patch the `ExternalMariaDB` status to clearly indicate this to the user.

What do you think?

internal/controller/external_mariadb_controller.go
```diff
@@ -0,0 +1,253 @@
...
+		Owns(&mariadbv1alpha1.Connection{}).
+		Owns(&corev1.ConfigMap{}).
+		Owns(&corev1.Secret{})
+
+	currentNamespaceOnly, err := env.CurrentNamespaceOnly()
+	if err != nil {
+		return fmt.Errorf("error checking operator watch scope: %v", err)
+	}
+	if !currentNamespaceOnly {
+		builder = builder.Owns(&rbacv1.ClusterRoleBinding{})
```
mmontes11: Unless I am missing something, I don't think this new controller reconciles `ClusterRoleBinding` objects. Could we omit this?

internal/controller/external_mariadb_controller.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	return builder.Complete(r)
+}
+
+func NewExternalMariaDBReconciler(client client.Client, refResolver *refresolver.RefResolver, conditionReady *condition.Ready,
+	builder *builder.Builder, scheme *runtime.Scheme,
+	sqlOpts ...sql.SqlOpt) *ExternalMariaDBReconciler {
```
mmontes11: I think we can omit `sqlOpts` ? doesn't seem to be used.

internal/controller/sqljob_controller.go
```diff
@@ -246,9 +247,13 @@ func (r *SqlJobReconciler) reconcileCronJob(ctx context.Context, sqlJob *mariadb
 }
 
 func (r *SqlJobReconciler) setDefaults(ctx context.Context, sqlJob *mariadbv1alpha1.SqlJob,
-	mariadb *mariadbv1alpha1.MariaDB) error {
+	mariadb interfaces.MariaDBGenericInterface) error {
 	return r.patch(ctx, sqlJob, func(s *mariadbv1alpha1.SqlJob) {
-		s.SetDefaults(mariadb)
+		if sqlJob.Spec.MariaDBRef.Kind != mariadbv1alpha1.ExternalMariaDBKind {
```
mmontes11: What about this?

```go
		if externalMariaDB, ok := mariadb.(*mariadbv1alpha1.ExternalMariaDB); ok {
			s.SetExternalDefaults(externalMariaDB)
		} else if mariaDBObj, ok := mariadb.(*mariadbv1alpha1.MariaDB); ok {
			s.SetDefaults(mariaDBObj)
		}
```

We may need to add a `SetExternalDefaults` method

pkg/builder/connection_builder.go
```diff
@@ -34,6 +35,16 @@ func (b *Builder) BuildConnection(opts ConnectionOpts, owner metav1.Object) (*ma
 			Database:             opts.Database,
 		},
 	}
+
+	if opts.ExternalMariaDB != nil {
+		conn.Spec.MariaDBRef = &mariadbv1alpha1.MariaDBRef{
+			ObjectReference: mariadbv1alpha1.ObjectReference{
+				Name: opts.ExternalMariaDB.Name,
+			},
+			Kind:      "ExternalMariaDB",
```
mmontes11: Could we create a constant for this along with the other kinds?
https://github.com/mariadb-operator/mariadb-operator/blob/3da8aa4d9cd306026293cb3a353c89b3b17c1de0/api/v1alpha1/groupversion_info.go#L15

pkg/builder/connection_builder.go
```diff
@@ -49,6 +60,15 @@ func (b *Builder) BuildConnection(opts ConnectionOpts, owner metav1.Object) (*ma
 	if opts.Template != nil {
 		conn.Spec.ConnectionTemplate = *opts.Template
 	}
+
+	if b == nil {
```
mmontes11: In which case gets `b` to be nil, if using the `NewBuilder`?
https://github.com/mariadb-operator/mariadb-operator/blob/3da8aa4d9cd306026293cb3a353c89b3b17c1de0/pkg/builder/builder.go#L15

pkg/command/backup.go
```diff
@@ -34,6 +35,11 @@ type BackupOpts struct {
 	ExtraOpts            []string
 }
 
+type ConnectionParamsAwareInterface interface {
```
mmontes11: Isn't this defined under `pkg/interfaces` already?

pkg/command/command.go
```diff
@@ -46,21 +46,26 @@ func ConnectionFlags(co *CommandOpts, mariadb *mariadbv1alpha1.MariaDB) (string,
...
-func host(mariadb *mariadbv1alpha1.MariaDB) string {
-	if mariadb.IsHAEnabled() {
-		return statefulset.ServiceFQDNWithService(
-			mariadb.ObjectMeta,
-			mariadb.PrimaryServiceKey().Name,
-		)
-	}
-	return statefulset.ServiceFQDN(mariadb.ObjectMeta)
-}
+// func host(mariadb interfaces.MariaDBGenericInterface) string {
```
mmontes11: I think this is equivalent to `GetHost`, probably we can remove

pkg/interfaces/interfaces.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+type TLSAwareInterface interface {
+	IsTLSEnabled() bool
+	TLSCABundleSecretKeyRef() mariadbv1alpha1.SecretKeySelector
+	TLSClientCertSecretKey() types.NamespacedName
+	TLSServerCertSecretKey() types.NamespacedName
+}
+
+type ReplicationAwareInterface interface {
```
mmontes11, line 24: I think this interface can be removed, please see my previous comments

mmontes11 (changes requested):
Thank you very much for taking the initiative, I do appreciate the effort made here 🙏🏻 

Whilst I believe you are on the right track, I think some adaptations are needed, specially around the new abstractions that we have created. Please see my comments, let me know what you think.

wfelipew:
> Thank you very much for taking the initiative, I do appreciate the effort made here 🙏🏻
> 
> Whilst I believe you are on the right track, I think some adaptations are needed, specially around the new abstractions that we have created. Please see my comments, let me know what you think.

Thanks for the review, I'll working on it today

pkg/controller/sql/controller.go
wfelipew: @mmontes11 The problem here is the caller function, if we revert the signature we'll need put an IF condition and cast it MariaDB object before calling the waitForMariaDB: 
```
func (r *SqlReconciler) Reconcile(ctx context.Context, resource Resource) (ctrl.Result, error) {
	if resource.IsBeingDeleted() {
		if result, err := r.Finalizer.Finalize(ctx, resource); !result.IsZero() || err != nil {
			return result, err
		}
		return ctrl.Result{}, nil
	}

	mariadb, err := r.RefResolver.GenericMariaDB(ctx, resource.MariaDBRef(), resource.GetNamespace())
	if err != nil {
		var errBundle *multierror.Error
		errBundle = multierror.Append(errBundle, err)

		err = r.WrappedReconciler.PatchStatus(ctx, r.ConditionReady.PatcherRefResolver(err, mariadb))
		errBundle = multierror.Append(errBundle, err)

		return ctrl.Result{}, fmt.Errorf("error getting MariaDB: %v", errBundle)
	}
	
	if result, err := waitForMariaDB(ctx, r.Client, mariadb, r.LogSql); !result.IsZero() || err != nil {
		var errBundle *multierror.Error

		if err != nil {
			errBundle = multierror.Append(errBundle, err)

			err := r.WrappedReconciler.PatchStatus(ctx, r.ConditionReady.PatcherWithError(err))
			errBundle = multierror.Append(errBundle, err)
		}

		return result, errBundle.ErrorOrNil()
	}
```
That will be require something like this: 
```
func (r *SqlReconciler) Reconcile(ctx context.Context, resource Resource) (ctrl.Result, error) {
	if resource.IsBeingDeleted() {
		if result, err := r.Finalizer.Finalize(ctx, resource); !result.IsZero() || err != nil {
			return result, err
		}
		return ctrl.Result{}, nil
	}

	mariadb, err := r.RefResolver.GenericMariaDB(ctx, resource.MariaDBRef(), resource.GetNamespace())
	if err != nil {
		var errBundle *multierror.Error
		errBundle = multierror.Append(errBundle, err)

		err = r.WrappedReconciler.PatchStatus(ctx, r.ConditionReady.PatcherRefResolver(err, mariadb))
		errBundle = multierror.Append(errBundle, err)

		return ctrl.Result{}, fmt.Errorf("error getting MariaDB: %v", errBundle)
	}

	if mariadb.GetObjectKind().GroupVersionKind().Kind != mariadbv1alpha1.ExternalMariaDBKind {
		mdb := mariadb.(*mariadbv1alpha1.MariaDB)
		if result, err := waitForMariaDB(ctx, r.Client, mdb, r.LogSql); !result.IsZero() || err != nil {
			var errBundle *multierror.Error

			if err != nil {
				errBundle = multierror.Append(errBundle, err)

				err := r.WrappedReconciler.PatchStatus(ctx, r.ConditionReady.PatcherWithError(err))
				errBundle = multierror.Append(errBundle, err)
			}

			return result, errBundle.ErrorOrNil()
		}
	}
```
I think the implementation with the interface will keep the code more clean and less complex. What do you think? should we keep the interface?

wfelipew: @mmontes11 I believe I fixed everything, except by this one:
> I think we can leave the signature of this function as it is, with `*mariadbv1alpha1.MariaDB`:
> 
> ```go
> func waitForMariaDB(ctx context.Context, client ctrlclient.Client, mdb *mariadbv1alpha1.MariaDB,
> 	logSql bool) (ctrl.Result, error) {
> 
> 	if mdb.GetObjectKind().GroupVersionKind().Kind == mariadbv1alpha1.ExternalMariaDBKind {
> 		return ctrl.Result{}, nil
> 	}
> 
> 	healthy, err := health.IsStatefulSetHealthy(
> 		ctx,
> 		client,
> 		ctrlclient.ObjectKeyFromObject(mdb),
> 		health.WithDesiredReplicas(mdb.Spec.Replicas),
> 		health.WithPort(mdb.Spec.Port),
> 		health.WithEndpointPolicy(health.EndpointPolicyAll),
> 	)
> 	if err != nil {
> 		return ctrl.Result{}, err
> 	}
> 	if !healthy {
> 		if logSql {
> 			log.FromContext(ctx).V(1).Info("MariaDB unhealthy. Requeuing SQL resource")
> 		}
> 		return ctrl.Result{RequeueAfter: 1 * time.Second}, nil
> 	}
> 	return ctrl.Result{}, nil
> }
> ```
> 
> `MariaDB` has already the `GetObjectKind()` method, so you could return early when it is a `ExternalMariaDBKind`. This implies that we can remove `GetReplicas` from the `ReplicationAwareInterface`, as this is the only place where it is used.

wfelipew:
@mmontes11 I believe I fixed everything except by this one -> https://github.com/mariadb-operator/mariadb-operator/pull/1311#discussion_r2213904586

internal/controller/external_mariadb_controller.go
wfelipew: @mmontes11 added this logic 
```
func (m *ExternalMariaDB) GetImage() string {
	var version string
	if m.Spec.Version != nil {
		version = *m.Spec.Version
	} else {
		version = m.Status.Version
	}
	return fmt.Sprintf("mariadb:%s", version) // ExternalMariaDB uses official MariaDB images (publicly available) for the Backups
}
```

It will use the version from spec if defined, otherwise it will use the version from status. What do you think?

mmontes11: Thanks! That way the user has more control on which image to use. I think I will merge this pretty soon to a branch and make some minor adjustments, like allowing the user to specify the complete image path, so they can point to their private registry. This would be useful for users running in air-gapped environments, with a local registry.

pkg/controller/sql/controller.go
mmontes11: Yes, let's keep the interface. Keeping an additional method  `GetReplicas` in the interface will be easier to maintain than the conditional logic.

pkg/interfaces/interfaces.go
mmontes11: Not really, see my last comment

PR #1313, opened 2025-07-03
Add --sync-period opt to manager

hedgieinsocks:
When having 10k+ db/user/grant resources, and requeueSql set to 0 I still notice periodic bulk reconciles every 10h that create 10k queue for ~7 minutes. Of course, one could raise maxConcurrentReconciles to lower the spike duration, but having an option to disable or tune informer sync would be nicer. 

The solution is taken from: https://github.com/kubernetes-sigs/controller-runtime/issues/521

By defaulting it to 10h https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.20.4/pkg/cache#Options we do not modify the default behavior for regular not hardcore users.

> Defaults to 10 hours if unset.

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -79,6 +79,8 @@ var (
 
 	requeueSqlMaxOffset time.Duration
 
+	syncPeriod time.Duration
+
 	webhookEnabled bool
 	webhookPort    int
 	webhookCertDir string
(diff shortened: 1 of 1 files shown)


mmontes11 (approved):
Happy to support this. Pretty useful when running at scale, like your case.

PR #1314, opened 2025-07-03
Allow tweaking kube api client ratelimiter

hedgieinsocks:
The latest controller-runtime decided to disable client-side ratelimit: https://github.com/kubernetes-sigs/controller-runtime/commit/ab40409635dc0c91781f575a87c64a579680be95

That is opinionated, so we can allow the end user to decide what is best. Let's use the values from the controller-runtime we are using https://github.com/kubernetes-sigs/controller-runtime/blob/v0.20.4/pkg/client/config/config.go#L101-L106

And if/when we upgrade, we can just change the defaults if need be.

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -68,6 +68,9 @@ var (
 	logMaxScale    bool
 	logSql         bool
 
+	kubeApiQps   float32
+	kubeApiBurst int
+
 	maxConcurrentReconciles         int
 	mariadbMaxConcurrentReconciles  int
 	maxscaleMaxConcurrentReconciles int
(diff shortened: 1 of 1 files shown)


mmontes11 (changes requested):
Thanks for bringin this up! 

Although the Slack threads from this commit https://github.com/kubernetes-sigs/controller-runtime/commit/ab40409635dc0c91781f575a87c64a579680be95 have been insightful, I still see the value of having client-side rate limit.

A good argument is the one that @aojea brought up:
> it can have the side effect of ddosing the apiserver too … don’t understimate thundering herds of controllers requests

> APF solves the problem in the server, but the traffic still reaches the server

Therefore, although I'm open to revisit this in the future, I think the best way to go would be adding a flag and default it to 5 QPS (current default), enabling you to set it to -1 by explicitly passing the flag .

What do you think?

hedgieinsocks:
Oh, you actually followed the links, kudos! I didn't go further than the commit itself. But I agree that flags are the best compromise. As for the defaults, I set the ones from the current controller version [kubernetes-sigs/controller-runtime@v0.20.4/pkg/client/config/config.go#L101-L106](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.20.4/pkg/client/config/config.go?rgh-link-date=2025-07-03T10%3A40%3A33Z#L101-L106) 

My linter shows that 5 is the default QPS https://pkg.go.dev/k8s.io/client-go@v0.32.3/rest#Config.QPS but the code above says otherwise, so the linter is just confused?

PR #1320, opened 2025-07-11
Bump requeue interval defaults

mmontes11:
Current requeue defaults are very aggresive, causing a higher memory, cpu and network usage in the controller. This PR bumps this defaults to mitigate this. Requeue intervals may be overriden at the resource level.

Related to:
- https://github.com/mariadb-operator/mariadb-operator/pull/1244


mmontes11:
FYI @hedgieinsocks

Issue #1318, opened 2025-07-08
[Feature] Fix golangci-lint static checks


mmontes11:
https://github.com/mariadb-operator/mariadb-operator/blob/bb961528a4a3be0afe6cd90d78b4370214deb276/.golangci.yml#L27

kashitaka:
@mmontes11 Hi, I’ve opened a PR to start addressing issue here: https://github.com/mariadb-operator/mariadb-operator/pull/1321

For now, I’ve only fixed `ST1001` as it was simple to tackle.  
I’d like to confirm if this direction is acceptable before moving on to the other checks.

In particular:

- **ST1003** would involve renaming identifiers, which could make breaking changes.  
  - For example: `type SqlFinalizer` → `type SQLFinalizer`.
- **ST1005** would slightly change error message formats.  

If those changes are okay, I’m happy to work on them as well.  
Looking forward to your feedback!

mmontes11:
> ST1003 would involve renaming identifiers, which could make breaking changes. For example: type SqlFinalizer → type SQLFinalizer.

Not a problem, as long as we identify the BREAKING CHANGE in the PR, so anyone can migrate seamlessly by looking at the PR description. For example: https://github.com/mariadb-operator/mariadb-operator/pull/1324


> ST1005 would slightly change error message formats.

That's acceptable, thanks for the heads up.

kashitaka:
Since ST1003 involves breaking changes, and #1321 already includes a number of modifications, I think it would be better to handle ST1003 in a separate PR.
If you'd like me to work on that, just let me know! I'm happy to open another PR for it if needed.

mmontes11:
> Since ST1003 involves breaking changes, and https://github.com/mariadb-operator/mariadb-operator/pull/1321 already includes a number of modifications, I think it would be better to handle ST1003 in a separate PR.
If you'd like me to work on that, just let me know! I'm happy to open another PR for it if needed.

Agreed, `ST1003` should go in a different PR, with the breaking changes identified as we did in https://github.com/mariadb-operator/mariadb-operator/pull/1324

kashitaka:
@mmontes11 I've fixed all the remaining staticcheck issues and made sure the CI passes. https://github.com/mariadb-operator/mariadb-operator/pull/1334
Since the diff is quite large, please take a look when you have time. Thanks!

PR #1321, opened 2025-07-11
fix: resolve staticcheck issues (ST1001, QF1008, ST1005)

kashitaka:
This is part of the fix for issue #1318.

I’ve addressed only ST1001 (“dot imports are discouraged”) for now, as it was a straightforward change.  
Since fixing all staticcheck issues would involve changes across many files, I'd prefer to proceed step by step and confirm the approach with maintainers.

Please let me know if this direction looks good.

### DIFF
--- a/internal/controller/physicalbackup_controller_job.go
+++ b/internal/controller/physicalbackup_controller_job.go
@@ -341,7 +341,7 @@ func (r *PhysicalBackupReconciler) deleteJobSync(ctx context.Context, job *batch
 		if err := r.Get(ctx, key, &j); apierrors.IsNotFound(err) {
 			return nil
 		}
-		return errors.New("Job still exists")
+		return errors.New("Job still exists") //nolint:staticcheck
 	})
 }
 
--- a/internal/controller/restore_controller.go
+++ b/internal/controller/restore_controller.go
@@ -112,10 +112,10 @@ func (r *RestoreReconciler) setDefaults(ctx context.Context, restore *mariadbv1a
 		return fmt.Errorf("error patching restore: %v", err)
 	}
 
-	if restore.Spec.RestoreSource.IsDefaulted() {
+	if restore.Spec.IsDefaulted() {
 		return nil
 	}
-	if restore.Spec.RestoreSource.BackupRef == nil {
+	if restore.Spec.BackupRef == nil {
 		var restoreErr *multierror.Error
 		restoreErr = multierror.Append(restoreErr, errors.New("unable to determine restore source, 'backupRef' is nil"))
 
@@ -125,7 +125,7 @@ func (r *RestoreReconciler) setDefaults(ctx context.Context, restore *mariadbv1a
 		return restoreErr
 	}
 
-	backup, err := r.RefResolver.Backup(ctx, restore.Spec.RestoreSource.BackupRef, restore.Namespace)
+	backup, err := r.RefResolver.Backup(ctx, restore.Spec.BackupRef, restore.Namespace)
 	if err != nil {
 		var backupErr *multierror.Error
 		backupErr = multierror.Append(backupErr, err)
@@ -147,7 +147,7 @@ func (r *RestoreReconciler) setDefaults(ctx context.Context, restore *mariadbv1a
 	}
 
 	if err := r.patch(ctx, restore, func(r *mariadbv1alpha1.Restore) error {
-		return r.Spec.RestoreSource.SetDefaultsWithBackup(backup)
+		return r.Spec.SetDefaultsWithBackup(backup)
 	}); err != nil {
 		return fmt.Errorf("error patching restore: %v", err)
 	}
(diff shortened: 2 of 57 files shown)


test/utils/utils.go
```diff
@@ -24,7 +24,7 @@ import (
 	"os/exec"
 	"strings"
 
-	. "github.com/onsi/ginkgo/v2" //nolint:golint,revive
+	"github.com/onsi/ginkgo/v2/dsl/core"
```
mmontes11: It is an unwritten rule but, we normally import the core Kubernetes APIs as `corev1`, importing ginkgo as `core` will cause some confusion. Maybe you can alias it as `ginkgocore`? I leave it up to you

mmontes11 (changes requested):
The approach looks legit to me. Left a naming suggestion to avoid clashing with our current imports.

Thanks for willing to contribute!

test/utils/utils.go
kashitaka: Thanks for the review! I’ve applied the suggested change here: https://github.com/mariadb-operator/mariadb-operator/pull/1321/commits/732f097f902f5f66b225b758c44f2088e738dc1d
Please take another look when you have time.

kashitaka:
> The approach looks legit to me. Left a naming suggestion to avoid clashing with our current imports.
> Thanks for willing to contribute!

Thank you for your feedback. I'll work on other typecheck issues except naming rule(`ST1003`) then.

test/utils/utils.go
mmontes11: LGTM, you may proceed with the rest of dot imports.

kashitaka:
I Fixed two more staticcheck issues. 

- QF1008: https://github.com/mariadb-operator/mariadb-operator/pull/1321/commits/43ccc251c797f90710daa668e2430f88ba7256c7
- ST1005: https://github.com/mariadb-operator/mariadb-operator/pull/1321/commits/7b8245c35717c7d6775ce55f058c95a8efee9b7e

Fixing QF1008 just slightly affects on readability, but I think it's small changes no problem on the software behavior.

However, fixing ST1005 changes error messages that users may see, and it can also introduce some inconsistencies (for example, `Galera` becomes lowercase when it appears at the beginning of a message).

Please take a look at the ST1005-related commit and let me know your preference. If you prefer, we can also add `//nolint:staticcheck` comments on specific lines where we want to ignore the warning.

mmontes11 (changes requested):
> QF1008

LGTM!

> ST1005

Could we ignore this rule in the following cases?
- When referring to an API type i.e. `Pod` or `Backup`
- When referring to names i.e. `Galera` 

Thank you! You are on the right track.

internal/controller/physicalbackup_controller_job.go
```diff
@@ -341,7 +341,7 @@ func (r *PhysicalBackupReconciler) deleteJobSync(ctx context.Context, job *batch
 		if err := r.Get(ctx, key, &j); apierrors.IsNotFound(err) {
 			return nil
 		}
-		return errors.New("Job still exists")
+		return errors.New("Job still exists") //nolint:staticcheck
```
kashitaka, line 344: add `nolint` as Job also seems API

pkg/kubernetes/auth/auth.go
```diff
@@ -78,7 +78,7 @@ func (a *KubernetesAuth) Handler(next http.Handler) http.Handler {
 func authToken(r *http.Request) (string, error) {
 	auth := r.Header.Get("Authorization")
 	if auth == "" {
-		return "", errors.New("Authorization header not found")
+		return "", errors.New("Authorization header not found") //nolint:staticcheck
```
kashitaka, line 81: add `nolint` here as `Authorization` is a parameter name

internal/controller/restore_controller.go
```diff
@@ -138,7 +138,7 @@ func (r *RestoreReconciler) setDefaults(ctx context.Context, restore *mariadbv1a
 
 	if !backup.IsComplete() {
 		var errBundle *multierror.Error
-		errBundle = multierror.Append(errBundle, errors.New("Backup not complete"))
+		errBundle = multierror.Append(errBundle, errors.New("Backup not complete")) //nolint:staticcheck
```
kashitaka, line 141: also `Backup` looks API so add `nolint`

kashitaka:
Thank you for the review! I’ve addressed the comments.
To make it easier to review, I’ve squashed all the ST1005-related changes into a single commit. You can check it here: https://github.com/mariadb-operator/mariadb-operator/pull/1321/commits/e691502659fa68835899619b58738f94997aa34a

Let me know if anything else needs adjustment!

mmontes11:
@kashitaka would you mind rebasing with main?

kashitaka:
@mmontes11 Thanks for updating the commits! Let me know if there's anything else you'd like me to do.

mmontes11 (approved):
LGTM! Thank you very much for this contribution! 🦭 

Will be released in [25.08](https://github.com/mariadb-operator/mariadb-operator/pull/1306)

Issue #1176, opened 2025-02-24
[Bug] Revoking privileges on a user with specific database and GRANT OPTION can produce invalid SQL statement


nicholaskuechler:
**Documentation**
- [x] I acknowledge that I have read the relevant [documentation](https://github.com/mariadb-operator/mariadb-operator/tree/main/docs).

**Describe the bug**

Revoking grants can produce an invalid SQL statement when attempting to revoke a user with both a specific database assignment and the user also has `GRANT OPTION`.

The SQL produced by revoke has a subtle problem which can be described in an old mysql bug report I found: https://bugs.mysql.com/bug.php?id=69313

Basically, you can't issue a:
```REVOKE ALL PRIVILEGES, GRANT OPTION ON `nova_cell0`.* FROM 'nova'@'%';``` as it produces an error:

```
MariaDB [(none)]> REVOKE ALL PRIVILEGES, GRANT OPTION ON `nova_cell0`.* FROM 'nova'@'%';
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'ON `nova_cell0`.* FROM 'nova'@'%'' at line 1
```

You can only do `REVOKE all privileges, grant option FROM user` - it can't be combined with the specific database nova_cell0 with ```ON `nova_cell0`.*```

Maybe there should be two separate REVOKEs - 1 for revoking on a specific database, and a second revoke for the grant options?

MariaDB revoke reference docs showing the `REVOKE ... ON` versus `REVOKE GRANT OPTION FROM`: https://mariadb.com/kb/en/revoke/

**Expected behaviour**

Expecting revoke to work without error.

**Steps to reproduce the bug**

Using manifest:

```
---
apiVersion: k8s.mariadb.com/v1alpha1
kind: Grant
metadata:
  name: nova-cell0-grant
  namespace: openstack
spec:
  mariaDbRef:
    name: mariadb
    waitForIt: true
  privileges:
    - "ALL"
  database: "nova_cell0"
  table: "*"
  username: nova
  grantOption: true
  host: "%"
  retryInterval: 5s
```

Then issuing a revoke. We added logging to see the actual SQL statement being executed, which is invalid:

```
REVOKE ALL,GRANT OPTION ON `nova_cell0`.* FROM 'nova'@'%';
```

```
[...]

PR #1326, opened 2025-07-13
Make grant privileges mutable

hedgieinsocks:
Since MariaDB does not yet support https://jira.mariadb.org/browse/MDEV-17476 we need another way to limit disk usage abusers, e.g. to edit grant permissions on the fly to allow only selecting until the abuser agrees to take action. 

However, since privileges are immutable, one will have to delete the grant resource. And if the grant resource is controlled by a custom operator, then uh-oh bad luck.

Luckily, it seems we can easily track last applied privileges in the status, and revoke the ones that are no longer present in the spec declaration.

---

I revoke only those privileges that are no longer needed. But we can make it simpler if need be and blindly remove the whole set, the ones we still need will be reapplied anyway the next moment.

Or we can refactor revoke sql completely and like mentioned in https://github.com/mariadb-operator/mariadb-operator/issues/1176 just run `REVOKE ALL PRIVILEGES, GRANT OPTION FROM user;` to nuke it.


internal/controller/grant_controller.go
```diff
@@ -104,6 +123,7 @@ func (wr *wrappedGrantReconciler) Reconcile(ctx context.Context, mdbClient *sqlC
 func (wr *wrappedGrantReconciler) PatchStatus(ctx context.Context, patcher condition.Patcher) error {
 	patch := client.MergeFrom(wr.grant.DeepCopy())
 	patcher(&wr.grant.Status)
+	wr.grant.Status.ActivePrivileges = wr.grant.Spec.Privileges
```
mmontes11: Updating `ActivePrivileges` should only be done after succesfully executing the grant statement. `PatchStatus` is invoked in multiple places where that is not the case, for example:
https://github.com/mariadb-operator/mariadb-operator/blob/main/pkg/controller/sql/controller.go#L90

Therefore it would be wrongly reporting that we have successfully applied the privileges.

Could we invoke `PatchStatus` in this line to patch `ActivePriviliges` here?
https://github.com/mariadb-operator/mariadb-operator/blob/main/internal/controller/grant_controller.go#L101

mmontes11 (changes requested):
Thanks for raising this @hedgieinsocks ! Useful feature that users have mentioned long time ago. 

I think we should track the existing privileges in the status and revoke them more granally, as you suggested. Left a couple of observations in the implementation.

Related to https://github.com/mariadb-operator/mariadb-operator/issues/1176.

internal/controller/grant_controller.go
hedgieinsocks: Oh, indeed. But it looks like `PatchStatus()` would require refactoring to become something like `patchStatus()` in some other places. Or we could just inject this instead of the function:
```go
	if wr.grant.Status.ActivePrivileges != nil && !slices.Equal(wr.grant.Spec.Privileges, wr.grant.Status.ActivePrivileges) {
		wr.grant.Status.ActivePrivileges = wr.grant.Spec.Privileges
		patch := client.MergeFrom(wr.grant.DeepCopy())
		if err := wr.Client.Status().Patch(ctx, wr.grant, patch); err != nil {
			return fmt.Errorf("error patching Grant status: %v", err)
		}
	}
```
But it kinda looks hackish given that just a few lines below we have a function that does almost the same. Do you see a way to refactor `PatchStatus()` ?

hedgieinsocks: In the meantime, tried to resolve the issue via https://github.com/mariadb-operator/mariadb-operator/pull/1326/files#diff-fa4ffed4822246bd8722bd8efccf8c83e34b0752473af6f68078cd3a130033ddR132-R134

mmontes11: How about this?

```go
func (wr *wrappedGrantReconciler) Reconcile(ctx context.Context, mdbClient *sqlClient.Client) error {
	var opts []sqlClient.GrantOption

	if wr.grant.Status.ActivePrivileges != nil && !slices.Equal(wr.grant.Spec.Privileges, wr.grant.Status.ActivePrivileges) {
		var revokePrivileges []string
		for _, p := range wr.grant.Status.ActivePrivileges {
			if !slices.Contains(wr.grant.Spec.Privileges, p) {
				revokePrivileges = append(revokePrivileges, p)
			}
		}
		if len(revokePrivileges) > 0 {
			if err := mdbClient.Revoke(
				ctx,
				revokePrivileges,
				wr.grant.Spec.Database,
				wr.grant.Spec.Table,
				wr.grant.AccountName(),
				opts...,
			); err != nil {
				return fmt.Errorf("error revoking grant in MariaDB: %v", err)
			}
		}
	}

	if wr.grant.Spec.GrantOption {
		opts = append(opts, sqlClient.WithGrantOption())
	}
	if err := mdbClient.Grant(
		ctx,
		wr.grant.Spec.Privileges,
		wr.grant.Spec.Database,
		wr.grant.Spec.Table,
		wr.grant.AccountName(),
		opts...,
	); err != nil {
		return fmt.Errorf("error granting privileges in MariaDB: %v", err)
	}
	if err := wr.patchStatusWithFunc(ctx, func(status *mariadbv1alpha1.GrantStatus) {
		wr.grant.Status.ActivePrivileges = wr.grant.Spec.Privileges
	}); err != nil {
		return fmt.Errorf("error patching active privileges: %v", err)
	}
	return nil
}

func (wr *wrappedGrantReconciler) PatchStatus(ctx context.Context, patcher condition.Patcher) error {
	return wr.patchStatusWithFunc(ctx, func(status *mariadbv1alpha1.GrantStatus) {
		patcher(status)
	})
}

func (wr *wrappedGrantReconciler) patchStatusWithFunc(ctx context.Context, patchFn func(status *mariadbv1alpha1.GrantStatus)) error {
	patch := client.MergeFrom(wr.grant.DeepCopy())
	patchFn(&wr.grant.Status)

	if err := wr.Client.Status().Patch(ctx, wr.grant, patch); err != nil {
		return fmt.Errorf("error patching Grant status: %v", err)
	}
	return nil
}
```

hedgieinsocks: oh, thank you for this nice solution!

api/v1alpha1/grant_types.go
```diff
@@ -54,6 +54,11 @@ type GrantStatus struct {
 	// +optional
 	// +operator-sdk:csv:customresourcedefinitions:type=status,xDescriptors={"urn:alm:descriptor:io.kubernetes.conditions"}
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
+	// ActivePrivileges is the list of last applied privileges used in the Grant.
+	// It allows tracking the difference between the declared and applied privileges and thus facilitates their mutability.
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=status
+	ActivePrivileges []string `json:"activePrivileges"`
```
mmontes11: To better align with the Kubernetes Operator pattern naming conventions (desired state vs current state), would you mind renaming this to `CurrentPrivileges` (desired state vs current state)?

mmontes11 (changes requested):
LGTM! Works like a charm, congrats!

I've left my nitpicky suggestion as usual. Happy to merge after addressing it.

api/v1alpha1/grant_types.go
hedgieinsocks: sure, I've additionally exported diff logic to `privilegesToRevoke()` to clean it up a bit

internal/controller/grant_controller.go
```diff
@@ -98,19 +115,44 @@ func (wr *wrappedGrantReconciler) Reconcile(ctx context.Context, mdbClient *sqlC
...
 
 	if err := wr.Client.Status().Patch(ctx, wr.grant, patch); err != nil {
 		return fmt.Errorf("error patching Grant status: %v", err)
 	}
 	return nil
 }
 
+func (wr *wrappedGrantReconciler) privilegesToRevoke() []string {
+	var revokePrivileges []string
+	if !slices.Equal(wr.grant.Status.CurrentPrivileges, wr.grant.Spec.Privileges) {
```
mmontes11: I think you mean:
```suggestion
	if slices.Equal(wr.grant.Status.CurrentPrivileges, wr.grant.Spec.Privileges) {
```
Right? Return early if no diff in the privileges i.e. noop. Otherwise, perform the diff and revoke.

api/v1alpha1/grant_types.go
mmontes11: Way cleaner, thank you! 🙏🏻

mmontes11 (commented):
Please take a look at my comment

internal/controller/grant_controller.go
hedgieinsocks: ah, indeed, forgot to revert it, thanks for spotting it!

mmontes11 (approved):
LGTM! Thank you very much for this initiative! 🙏🏻

PR #1275, opened 2025-05-11
Add mariadb-cluster helm chart

hedgieinsocks:
The initial version of mariadb-cluster helm chart will be able to template:

 * 1 MariaDB
 * list of Database
 * list of User
 * list of Grant
 * list of Backup

Other resources might be included via prs in the future.

closes: https://github.com/mariadb-operator/mariadb-operator/issues/1273

PR #1335, opened 2025-07-16
`mariadb-cluster` helm chart improvements

mmontes11:
Related to https://github.com/mariadb-operator/mariadb-operator/pull/1275


hedgieinsocks:
Am I right that .name for backups and grants will be just ignored since this key is not in the CRD spec? If yes, then we are good. Otherwise, we might need to unset that key like we unset mdbref

mmontes11:
Good point @hedgieinsocks ! Thanks for raising this. I was keen on achieving consistency in all CRs values, but this is def something to look at.

I've tested with both kubectl and helm. The former one spits a validation error but the latter just returns a warning:
```bash
❯ helm upgrade --install mariadb-cluster deploy/charts/mariadb-cluster
W0716 20:01:38.862721  671533 warnings.go:70] unknown field "spec.name"
``` 

I think I will create a template to omit both `mariaDbRef` and `name`. In the future, when suporting `MaxScale`, we will need to omit `maxScaleRef` as well.

What do you think?

hedgieinsocks:
Thank you for the quick test, your plan sounds good!

mmontes11:
Hey @hedgieinsocks ! Here it is how it looks like: https://github.com/mariadb-operator/mariadb-operator/pull/1335/commits/64840e08ffbc71fd1eec5ff21d040b57faef0351

Let me know if you see any issues

hedgieinsocks:
I think you accidentally removed .name from resources where it is a valid key like database and user crds. Otherwise, lgtm!

mmontes11:
Good catch, fixed: https://github.com/mariadb-operator/mariadb-operator/pull/1335/commits/2eb5aad66bb188098e10a4514229c21dd87d28b0

deploy/charts/mariadb-cluster/README.md
```diff
@@ -0,0 +1,40 @@
...
+```bash
+helm repo add mariadb-operator https://helm.mariadb.com/mariadb-operator
+helm install mariadb-operator-crds mariadb-operator/mariadb-operator-crds
+helm install mariadb-operator mariadb-operator/mariadb-operator
+```
+
+Once the operator is deployed, you can provision a MariaDB cluster using the `mariadb-cluster` Helm chart:
+
+```bash
+helm install mariadb-operator mariadb-operator/mariadb-cluster
```
hedgieinsocks: wrong release name

mmontes11: Fixed https://github.com/mariadb-operator/mariadb-operator/pull/1335/commits/d38aec1966c2cef2ebb0115c37577e90b4548915

hedgieinsocks:
@mmontes11 extra fourth ` on https://github.com/mariadb-operator/mariadb-operator/blob/main/deploy/charts/mariadb-cluster/README.md.gotmpl#L24 :smiley:

mmontes11:
Fixed: https://github.com/mariadb-operator/mariadb-operator/commit/29311b36d9d12f059f9a51a933d3d33cd99334ef

@hedgieinsocks

PR #1337, opened 2025-07-18
[WIP] Add PrometheusRule to mariadb-cluster

hedgieinsocks:
I believe it will be convenient to have several basic fully customizable cluster-scoped prometheus alerts in `mariadb-cluster` with an easy way to add custom ones. 

**TO DO:**
- [x] create templating for bundled rules with optional overrides
- [x] create templating for custom rules
- [ ] find good useful alerts (generic, replication and galera)
- [ ] add tests

### DIFF
--- a/deploy/charts/mariadb-cluster/prometheus/galera.yaml
+++ b/deploy/charts/mariadb-cluster/prometheus/galera.yaml
@@ -1,2 +1,36 @@
 # The rules are helm templated, so we must escape labels:
 # {{ $labels.target }} => {{"{{ $labels.target }}"}}
+- alert: MariaDBGaleraNotReady
+  expr: |
+    mysql_global_status_wsrep_ready{target=~"{{ include "mariadb-cluster.fullname" . }}.*"} != 1
+  for: 1m
+  labels:
+    severity: critical
+  annotations:
+    summary: MariaDB instance in a Galera cluster is not ready
+    description: |
+      MariaDB instance {{"`{{ $labels.target }}`"}} in a Galera cluster is not ready
+- alert: MariaDBGaleraOutOfSync
+  expr: |
+    mysql_global_status_wsrep_local_state{target=~"{{ include "mariadb-cluster.fullname" . }}.*"} != 4
+    and
+    mysql_global_variables_wsrep_desync{target=~"{{ include "mariadb-cluster.fullname" . }}.*"} == 0
+  for: 1m
+  labels:
+    severity: critical
+  annotations:
+    summary: MariaDB instance in a Galera cluster is out of sync
+    description: |
+      MariaDB instance {{"`{{ $labels.target }}`"}} in a Galera cluster in out of sync
+- alert: MariaDBGaleraDonorFallingBehind
+  expr: |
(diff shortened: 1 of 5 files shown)


mmontes11:
Agreed with this feature. Optionally adding some `PrometheusRules` to the `mariadb-cluster` would be great. This is something other projects are doing on their helm charts intended to deploy CR, for example the cluster helm chart from `rook-ceph`:
- https://github.com/rook/rook/blob/d720f2ae6c6ccac7b6b0d07e23e270702ad7fee2/deploy/charts/rook-ceph-cluster/templates/prometheusrules.yaml 

The first thing would be doing some research and gather a list of commonly used alerts, I don't have any well-known source for this at this time, but I am open to discuss.

hedgieinsocks:
The linked template happens to be updated by me :D 
It allows keeping the rules in a separate source file and overriding them if necessary.
The con is that you cannot add your custom rules there. But that could be tackled by a new map for custom rules.
And we can choose the right source file depending on galera, replication or single-master mode.
So if you like that file sourcing approach, I can pursue it.

As for the alerts themselves, indeed we would need to research them, e.g. https://samber.github.io/awesome-prometheus-alerts/rules#mysql or something proved useful in prod environments.

mmontes11:
> The linked template happens to be updated by me :D

Oh indeed, great impact on cloud native landscape! 🚀 

I think it is a legit approach that we could follow. I wouldn't worry too much about custom rules, anyone could define a `PrometheusRule` resource outside of the helm chart, if really needed. We are more keen on a out of the box experience that works more than a tailored solution.

> https://samber.github.io/awesome-prometheus-alerts/rules#mysql

I think we could reuse some of these rules, as you said, conditionally enabling them based on the topology chosen by the user.

Great initiative!

hedgieinsocks:
galera rules suggested by exporter: https://github.com/prometheus/mysqld_exporter/blob/main/mysqld-mixin%2Falerts%2Fgalera.yaml

Nils98Ar:
@mmontes11 @hedgieinsocks

Ready to merge, or are further changes required? Would be a great feature.

hedgieinsocks:
@Nils98Ar I think templating part should be ok, what is needed here is to at least 1) add galera rules and 2) ensure the resulting templates are valid working rules (manual kind deployment to test how it looks and behaves). And of course agree on the wording in the rules and the numbers. The problem is I'm not a galera user, so I have no idea what rules make sense. And I kinda lack the motivation to actively push it further since we use our custom global rules and don't need this feature. And Martin is clearly busy with replication GA epic so I don't revive the talks on it. 

But I'm ready to coop if you or someone else wishes to contribute by reviewing it and suggesting new rules/changes, etc.

Nils98Ar:
I’ll deploy this branch, add the Galera rules, and observe for a while to see how it behaves.

hedgieinsocks:
Thank you! In the meantime, I've added "placeholder" galera rules inspired by the linked mysql-exporter source.

hedgieinsocks:
not stale

hedgieinsocks:
let it sleep here

PR #1358, opened 2025-07-29
Fixed Galera failover

mmontes11:
When Galera primary `Pod` goes down, the operator does not switch writes to a healthy `Pod`.

A regression has been introduced in https://github.com/mariadb-operator/mariadb-operator/pull/1287, fixed in the current PR.


mmontes11:
FYI @hedgieinsocks

PR #1360, opened 2025-08-01
Lower grant reconcile time

hedgieinsocks:
I believe the changes introduced in https://github.com/mariadb-operator/mariadb-operator/pull/1326 inadvertently negatively affected Grant reconcile time.

image

I think it is due to unconditional patching of the status. Perhaps if we skip patch when it is not required, we can avoid extra api requests and latency undesirable at scale.

### DIFF
--- a/internal/controller/grant_controller.go
+++ b/internal/controller/grant_controller.go
@@ -116,10 +116,12 @@ func (wr *wrappedGrantReconciler) Reconcile(ctx context.Context, mdbClient *sqlC
 		return fmt.Errorf("error granting privileges in MariaDB: %v", err)
 	}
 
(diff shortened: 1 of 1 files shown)


hedgieinsocks:
To note, with `--kube-api-qps=-1` reconciliation with 10 workers becomes fast enough to not care, and 20k db/grant/user are reconciled within a couple of minutes. So I'm not sure yet whether this pr is justified.

mmontes11 (approved):
LGTM!

We do unconditional patching in many places, but the good practise is to avoid it when possible and when it is intented to run at scale, like in this case. It won't hurt for this case, but probably won't worth it for many of the unconditional patches we do.

The trade-off will be that the operator Pod will consume more compute resources, as it will now have to compare 20k slices in your case, keep this in mind, and take a look at the metrics as you seem to be doing already.

Thanks for bringing this up!

PR #1401, opened 2025-09-06
Add lvm-localpv error explanation to PhysicalBackup docs

hedgieinsocks:


mmontes11 (approved):
Thanks for documenting this @hedgieinsocks ! 

I have tested local storage, but with [topolvm](https://github.com/topolvm/topolvm) instead, which allows sharing an lvm volume with multiple `Pods` on the same node by default, without any extra configuration option on the `StorageClass`. 

Happy to see that there is this possibility with OpenEBS, as it very widely adopted as well.

PR #1416, opened 2025-09-16
chore: Bump to Go 1.25

Michaelpalacce:
Some necessary changes were done as golangci-lint and go tools needed to be updated.

Linting changes are related to that bump as the newer version has new/stricter rules.

`make gen docs` was ran


hack/get_kind_cidr_prefix/main.go
```diff
@@ -1,16 +1,21 @@
...
 
 func main() {
-	prefix, err := docker.GetKindCidrPrefix()
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	prefix, err := docker.GetKindCidrPrefix(ctx)
 	if err != nil {
 		fmt.Println(err)
+		cancel()
```
mmontes11, line 18: We are deferring this cancel, do we really need it?

test/utils/utils.go
```diff
@@ -61,33 +62,33 @@ func Run(cmd *exec.Cmd) (string, error) {
 }
 
 // InstallPrometheusOperator installs the prometheus Operator to be used to export the enabled metrics.
-func InstallPrometheusOperator() error {
+func InstallPrometheusOperator(ctx context.Context) error {
 	url := fmt.Sprintf(prometheusOperatorURL, prometheusOperatorVersion)
```
mmontes11, line 66: We are still not making use of this, thanks for updating anyway. We will introduce e2e tests at some point

hack/get_kind_cidr_prefix/main.go
Michaelpalacce: Technically, yes.

```
// Exit causes the current program to exit with the given status code.
// Conventionally, code zero indicates success, non-zero an error.
// The program terminates immediately; deferred functions are not run.   -> This
//
// For portability, the status code should be in the range [0, 125].
```
Realistically, in the way the context is used now it makes little to no difference... Still, I think it's a good change

Michaelpalacce:
> We should close the following PR when merged:

I think a few will need to be closed. I will go through the dependabot PRs after merging this and the bump kube to 1.34

PR #1422, opened 2025-09-18
Fix `Grant` finalizing logic to allow `Grants` that point to `Users` with custom name

mmontes11:
Follow up PR https://github.com/mariadb-operator/mariadb-operator/pull/1373

FYI @hedgieinsocks

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -27,7 +27,8 @@
         "MARIADB_GALERA_LIB_PATH": "/usr/lib/galera/libgalera_smm.so",
         "MARIADB_DEFAULT_VERSION": "11.8",
         "WATCH_NAMESPACE": "",
-      }
+      },
+      "dlvFlags": ["--check-go-version=false"] 
     },
     {
       "name": "Webhook",
@@ -41,6 +42,7 @@
         "--log-level=info",
         "--log-time-encoder=iso8601",
       ],
+      "dlvFlags": ["--check-go-version=false"] 
     },
     {
       "name": "Cert controller",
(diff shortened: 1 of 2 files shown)


mmontes11:
@hedgieinsocks after merging https://github.com/mariadb-operator/mariadb-operator/pull/1373 I've noticed that it was not possible to delete invalid `Grants` i.e. pointing to a non existing user. Didn't catch this in my initial review, there are good reasons for that polling

hedgieinsocks:
@mmontes11 Thank you for the heads-up, I'm glad you caught the bug early! But I'd ask you to give it another look. Initially, we did k8s resource lookup for user. But after my PR, we started doing direct sql lookup inside the cluster. So if the user does not really exist in the cluster, we aborted. If it was found, we tried to nuke it. This being said, it's unclear why you faced issues with grant removal for non-existing users. Maybe the real issue was that mariadb in the reference was not available and `Reconcile()` returned error so it got into a loop? I kinda blindly presumed `Reconcile()` was one-time run and if error, ignore and give up.

mmontes11:
Seems like the only issue was that I was missing a cup of coffee ☕ 

Thanks for your responsiveness, this PR has been reverted: https://github.com/mariadb-operator/mariadb-operator/commit/921b2d7e925d7b6db0252b9413681986f6efcf49

The plan is to release it in 25.8.4 shortly

hedgieinsocks:
So if linked mariadb does not exist anymore, there will be no blocking?

mmontes11:
That case is handled here:
https://github.com/mariadb-operator/mariadb-operator/blob/921b2d7e925d7b6db0252b9413681986f6efcf49/pkg/controller/sql/finalizer.go#L61

Whenever the `MariaDB` is deleted, the finalized is removed from SQL resources, therefore the logic discussed here is not executed.

PR #1443, opened 2025-09-26
feature: Allow externalMariadb calls without providing client cert

snaax:
Hello,

I'm using the brand new ExternalMariaDB to connect to an AWS RDS database. We put the option require_secure_transport to 1 to enforce ssl connection. But the thing is that AWS doesn't wait for a client cert, and if you give one, it'll check it against an internal aws truststore and we can't generate a cert that match this truststore.

So I added a mutual field on TLS type (I wanted to just impact ExternalMariaDB but this type is common so I just set true by default so it doesn't impact MariaDB) and added a condition around client certificate and client key options to not add them if we are not in TLS mutual

I'm pretty new to pull requests and go so don't hesitate to come back to me


snaax:
Hi @mmontes11,

I'm sorry to ping you directly but I don't really know if I missed something on the PR to be validated/reviewed. Would it be possible to indicate if I miss a process step.

Thanks!

api/v1alpha1/mariadb_types.go
```diff
@@ -478,6 +478,11 @@ type TLS struct {
 	// +optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:booleanSwitch"}
 	Required *bool `json:"required,omitempty"`
+	// Mutual specifies whether TLS must be mutual between server and client.
+	// It enabled by default.
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:booleanSwitch"}
+	Mutual *bool `json:"mutual,omitempty"`
```
mmontes11: I think this would create some confussion to the `MariaDB` users, if they set it to `false` the connections would present the client certificate anyway. 

Could you create a new `ExternalTLS` struct, which would embed  the current `TLS` struct and additionally inclue the `Mutual` field? This external struct, as you probably guessed, will be used exclusively in the `ExternalMariaDB`.

docs/external_mariadb.md
```diff
@@ -4,8 +4,10 @@
 
 ## Table of contents
 <!-- toc -->
-- [`ExternalMariaDB` configuration](#externalmariadb-configuration)
-- [Supported objects](#supported-objects)
+- [External MariaDB](#external-mariadb)
```
mmontes11, line 7: Could we revert the TOC change? I will take care of updating it if needed, before the release.

docs/external_mariadb.md
```diff
@@ -57,6 +59,32 @@ spec:
 ```
 As a result, you will be able to specify the `ExternalMariaDB` as a reference in [multiple objects](#supported-objects), the same way you would do for a internal `MariaDB` resource.
 
+If you need to use TLS to connect to the external MariaDB but don't want to send the client certificate, you can set the `tls.mutual` field to false and only set the client CA:
```
mmontes11, line 62: Thanks for updating the docs!

internal/controller/external_mariadb_controller_test.go
```diff
@@ -92,6 +92,50 @@ var _ = Describe("External MariaDB spec", func() {
...
+		DeferCleanup(func() {
+			deleteExternalMariadb(key, false)
+		})
+
+		By("Expecting IsTLSMutual to return false")
+		Eventually(func() bool {
+			if err := k8sClient.Get(testCtx, key, &emdb); err != nil {
+				return false
+			}
+			return !emdb.IsTLSMutual()
```
mmontes11, line 154: This is just testing the helper, could you verify that the connection is succeeding like we are doing here?
https://github.com/mariadb-operator/mariadb-operator/blob/36b2354a575875ef1189a9b9ac028336a83ac945/internal/controller/external_mariadb_controller_test.go#L81

mmontes11 (changes requested):
Thank you very much for this PR! Left some comments, could you please take a look?

api/v1alpha1/mariadb_types.go
snaax: Indeed, good idea. I moved it in a dedicated ExternalTLS

docs/external_mariadb.md
snaax: It seems I can't put the old lines back because it's autogenerated when I save. I don't know if it's embedded in visual studio code

docs/external_mariadb.md
snaax: You're welcome :)

internal/controller/external_mariadb_controller_test.go
snaax: I tried to add the conncetion testing part, let me know if it's better

snaax:
Hello,
Thanks a lot for your answer, the rebase and your review.
I tried to make the changes, I hope it fits better.
Let me know if I should change anything else

docs/external_mariadb.md
mmontes11: No prob, I'll amend the docs changes myself. Thanks!

mmontes11:
@snaax Some steps of the CI failed, could you please take a look? New changes look promising, thanks for the effort.

snaax:
@mmontes11 Really sorry about this, I didn't check all steps locally.
The lint now pass on my computer and I've updated the crds.yml using make gen

snaax:
Thanks a lot for your feedback, it feels good to merge the first one 😊

PR #1446, opened 2025-09-29
Refactor/async replication refactoring

Michaelpalacce:
- [ ] Test maxscale

Getting a weird error from the pod:

```log
Error: Failed to initialize groups for 'maxscale: 1 Operation not permitted
```


pkg/controller/replication/switchover.go
Michaelpalacce, line 1: I cna revert the `ctrl.Result{}` changes as that was done for the previous strategy, but it may be worth to keep them since they are done anyway?

Michaelpalacce:
Closing for now will split in multiple smaller prs

PR #1449, opened 2025-09-30
Refactor/set defaults

Michaelpalacce:


api/v1alpha1/mariadb_replication_types.go
```diff
@@ -198,27 +198,18 @@ type Replication struct {
...
-// SetDefaults sets reasonable defaults.
-func (r *Replication) SetDefaults(mdb *MariaDB, env *environment.OperatorEnv) error {
-	if r.GtidStrictMode == nil {
-		r.GtidStrictMode = ptr.To(true)
-	}
-	if reflect.ValueOf(r.InitContainer).IsZero() {
-		r.InitContainer = InitContainer{
-			Image: env.MariadbOperatorImage,
-		}
+func (r *Replication) GetReplica() *ReplicaReplication {
```
mmontes11: Instead of providing methods for safely accessing both primary and replica, we should change the primary and replica types to be structs like we did in Galera. That way we avoid the risk of dereferencing a nil pointer.

api/v1alpha1/mariadb_replication_types.go
```diff
@@ -147,9 +147,9 @@ type ReplicaReplication struct {
 	SyncTimeout *metav1.Duration `json:"syncTimeout,omitempty"`
 }
 
-// FillWithDefaults fills the current ReplicaReplication object with DefaultReplicationSpec.
+// SetDefaults fills the current ReplicaReplication object with DefaultReplicationSpec.
 // This enables having minimal ReplicaReplication objects and provides sensible defaults.
-func (r *ReplicaReplication) FillWithDefaults() {
+func (r *ReplicaReplication) SetDefaults() {
 	if r.WaitPoint == nil {
 		waitPoint := *DefaultReplicationSpec.Replica.WaitPoint
```
mmontes11: Having a separate `DefaultReplicationSpec` does not add any value, and adds another indirection. Let's set its literals here in `SetDefaults`

api/v1alpha1/mariadb_replication_types.go
```diff
@@ -252,20 +243,38 @@ type ReplicationSpec struct {
...
+		}
+		r.InitContainer.Image = initBumped
+	}
+
 	if r.Primary == nil {
 		primary := *DefaultReplicationSpec.Primary
 		r.Primary = &primary
 	} else {
-		r.Primary.FillWithDefaults()
+		r.Primary.SetDefaults()
```
mmontes11: Given that primary will be a struct we can directly call `SetDefaults` here and avoid this conditional defaulting logic.

internal/webhook/v1alpha1/mariadb_webhook.go
```diff
@@ -44,10 +45,14 @@ func (d *MariaDBCustomDefaulter) Default(ctx context.Context, obj runtime.Object
 	}
 	mariadblog.V(1).Info("Defaulting for MariaDB", "name", mariadb.GetName())
 
-	if mariadb.Spec.Replication != nil && mariadb.Spec.Replication.Enabled {
+	env, err := environment.GetOperatorEnv(ctx)
+	if err != nil {
+		return fmt.Errorf("error getting the environment: %v", err)
+	}
+
+	if mariadb.Replication().Enabled {
```
mmontes11: We should align the implementation with Galera, and rely on `IsReplicationEnabled` for this cases.

internal/webhook/v1alpha1/mariadb_webhook.go
```diff
@@ -44,10 +45,14 @@ func (d *MariaDBCustomDefaulter) Default(ctx context.Context, obj runtime.Object
...
 		mariadblog.V(1).Info("Defaulting spec.replication", "mariadb", mariadb.Name)
-		mariadb.Spec.Replication.FillWithDefaults()
-		return nil
+		return mariadb.Spec.Replication.SetDefaults(mariadb, env)
```
mmontes11: We should completely remove the mutating webhook for defaulting, and rely on the `SetDefaults` method invoked during reconciliation.

mmontes11: This also involves removing the `MutatingWebhookConfig` objects from the helm chart and the kubebuilder annotations.

internal/webhook/v1alpha1/mariadb_webhook.go
mariadb-stefangenov: I agree! I did it initially but wanted to limit the scope of the PR to be easier to review. Will add them back!

mmontes11: I think incrementally doing refactors is way better than making any assumptions. This is a big one, so it makes even more sense!

api/v1alpha1/mariadb_replication_types.go
```diff
@@ -268,52 +228,54 @@ type ReplicationSpec struct {
...
 	}
+
+	r.Primary.SetDefaults()
+	r.Replica.SetDefaults()
+
+	return nil
 }
 
 var (
 	tenSeconds = metav1.Duration{Duration: 10 * time.Second}
```
mmontes11: I think we can get rid of this one as well and use a literal.

internal/webhook/v1alpha1/mariadb_webhook_test.go
```diff
@@ -571,72 +571,6 @@ var _ = Describe("v1alpha1.MariaDB webhook", func() {
 				false,
 			),
 		)
-
-		It("Should default replication", func() {
```
mmontes11, line 575: I think it would be worth adding some tests cases here based on the ones we've deleted:
https://github.com/mariadb-operator/mariadb-operator/blob/20426942702b0794d1dae3946914faef94db366a/api/v1alpha1/mariadb_types_test.go#L25

Not all of them, just making sure we ensure that the defaults are set.

internal/webhook/v1alpha1/mariadb_webhook.go
```diff
@@ -172,7 +147,7 @@ func validateGalera(mariadb *v1alpha1.MariaDB) error {
 		if *galera.Primary.PodIndex < 0 || *galera.Primary.PodIndex >= int(mariadb.Spec.Replicas) {
 			return field.Invalid(
 				field.NewPath("spec").Child("galera").Child("primary").Child("podIndex"),
-				mariadb.Replication().Primary.PodIndex,
+				ptr.Deref(mariadb.Spec.Galera, v1alpha1.Galera{}).Primary.PodIndex,
```
mmontes11, line 150: Good catch!

pkg/controller/replication/config.go
```diff
@@ -128,7 +128,7 @@ func (r *ReplicationConfigClient) changeMaster(ctx context.Context, mariadb *mar
 		return fmt.Errorf("error getting replication password: %v", err)
 	}
 
-	gtid := ptr.Deref(mariadb.Replication().Replica.Gtid, mariadbv1alpha1.GtidCurrentPos)
+	gtid := ptr.Deref(mariadb.Spec.Replication.Replica.Gtid, mariadbv1alpha1.GtidCurrentPos)
```
mmontes11: Since this was calling`Replication()`, could we safely derefence `Replication` here? Maybe we can move statement currently in line 142 before this one and use rhe  `replication` variable

api/v1alpha1/mariadb_types.go
```diff
@@ -841,9 +841,9 @@ func (m *MariaDB) SetDefaults(env *environment.OperatorEnv) error {
 			return fmt.Errorf("error setting Galera defaults: %v", err)
 		}
 	}
-	if m.Spec.Replication != nil && m.Spec.Replication.Enabled {
+	if m.IsReplicationEnabled() {
 		if err := m.Spec.Replication.SetDefaults(m, env); err != nil {
-			return fmt.Errorf("error setting replication defaults: %v", err)
+			return fmt.Errorf("error setting Replication defaults: %v", err)
```
mmontes11: The rule I've been following: Galera is a product name, so it can go in capital letter, replication is a feature so it goes lowercase. Errors might not be fully consistent across the codebase, but I think we should try to converge to a standard.

pkg/controller/replication/config.go
mariadb-stefangenov: Yeah! I think this one got lost in the rebase conflicts. Thanks for the catch!

internal/webhook/v1alpha1/mariadb_webhook_test.go
mariadb-stefangenov: I am not sure I understand to be honest, we've removed the webhook and nothing defaults anymore, outside of just go creating a zero struct. Can you elaborate?

mmontes11: We've removed the webhook and its tests, that's fine.

Now the defaulting related to replication is happening in the `SetDefaults` func, therefore I suggest we extend the `SetDefaults` test cases here, adding replication cases:
https://github.com/mariadb-operator/mariadb-operator/blob/20426942702b0794d1dae3946914faef94db366a/api/v1alpha1/mariadb_types_test.go#L27

Issue #1423, opened 2025-09-18
Replication GA


mmontes11:
Replication feature will be generally available, see:
- https://github.com/mariadb-operator/mariadb-operator/pull/1441

mmontes11:
> Disable replicasAllowEvenNumber for replication

@Michaelpalacce I think this could be a good one for you. It has been contributed in this PR:
- https://github.com/mariadb-operator/mariadb-operator/pull/783

See the related code:
- https://github.com/mariadb-operator/mariadb-operator/blob/20426942702b0794d1dae3946914faef94db366a/api/v1alpha1/mariadb_types.go#L779
- https://github.com/mariadb-operator/mariadb-operator/blob/20426942702b0794d1dae3946914faef94db366a/api/v1alpha1/mariadb_types.go#L638

Replication is not quorum based, therefore an odd number of replicas is not required. Therefore this validation should be disabled when replication is enabled.

mmontes11:
> Refactor APIs, see https://github.com/mariadb-operator/mariadb-operator/issues/1436#issuecomment-3327669683

@Michaelpalacce regarding this, I suggest the following refactoring of the `replication` field:

```yaml
  replication:
    enabled: true
    primary:
      podIndex: 0
      automaticFailover: true
      automaticFailoverDelay: 0s
    replica:
      maxLagSeconds: 0s # Do set a default here. Any lag should be taken into account
      replPasswordSecretKeyRef:
        name: "$MARIADB-repl-password"
        key: password
        generate: false
      gtid: CurrentPos # Do set a default here, CurrentPos is recommended when using the GTID strict mode.
      connectionRetries: 10 # Maps to a field in CHANGE MASTER. Don't set a default, leave it to its zero value,  so we rely on the server default.
      syncTimeout: 10s # This is used by the switchover. When timing out it is requeued, so we can keep 10s
    gtidStrictMode: true # Do set a default to true here. By the default the server does not enforce this, but it is very convenient to prevent 
    ackTimeout: 10s # Previous spec.replica.connectionTimeout. Maps to rpl_semi_sync_master_timeout sys var.  Don't set a default, leave it to its zero value,  so we rely on the server default.
    waitPoint: AfterSync # Maps to rpl_semi_sync_master_wait_point sys var.  Don't set a default, leave it to its zero value,  so we rely on the server default.
    syncBinlog: 1 # Maps to sync_binlog. Don't set a default, leave it to its zero value,  so we rely on the server default.
[...]

PR #1452, opened 2025-10-01
Replication backups

mmontes11:
Related to https://github.com/mariadb-operator/mariadb-operator/issues/1423


api/v1alpha1/mariadb_replication_types.go
```diff
@@ -307,28 +341,53 @@ type ReplicationState string
...
 )
 
-// ReplicationStatus is the replication current status per each Pod.
-type ReplicationStatus map[string]ReplicationState
+// ReplicationStatus is the replication current state.
+type ReplicationStatus struct {
+	// State is the observed replication state for each Pod.
+	State map[string]ReplicationState `json:"state,omitempty"`
+	// ReplicasToConfigure are the replicas that need to configure replication.
+	ReplicasToConfigure []string `json:"replicasToConfigure,omitempty"`
```
Michaelpalacce, line 353: It's unfortunate that we don't have a better way to do this

api/v1alpha1/mariadb_replication_types.go
```diff
@@ -307,28 +341,53 @@ type ReplicationState string
...
 	ReplicationStateNotConfigured ReplicationState = "NotConfigured"
 )
 
-// ReplicationStatus is the replication current status per each Pod.
-type ReplicationStatus map[string]ReplicationState
+// ReplicationStatus is the replication current state.
+type ReplicationStatus struct {
+	// State is the observed replication state for each Pod.
+	State map[string]ReplicationState `json:"state,omitempty"`
+	// ReplicasToConfigure are the replicas that need to configure replication.
```
Michaelpalacce:
```suggestion
	// ReplicasToConfigure are the replicas that the operator will configure replication on.
```

cmd/agent/main.go
```diff
@@ -78,6 +81,16 @@ var RootCmd = &cobra.Command{
 	Long:  "Sidecar agent that co-operates with mariadb-operator.",
 }
 
+func newContext() (context.Context, context.CancelFunc) {
```
Michaelpalacce, line 84: I wonder if it will be more idiomatic if we do

```go
// withSignal returns a derived context cancellable by system signals
func withSignal(ctx context.Context) (context.Context, context.CancelFunc) {
    return signal.NotifyContext(ctx, []os.Signal{
		syscall.SIGINT,
		syscall.SIGTERM,
		syscall.SIGKILL,
		syscall.SIGHUP,
		syscall.SIGQUIT}...,
	)
}
```

cmd/agent/replication.go
```diff
@@ -35,26 +35,69 @@ var replicationCommand = &cobra.Command{
...
 			logger.Error(err, "Error creating probe server")
 			os.Exit(1)
 		}
 
-		ctx, cancel := signal.NotifyContext(context.Background(), syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)
+		ctx, cancel := newContext()
 		defer cancel()
+		errChan := make(chan error, 2)
+
+		var wg sync.WaitGroup
```
Michaelpalacce, line 76: Wouldn't `errgroup` package be usefull here too?

mmontes11: Thought about that as well, but decided to leave it as it is in galera currently. I remember having issues with the graceful shutdown and goroutine management, so I didn't want to get into the rabbit hole again and this case is quite simple, just 2 goroutines. We can take a note for this.

cmd/agent/main.go
mmontes11: This is more like a [functional options pattern ](https://dev.to/kittipat1413/understanding-the-options-pattern-in-go-390c), here we are just creating a context

api/v1alpha1/mariadb_replication_types.go
mmontes11: I don't like to statefully reconcile, it could lead to problems if the state is not properly handled i.e. there could be coner cases you don't handle well.  The motivation behind this change is handling the replica configuration in a single place:
- https://github.com/mariadb-operator/mariadb-operator/blob/ce503b530d30b6870b2a07a051d1fe30a826024b/pkg/controller/replication/controller.go#L221

And only in this place we could mark the replica as configured:
- https://github.com/mariadb-operator/mariadb-operator/blob/ce503b530d30b6870b2a07a051d1fe30a826024b/pkg/controller/replication/controller.go#L239

Therefore, althought it is stateful, this allows to trigger the configuration from previous reconciliation steps (`Init`, `ScaleOut`, `Recovery`) but only [here](https://github.com/mariadb-operator/mariadb-operator/blob/ce503b530d30b6870b2a07a051d1fe30a826024b/pkg/controller/replication/controller.go#L239) is where it is marked as configured. 

I will need to trigger this during the scale out and replica recovery operations, there will be chances to rethink this and test its robustness.

cmd/agent/replication.go
Michaelpalacce: Added task internally

cmd/agent/main.go
Michaelpalacce: Yes it is, still could have benefitted, pretty minor tho

api/v1alpha1/mariadb_replication_types.go
Michaelpalacce: As we talked initially I think it's a potential approach. The status is supposed to hold the latest observed state and this is exactly that from our point of view. I would say we are sticking to best practices.

PR #1454, opened 2025-10-03
Refactor/async repl auth

Michaelpalacce:
This PR aims to expand the Auth controller and give us the ability to create users and grant them privileges via the headless service if needed. This is required for replication, as the clusterIP service will not be ready due to probes.

PR also adds tests to validate this functionality


Michaelpalacce:
Closing this now, will re-open at a later point @mmontes11

PR #1459, opened 2025-10-06
feat: replicasAllowEvenNumber will now work only for galera

Michaelpalacce:


mmontes11 (changes requested):
Got the following error when testing this with a replication instance:

```bash
❯ kaf examples/manifests/mariadb_replication_scaleout_physicalbackup.yaml
The MariaDB "mariadb-repl" is invalid: spec: Invalid value: "object": no such key: galera evaluating rule: An odd number of MariaDB instances (mariadb.spec.replicas) is required to avoid split brain situations for Galera. Use 'mariadb.spec.replicasAllowEvenNumber: true' to disable this validation.
```

Here my manifest:
```yaml
kind: MariaDB
metadata:
  name: mariadb-repl
spec:
  rootPasswordSecretKeyRef:
    name: mariadb
    key: root-password

  storage:
    size: 1Gi
    storageClassName: csi-hostpath-sc

  # Increase number of replicas to trigger scaleout process
  replicas: 4

  replication:
    enabled: true
    # replica:
    #   bootstrapFrom:
    #     physicalBackupTemplateRef:
    #       name: physicalbackup

  service:
    type: LoadBalancer
    metadata:
      annotations:
        metallb.universe.tf/loadBalancerIPs: 172.18.0.120

  primaryService:
    type: LoadBalancer
    metadata:
      annotations:
        metallb.universe.tf/loadBalancerIPs: 172.18.0.130

  secondaryService:
    type: LoadBalancer
    metadata:
      annotations:
        metallb.universe.tf/loadBalancerIPs: 172.18.0.131

  metrics:
    enabled: true
```

Galera object may not be set

Michaelpalacce:
> Got the following error when testing this with a replication instance:
> 
> ```shell
> ❯ kaf examples/manifests/mariadb_replication_scaleout_physicalbackup.yaml
> The MariaDB "mariadb-repl" is invalid: spec: Invalid value: "object": no such key: galera evaluating rule: An odd number of MariaDB instances (mariadb.spec.replicas) is required to avoid split brain situations for Galera. Use 'mariadb.spec.replicasAllowEvenNumber: true' to disable this validation.
> ```
> 
> Here my manifest:
> 
> ```yaml
> ...
> ```
> 
> Galera object may not be set

Addressed

api/v1alpha1/mariadb_types_validation_test.go
```diff
@@ -0,0 +1,158 @@
...
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+)
+
+var _ = Describe("MariaDB validation", func() {
+	Context("When creating a MariaDB Resource", func() {
```
mmontes11: Could you add it to this block? 

https://github.com/mariadb-operator/mariadb-operator/blob/20426942702b0794d1dae3946914faef94db366a/api/v1alpha1/mariadb_types_test.go#L23

mmontes11 (changes requested):

api/v1alpha1/mariadb_types_validation_test.go
Michaelpalacce: Addressed

mmontes11:
@Michaelpalacce don't worry about rebase, triggered it myself

api/v1alpha1/mariadb_types_test.go
```diff
@@ -2259,4 +2259,151 @@ var _ = Describe("MariaDB types", func() {
...
+					Expect(err).ToNot(HaveOccurred(), "Did not expect there to be a validation error, but there was one.")
+				}
+			},
+
+			Entry(
+				"Valid replicas when not even",
+				&MariaDB{
+					ObjectMeta: meta,
+					Spec: MariaDBSpec{
+						Replicas: 3,
```
mmontes11: Technically, multiple replicas should only be allowed whenever replication or Galera is enabled, that's covered by one of the webhook validations. But I'm guessing that the API suite does not have the webhook setup? We can omit it, but probably worth removing this cases to avoid confusion.

api/v1alpha1/mariadb_types_test.go
```diff
@@ -2259,4 +2259,151 @@ var _ = Describe("MariaDB types", func() {
...
+						Storage: Storage{
+							Size: ptr.To(resource.MustParse("100Mi")),
+						},
+					},
+				},
+				false,
+				"",
+			),
+			Entry(
+				"Valid replicas when even",
```
mmontes11: See my previous comment

api/v1alpha1/mariadb_types_test.go
Michaelpalacce: I will set it as 1, just for very basic testing, wil lremove the next one

PR #1460, opened 2025-10-06
refactor: Moved `waitPoint` from `spec.replication.replica` to `spec.replication`

Michaelpalacce:
### ⚠️ Changlelog 
- Moved `spec.replication.replica.waitPoint` to `spec.replication.waitPoint` 
- Remove `spec.replication.waitPoint`  default and rely on server defaults.


mmontes11:
@Michaelpalacce don't worry, it is rebased

PR #1467, opened 2025-10-08
Feature/migration script for 25.10.0

Michaelpalacce:


hack/migrate_v25.10.0.sh
```diff
@@ -0,0 +1,95 @@
+#!/bin/bash
+
+set -eo pipefail
+
+MARIADB_INPUT="$1"
+MARIADB_OUTPUT="migrated.$MARIADB_INPUT"
+MARIADB_STATUS_OUTPUT="status.$MARIADB_INPUT"
+
+if [ -z "$MARIADB_INPUT" ]; then
+  echo "Error: MariaDB manifest file from a version older than v0.0.26 must be provided as the first argument."
```
mmontes11:
```suggestion
  echo "Error: MariaDB manifest file from a version older than 25.8.4 must be provided as the first argument."
```

hack/migrate_v25.10.0.sh
```diff
@@ -0,0 +1,95 @@
...
+    del(.spec.replication.replica.connectionTimeout)
+  ' -i "$MARIADB_OUTPUT"
+fi
+
+ echo "Creating status patch..."
+ cp "$MARIADB_INPUT" "$MARIADB_STATUS_OUTPUT"
+
+"$YQ" '. |= pick(["status"])' -i "$MARIADB_STATUS_OUTPUT"
+
+# Show a summary if `diff` is installed. Which it should be on Linux and Mac, not sure for Windows...
```
mmontes11:
```suggestion
# Show a summary if `diff` is installed.
```

hack/migrate_v25.10.0.sh
Michaelpalacce: Ah, the copy-pasta....

PR #1471, opened 2025-10-10
Test/add restore from physical backup tests

Michaelpalacce:


internal/controller/mariadb_controller_replication_test.go
```diff
@@ -384,3 +286,336 @@ var _ = Describe("MariaDB replication", Ordered, func() {
...
+			}, testHighTimeout, testInterval).Should(BeTrue())
+		},
+		Entry(
+			"from physical backup",
+			types.NamespacedName{Name: "replication-pvc-backup-test", Namespace: key.Namespace},
+			buildPhysicalBackupWithPVCStorage(key),
+			func(backupKey types.NamespacedName) *mariadbv1alpha1.BootstrapFrom {
+				return &mariadbv1alpha1.BootstrapFrom{
+					BackupContentType: mariadbv1alpha1.BackupContentTypePhysical,
+					Volume: &mariadbv1alpha1.StorageVolumeSource{
```
mmontes11: Although we do support PVCs with `PhysicalBackups`, considering the time constraints for running tests, I think it would be a better use of our time to test the bootstrap from S3 path, which is way more valuable than PVCs

internal/controller/mariadb_controller_replication_test.go
```diff
@@ -384,3 +286,336 @@ var _ = Describe("MariaDB replication", Ordered, func() {
...
+				max_allowed_packet=256M`,
+			),
+			Replication: &mariadbv1alpha1.Replication{
+				ReplicationSpec: mariadbv1alpha1.ReplicationSpec{
+					Primary: mariadbv1alpha1.PrimaryReplication{
+						PodIndex:          ptr.To(0),
+						AutomaticFailover: ptr.To(true),
+					},
+					Replica: mariadbv1alpha1.ReplicaReplication{
+						Gtid: ptr.To(mariadbv1alpha1.GtidCurrentPos),
```
mmontes11: Could we test the default settings i.e. remove the `Gtid` field?

internal/controller/mariadb_controller_replication_test.go
```diff
@@ -384,3 +286,336 @@ var _ = Describe("MariaDB replication", Ordered, func() {
...
+						client.InNamespace(backupKey.Namespace),
+					}
+					Expect(k8sClient.DeleteAllOf(testCtx, &volumesnapshotv1.VolumeSnapshot{}, opts...)).To(Succeed())
+				}
+			},
+		),
+	)
+})
+
+func buildTestMariaDB(key types.NamespacedName) *mariadbv1alpha1.MariaDB {
```
mmontes11:
```suggestion
func buildTestMariaDBRepl(key types.NamespacedName) *mariadbv1alpha1.MariaDB {
```

as it has replication enabled, and the name could clas with other funcs in this pkg

internal/controller/mariadb_controller_replication_test.go
```diff
@@ -384,3 +286,336 @@ var _ = Describe("MariaDB replication", Ordered, func() {
...
+					PhysicalBackupTemplateRef: mariadbv1alpha1.LocalObjectReference{
+						Name: backupKey.Name,
+					},
+				}
+
+				return k8sClient.Update(testCtx, mdb) == nil
+			}, testTimeout, testInterval).Should(BeTrue())
+
+			By("Expecting MariaDB to be ready eventually")
+			Eventually(func() bool {
```
mmontes11, line 467: Could you add an assertion of the number of replicas?

internal/controller/mariadb_controller_replication_test.go
```diff
@@ -384,3 +286,336 @@ var _ = Describe("MariaDB replication", Ordered, func() {
...
+			func(backupKey types.NamespacedName) func() {
+				return func() {
+					By("Deleting Backup Resources")
+					var pvc corev1.PersistentVolumeClaim
+					Expect(k8sClient.Get(testCtx, backupKey, &pvc)).To(Succeed())
+					Expect(k8sClient.Delete(testCtx, &pvc)).To(Succeed())
+				}
+			},
+		),
+		// VolumeSnapshot doesn't get created, to investigate
```
mmontes11: I believe this is resolved, right?

Michaelpalacce: Yes!

PR #1483, opened 2025-10-20
Test/add replica recovery integration tests

Michaelpalacce:


internal/controller/mariadb_controller_replica_recovery_test.go
```diff
@@ -103,7 +112,7 @@ var _ = Describe("isRecoverableError", func() {
 							Replica: mariadbv1alpha1.ReplicaReplication{
 								ReplicaRecovery: &mariadbv1alpha1.ReplicaRecovery{
 									Enabled:                true,
-									ErrorDurationThreshold: &metav1.Duration{Duration: 30 * time.Second},
+									ErrorDurationThreshold: &metav1.Duration{Duration: 15 * time.Second},
```
mmontes11: Did you face any issue with this value i.e. the test failing unexpectedly?

internal/controller/mariadb_controller_replica_recovery_test.go
```diff
@@ -114,3 +123,259 @@ var _ = Describe("isRecoverableError", func() {
...
+			query = `PURGE BINARY LOGS BEFORE NOW();`
+			executeSqlInPodByIndex(mdb, 0, query)
+
+			By("Deleting the First Replica PVC")
+			deletePVCByPodIndex(mdb, podIndexToDelete)
+
+			By("Deleting the First Replica Pod")
+			deletePodByIndex(mdb, podIndexToDelete)
+
+			By("Flushing Binary Logs")
```
mmontes11: Why is this executed multiple times?

internal/controller/mariadb_controller_replica_recovery_test.go
```diff
@@ -114,3 +123,259 @@ var _ = Describe("isRecoverableError", func() {
...
+			query = `FLUSH LOGS;`
+			executeSqlInPodByIndex(mdb, 0, query)
+			query = `PURGE BINARY LOGS BEFORE NOW();`
+			executeSqlInPodByIndex(mdb, 0, query)
+
+			// Otherwise the `pod` doesn't get deleted and gets stuck in `Completed`
+			By("Removing PVC finalizer after a short delay")
+			time.Sleep(10 * time.Second)
+
+			By("Flushing Binary Logs")
```
mmontes11: Same, see my previous comment.

internal/controller/mariadb_controller_replica_recovery_test.go
```diff
@@ -114,3 +123,259 @@ var _ = Describe("isRecoverableError", func() {
...
+						ErrorDurationThreshold: ptr.To(metav1.Duration{Duration: time.Second * 15}),
+					},
+				}
+
+				return k8sClient.Update(testCtx, mdb) == nil
+			}, testTimeout, testInterval).Should(BeTrue())
+
+			By("Flushing Binary Logs")
+			query := `FLUSH LOGS;`
+			executeSqlInPodByIndex(mdb, 0, query)
```
mmontes11: Instead of deleting the default `0` index, delete the index pointed by `status.currentPrimaryPodIndex`. It would make  the test more readable and flexible.

internal/controller/mariadb_controller_replica_recovery_test.go
```diff
@@ -114,3 +123,259 @@ var _ = Describe("isRecoverableError", func() {
...
+			query = `PURGE BINARY LOGS BEFORE NOW();`
+			executeSqlInPodByIndex(mdb, 0, query)
+
+			var pvc corev1.PersistentVolumeClaim
+			pvcKey := mdb.PVCKey(builder.StorageVolume, podIndexToDelete)
+			err := k8sClient.Get(testCtx, pvcKey, &pvc)
+			if apierrors.IsNotFound(err) {
+				return
+			}
+			Expect(err).NotTo(HaveOccurred())
```
mmontes11: This could lead to flaky tests, for example, if the PVC is not completely termined and the `Expect` executes, the entire test will fail. PVC are asynchronously terminated so we need to run this in a `Evantually` block. Also, we shouldn't `return` during a test, it will skip the assertions done ahead, but... if you wrap it with a `Eventually` and return false, it gets retried until reaching the timeout.

Please wrap it with a `Eventually` to  fail the test after a `testTimeout` duration if the PVC is not completely terminated. I would include the finalizing logic within the function block of `Eventually` as well i.e. removing the finalizer.

internal/controller/mariadb_controller_replica_recovery_test.go
```diff
@@ -114,3 +123,259 @@ var _ = Describe("isRecoverableError", func() {
...
+				Expect(client.IgnoreNotFound(k8sClient.Delete(testCtx, backup))).To(Succeed())
+			})
+
+			DeferCleanup(func() {
+				deletePhysicalBackup(backupKey)
+				cleanupFn(backupKey)()
+			})
+
+			By("Bootstrapping Recovery")
+			Eventually(func() bool {
```
mmontes11, line 181: What is the motivation of this `Eventually` block? Did you get any error updating the `MariaDB` resource? At this point we are just updating the resource, not triggering the recovery.

internal/controller/mariadb_controller_replica_recovery_test.go
```diff
@@ -114,3 +123,259 @@ var _ = Describe("isRecoverableError", func() {
...
+			},
+			Service: &mariadbv1alpha1.ServiceTemplate{
+				Type: corev1.ServiceTypeLoadBalancer,
+				Metadata: &mariadbv1alpha1.Metadata{
+					Annotations: map[string]string{
+						"metallb.universe.tf/loadBalancerIPs": testCidrPrefix + ".0.120",
+					},
+				},
+			},
+			Connection: &mariadbv1alpha1.ConnectionTemplate{
```
mmontes11: This can be omitted.

internal/controller/mariadb_controller_replica_recovery_test.go
```diff
@@ -114,3 +123,259 @@ var _ = Describe("isRecoverableError", func() {
...
+			},
+			PrimaryService: &mariadbv1alpha1.ServiceTemplate{
+				Type: corev1.ServiceTypeLoadBalancer,
+				Metadata: &mariadbv1alpha1.Metadata{
+					Annotations: map[string]string{
+						"metallb.universe.tf/loadBalancerIPs": testCidrPrefix + ".0.130",
+					},
+				},
+			},
+			PrimaryConnection: &mariadbv1alpha1.ConnectionTemplate{
```
mmontes11: This can be omitted.

mmontes11 (changes requested):

internal/controller/mariadb_controller_replica_recovery_test.go
Michaelpalacce: No, not necessarily, just lowered it so the tests are reasonable and fast at the same time.

internal/controller/mariadb_controller_replica_recovery_test.go
Michaelpalacce: Originally I observed flakiness while testing and wanted to make sure that it's not due to logs, will try again and see if any flakiness happens

internal/controller/mariadb_controller_replica_recovery_test.go
Michaelpalacce: Got it, will do

internal/controller/mariadb_controller_replica_recovery_test.go
Michaelpalacce: I can revert it?

internal/controller/mariadb_controller_replica_recovery_test.go
Michaelpalacce: I was following the example I've seen in other places when update operations were done

internal/controller/mariadb_controller_replica_recovery_test.go
Michaelpalacce: It can, my bad. I took this partially from a function and missed this

internal/controller/mariadb_controller_replica_recovery_test.go
Michaelpalacce: Will revert this

internal/controller/mariadb_controller_replica_recovery_test.go
mmontes11: If that's the case, consider wrapping them inside a `Eventually` til succeeds.

mmontes11 (approved):

PR #1492, opened 2025-10-25
Add explicit ns in mariadb-cluster chart

hedgieinsocks:
This should make templating predictable: https://github.com/helm/helm/issues/3553

P.S. I don't need this feature for mariadb-operator chart, thus this pr is scoped to mariadb-cluster chart only. But you are welcome to port it if need be.


mmontes11 (approved):
LGTM. Thanks for the contribution. This will be released in`25.10.2` with other replication fixes.

I'll update the remaining helm charts in upcoming releases.

PR #1515, opened 2025-11-14
Using Specific helm version for locally and github workflows

Michaelpalacce:
This PR modifies make targets and some `hack` files to use a local installation of Helm. Additionally, we are bumping the helm version in the Github Workflows.

This is done after the new Helm 4.0.0 release during Kubecon

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -35,7 +35,7 @@ jobs:
             echo "Version('$VERSION') and Makefile version('$MAKE_VERSION') don't match!"
             exit 1;
           fi
-          
+
           RELEASE_ARGS=release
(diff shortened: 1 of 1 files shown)


make/helm.mk
```diff
@@ -82,11 +82,11 @@ endif
 ifndef HELM_CLUSTER_CHART_FILE
 	$(error HELM_CLUSTER_CHART_FILE is not set. Please set it before running this target)
 endif
-ifndef HELM_VERSION
-	$(error HELM_VERSION is not set. Please set it before running this target)
+ifndef HELM_CHART_VERSION
```
mmontes11, line 85: This variable needs to be updated here:
https://github.com/mariadb-operator/mariadb-operator/blob/36b2354a575875ef1189a9b9ac028336a83ac945/.github/workflows/release.yml#L149

Michaelpalacce: Done

PR #1517, opened 2025-11-22
Release 26.03: `PointInTimeRecovery`, Azure Blob Storage & on-demand `PhysicalBackups`

mmontes11:
Close https://github.com/mariadb-operator/mariadb-operator/issues/507

## TODO

- [x] Documentation
- [x] Examples catalog
- [x] Migration guide
- [x] Release notes
- [x] [MariaDB docs](https://github.com/mariadb-corporation/mariadb-docs/pull/395)
- [x] Add `PointInTimeRecovery` to [mariadb-operator-helm](https://github.com/mariadb-operator/mariadb-operator-helm):  https://github.com/mariadb-operator/mariadb-operator-helm/commit/34751caa866127452a5437adb2b271b7b31591f6
- [x] https://github.com/mmontes11/k8s-infrastructure/pull/21
- [x] https://github.com/mmontes11/k8s-media/pull/2
 
## PRs

- https://github.com/mariadb-operator/mariadb-operator/pull/1517
- https://github.com/mariadb-operator/mariadb-operator/pull/1630
- https://github.com/mariadb-operator/mariadb-operator/pull/1635
- https://github.com/mariadb-operator/mariadb-operator/pull/1610
- https://github.com/mariadb-operator/mariadb-operator/pull/1632
- https://github.com/mariadb-operator/mariadb-operator/pull/1631
- https://github.com/mariadb-operator/mariadb-operator/pull/1606
- https://github.com/mariadb-operator/mariadb-operator/pull/1650
- https://github.com/mariadb-operator/mariadb-operator/pull/1601
- https://github.com/mariadb-operator/mariadb-operator/pull/1589
- https://github.com/mariadb-operator/mariadb-operator/pull/1588
- https://github.com/mariadb-operator/mariadb-operator/pull/1575

## ⚠️ BREAKING changes ⚠️ :
- Logical and physical backups file whose timestamp is closest to `targetRecoveryTime`, __but not after__, will be matched. This is to align with PITR restoration logic.

## PITR
- [x] Pause `gtid_strict_mode` before replaying binlogs
- [x] Cleanup PITR job after replaying binlogs
- [x] Improve logging and reduce verbosity as needed
- [x] Strict mode 
- [x] Last recoverable time
- [x] Azure blob storage support for `PhysicalBackup` and `PointInTimeRecovery`
  - https://github.com/mariadb-operator/mariadb-operator/pull/1610
[...]

PR #1520, opened 2025-11-26
Backup & Restore: support for SSE-C

### LINKED ISSUE #1084: [Feature] S3 encryptio via SSE-C

While backup encryption is still an open issue (#652), a separate desire would be to provide data at rest encryption for externally managed storage as offered via S3-compatible interfaces.

Implement SSE-C.

Backup encryption, which is still a work in progress and less trivial to implement than adding a few request headers for security improvement, depending on provider trust.

**Additional context**

[SSE-C](https://min.io/docs/minio/linux/administration/server-side-encryption/server-side-encryption-sse-c.html) is supported by MinIO and AWS, and it really revolves around adding a few request headers and then trusting the opposing party to play ball and treat the key as ephemeral. [Developer documentation in the AWS SDK](https://docs.aws.amazon.com/AmazonS3/latest/userguide/ServerSideEncryptionCustomerKeys.html)

xavierleune:
This pull requests adds support for SSE-C, thanks to minio-go that already supports it.

Fixes #1084

### DIFF
--- a/docs/logical_backup.md
+++ b/docs/logical_backup.md
@@ -95,7 +95,7 @@ kind: Secret
 metadata:
   name: ssec-key
 type: Opaque
-data:
+stringData:
   # 32-byte key encoded in base64 (use: openssl rand -base64 32)
   customer-key: YWJjZGVmZ2hpamtsbW5vcHFyc3R1dnd4eXoxMjM0NTY=
 ```


mmontes11:
Thanks for this contribution @xavierleune, we are willing to support this.

I'm on holiday currently, but @Michaelpalacce might be able to do an initial review

xavierleune:
Thanks @mmontes11, enjoy your holiday !
@Michaelpalacce I generated the missing docs and so, I think the next CI should be all green !

cmd/backup/main.go
```diff
@@ -16,6 +16,10 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
+const (
+	s3SSECCustomerKeyEnv = "MARIADB_OPERATOR_S3_SSEC_CUSTOMER_KEY"
```
mmontes11, line 20: This could be added to the `PodEnvironment` struct:
https://github.com/mariadb-operator/mariadb-operator/blob/0533efca9105c2f9724c523dda8793b86d7cf045/pkg/environment/environment.go#L61

Could you you add it to the struct as optional? Then we can access the value using the struct, without additional `os.Getenv` calls.

docs/logical_backup.md
```diff
@@ -84,6 +85,58 @@ spec:
 ```
 By providing the authentication details and the TLS configuration via references to `Secret` keys, this example will store the backups in a local Minio instance.
 
+#### Server-Side Encryption with Customer-Provided Keys (SSE-C)
```
mmontes11, line 88: This section should go after `Alternatively you can use dynamic credentials from an EKS Service Account using EKS Pod Identity or IRSA:`

docs/logical_backup.md
```diff
@@ -84,6 +85,58 @@ spec:
...
+
+You can enable server-side encryption using your own encryption key (SSE-C) by providing a reference to a `Secret` containing a 32-byte (256-bit) key encoded in base64:
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: ssec-key
+type: Opaque
+data:
```
mmontes11:
```suggestion
stringData:
```
The provided customer key is base64, if we specify it as `data`, then it will be decoded when mounted by the backup/restore Pod, which is not what we expect, as we are decoding the customer key as part of the process.  

Using this example as it is results in the backup process erroring, as the customer key is not a valid base64, as it gets decoded by Kubernetes before mounted to the Pod. Using `stringData` fixes the problem, it avoids decoding, and the Pod sees the base64

Could you confirm if this is correct? Maybe I am missing something

docs/physical_backup.md
```diff
@@ -318,6 +319,56 @@ spec:
...
+
+You can enable server-side encryption using your own encryption key (SSE-C) by providing a reference to a `Secret` containing a 32-byte (256-bit) key encoded in base64:
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: ssec-key
+type: Opaque
+data:
```
mmontes11:
```suggestion
stringData:
```

mmontes11: Same issue, please see my previous comment.

docs/logical_backup.md
xavierleune: Good catch ! This example should use `stringData` instead of `data`

cmd/backup/main.go
xavierleune: Shouldn't we have a `BackupEnvironment` or something like that ? I'm not sure if `PodEnvironment` is the best place to hold the encryption key, as it should be available only for backup and restore purpose, not all pods. WDYT ?

mmontes11: The naming is a bit ambious to be honest. `PodEnvironment` only applies to the operand Pods i.e. the MariaDB and the backup/restore Pods. Additionally, this will be used in the data-plane when this gets merged:
- https://github.com/mariadb-operator/mariadb-operator/pull/1517

Please, add it to `PodEnvironment` and I will refactor it later to something more specific, thinking about `OperandEnvironment`, open to suggestions.

mmontes11: Actually, forget about this, we are not using `PodEnvironment` in the backup/restore yet, my bad. It looks good as it currently is.

mmontes11 (approved):
LGTM! Thanks for this contribution @xavierleune ! I'll make the adaptations of the docs myself

xavierleune:
Thanks @mmontes11

Issue #1438, opened 2025-09-25
[Help] How to set longer timeout on operator


siwyroot:
Hello, we have a cluster where operator timeouts on reconcile on resources listing, we know it's related to high load on kubeapi, is there a way to extend that timeout? Operator is installed on OpenShift via operator hub, OCP is in version 4.18. Problem is with mariadb-operator pod

```
{"level":"info","ts":1758822217.6244504,"msg":"Stopping and waiting for HTTP servers"}
{"level":"info","ts":1758822217.6244879,"logger":"controller-runtime.metrics","msg":"Shutting down metrics server with timeout of 1 minute"}
{"level":"info","ts":1758822217.6244948,"msg":"shutting down server","name":"health probe","addr":"[::]:8081"}
{"level":"info","ts":1758822217.624747,"msg":"Wait completed, proceeding to shutdown the manager"}
{"level":"error","ts":1758822217.6248126,"logger":"setup","msg":"Error running manager","error":"failed to wait for pod-replication caches to sync kind source: *v1.Pod: timed out waiting for cache to be synced for Kind *v1.Pod","stacktrace":"main.init.func2\n\t/app/cmd/controller/main.go:544\ngithub.com/spf13/cobra.(*Command).execute\n\t/go/pkg/mod/github.com/spf13/cobra@v1.9.1/command.go:1019\ngithub.com/spf13/cobra.(*Command).ExecuteC\n\t/go/pkg/mod/github.com/spf13/cobra@v1.9.1/command.go:1148\ngithub.com/spf13/cobra.(*Command).Execute\n\t/go/pkg/mod/github.com/spf13/cobra@v1.9.1/command.go:1071\nmain.main\n\t/app/cmd/controller/main.go:557\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:283"}
```

siwyroot:
Problem is with mariadb-operator pod

crabique:
It is [2 minutes](https://github.com/kubernetes-sigs/controller-runtime/blob/8539590135efa55e3eaff85bf5386929146f1ad6/pkg/controller/controller.go#L51-L53) by default, are you sure giving it more time will actually fix the problem in your setup?

Have you tried `extraArgs: ["--requeue-sql=0",  "--sync-period=0"]` and raising QPS/burst values?


If nothing else helps and you actually need to increase this timeout, unfortunately at the moment there's no mechanism to configure this. But it should be fairly easy to add here:
[...]

PR #1521, opened 2025-11-27
Add cacheSyncTimeout

hedgieinsocks:
Fixes: https://github.com/mariadb-operator/mariadb-operator/issues/1438

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -388,9 +388,7 @@ var rootCmd = &cobra.Command{
 			MaxScaleReconciler:    mxsReconciler,
 			ReplicationReconciler: replicationReconciler,
 			GaleraReconciler:      galeraReconciler,
-		}).SetupWithManager(ctx, mgr, env, ctrlcontroller.Options{
-			MaxConcurrentReconciles: mariadbMaxConcurrentReconciles,
-			CacheSyncTimeout:        cacheSyncTimeout}); err != nil {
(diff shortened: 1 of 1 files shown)


cmd/controller/main.go
```diff
@@ -385,7 +388,9 @@ var rootCmd = &cobra.Command{
 			MaxScaleReconciler:    mxsReconciler,
 			ReplicationReconciler: replicationReconciler,
 			GaleraReconciler:      galeraReconciler,
-		}).SetupWithManager(ctx, mgr, env, ctrlcontroller.Options{MaxConcurrentReconciles: mariadbMaxConcurrentReconciles}); err != nil {
+		}).SetupWithManager(ctx, mgr, env, ctrlcontroller.Options{
+			MaxConcurrentReconciles: mariadbMaxConcurrentReconciles,
+			CacheSyncTimeout:        cacheSyncTimeout}); err != nil {
```
mmontes11: I think the `CacheSyncTimeout` is a bit redundant at the reconciler level if we are already setting the same value at the manager level. The only reason I see for setting it at the reconciler level is when having a specific value for each reconciler, like we did with `MaxConcurrentReconciles`, but that's not the case for `CacheSyncTimeout`.

Could we remove the `CacheSyncTimeout` from the reconcilers?

hedgieinsocks: oh, indeed, that was an accidental brainrot moment :)

Issue #1376, opened 2025-08-17
[Feature] allow loadBalancerClass attribute on Service


jr200:
**Is your feature request related to a problem? Please describe.**
Tailscale integration requires `loadBalancerClass` to be set to `tailscale` on the `Service` object..

**Describe the solution you'd like**
Update the service parameters to allow this field to be optionally set.

**Describe alternatives you've considered**
I tried editing the Service manually, but this results in an invalid mariadb object.

**Environment details**:
- Kubernetes version: 1.32.1
- Kubernetes distribution: k3s
- MariaDB Operator version: 25.8.3
- MariaDB Server version: 11.8.2
- MaxScale version: N/A
- Install method: helm

**Additional context**
Here is an example of a full yaml service I want to deploy.

```yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
    external-dns.alpha.kubernetes.io/hostname: mymariadb.example.com
    tailscale.com/hostname: mymariadb
  name: mymariadb
  namespace: mymariadb-namespace
spec:
  allocateLoadBalancerNodePorts: true
  clusterIP: 10.7.117.126
  clusterIPs:
    - 10.7.117.126
  externalTrafficPolicy: Cluster
  internalTrafficPolicy: Cluster
  ipFamilies:
    - IPv4
  ipFamilyPolicy: SingleStack
  loadBalancerClass: tailscale
  ports:
    - name: mariadb
      nodePort: 31204
      port: 3306
      protocol: TCP
      targetPort: 3306
  selector:
    app.kubernetes.io/instance: mymariadb
    app.kubernetes.io/name: mymariadb
  sessionAffinity: None
  type: LoadBalancer
```

pkalsi97:
@mmontes11 is this still up for grabs?

PR #1544, opened 2025-12-24
Add Service LoadBalancerClass

daudcanugerah:
changes:
- add load balancer class

fix: https://github.com/mariadb-operator/mariadb-operator/issues/1376

### DIFF
--- a/pkg/builder/service_builder.go
+++ b/pkg/builder/service_builder.go
@@ -55,11 +55,10 @@ func (b *Builder) BuildService(key types.NamespacedName, owner metav1.Object, op
 			return nil, fmt.Errorf("error validating Ports: %v", err)
 		}
 	}
-	objMeta :=
-		metadata.NewMetadataBuilder(key).
-			WithMetadata(opts.ExtraMeta).
-			WithMetadata(opts.Metadata).
-			Build()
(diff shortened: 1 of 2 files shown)


mmontes11 (changes requested):
Can you run make gen and commit the changes?

Also could you provide a test for the builder package to verify that the field gets correctly added to the Service?

Thanka for the contribution

daudcanugerah:
got it

pkg/builder/service_builder.go
```diff
@@ -74,6 +73,9 @@ func (b *Builder) BuildService(key types.NamespacedName, owner metav1.Object, op
 	if !opts.ExcludeSelectorLabels {
 		svc.Spec.Selector = opts.SelectorLabels
 	}
+	if opts.LoadBalancerClass != nil {
```
mmontes11, line 76: You should also be providing this additional option when invoking `BuildService`:
- https://github.com/mariadb-operator/mariadb-operator/blob/c3079efacf197c17a6ecce82acd162a0d25489b9/pkg/builder/service_builder.go#L49

mmontes11:
Addressed by https://github.com/mariadb-operator/mariadb-operator/issues/1376

PR #1554, opened 2025-12-30
feat(storage): add pvcRetentionPolicy field to configure PVC retention

### LINKED ISSUE #1347: [Feature] Allow to configure PersistentVolumeClaimRetentionPolicy

We use `mariadb-operator` to provision relatively short-lived (weeks to months) `MariaDB` instances. When deleting those instances the `PersistentVolumeClaim` created by the `StatefulSet` remains.  We understand that this is the desired behavior for most use-cases. However, for our use-case we would like to be able to configure whether the `PersistentVolumeClaim` gets deleted.

In Kubernetes 1.32 the [PersistentVolumeClaim retention][1] feature became stable and is enabled by default. The `mariadb-operator` already allows to configure some aspects of the `PersistentVolumeClaim` used for `MariaDB` via the [`PersistentVolumeClaimSpec` type][2]. We would therefore like to suggest to allow configuring `PersistentVolumeClaim` retention using this type. This should be as simple as adding a field `PersistentVolumeClaimRetentionPolicy` to `PersistentVolumeClaimSpec` and then using it during reconciliation.

If there is interest in such a feature we would be happy to provide a PR.

- Ephemeral storage. But we would like the data to be retained if the pod restarts.
- Deletion within our own automation. Would always be possible. But maybe this request valuable for other users as well.

**Environment details**:
- Kubernetes version: 1.33.1
- Kubernetes distribution: Talos
- MariaDB Operator version: 0.38.1
- MariaDB Server version: 11.4.5

[1]: https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#persistentvolumeclaim-retention
[2]: https://pkg.go.dev/github.com/mariadb-operator/mariadb-operator/api/v1alpha1#PersistentVolumeClaimSpec

majiayu000:
## Summary
This PR fixes #1347 by adding the ability to configure PersistentVolumeClaimRetentionPolicy for MariaDB StatefulSets.

## Changes
- Add `pvcRetentionPolicy` field to the `Storage` type in `api/v1alpha1/mariadb_types.go`
- Propagate the field to StatefulSet in `pkg/builder/statefulset_builder.go`
- Update CRD schemas with proper documentation for `whenDeleted` and `whenScaled` options
- Generate deepcopy functions for the new field

## Usage Example
```yaml
kind: MariaDB
metadata:
  name: mariadb
spec:
  storage:
    size: 10Gi
    pvcRetentionPolicy:
      whenDeleted: Delete
      whenScaled: Retain
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/make/docs.mk
+++ b/make/docs.mk
@@ -9,6 +9,7 @@ docs-api: crd-ref-docs ## Generate API reference docs.
 		--renderer=markdown \
 		--output-path=$(DOCS_API_REFERENCE) \
 		--max-depth=13
+        # Detect GNU vs BSD sed (sed --version only works in GNU)
 	@sed -i '/nolint:lll/d' $(DOCS_API_REFERENCE) 2>/dev/null || \
 	  sed -i '' '/nolint:lll/d' $(DOCS_API_REFERENCE)
 


api/v1alpha1/mariadb_types.go
```diff
@@ -91,6 +91,13 @@ type Storage struct {
 	// +optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:advanced"}
 	VolumeClaimTemplate *VolumeClaimTemplate `json:"volumeClaimTemplate,omitempty"`
+	// PersistentVolumeClaimRetentionPolicy describes the lifecycle of PVCs created from volumeClaimTemplates.
+	// By default, all persistent volume claims are created as needed and retained until manually deleted.
+	// This policy allows the lifecycle to be altered, for example by deleting PVCs when their statefulset is deleted,
+	// or when their pod is scaled down.
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	PVCRetentionPolicy *appsv1.StatefulSetPersistentVolumeClaimRetentionPolicy `json:"pvcRetentionPolicy,omitempty"`
```
mmontes11: We don't directly embed the Kubernetes types in our CRs, we create our own here pointing to the upstream documentation:
- https://github.com/mariadb-operator/mariadb-operator/blob/main/api/v1alpha1/kubernetes_types.go

We also have an adapter here to transform between them:
- https://github.com/mariadb-operator/mariadb-operator/tree/main/pkg/kubernetes/adapter

This is to avoid hitting the [1MB hard limit of etcd when installing helm releases](https://github.com/helm/helm/issues/1413)

mmontes11 (changes requested):
Thanks for the PR, left a comment.

Additionally, you need to run `make gen` to generate artifacts.

pkg/builder/statefulset_builder.go
```diff
@@ -91,10 +91,11 @@ func (b *Builder) BuildMariadbStatefulSet(mariadb *mariadbv1alpha1.MariaDB, key
...
 		Spec: appsv1.StatefulSetSpec{
-			ServiceName:         mariadb.InternalServiceKey().Name,
-			Replicas:            &mariadb.Spec.Replicas,
-			PodManagementPolicy: appsv1.ParallelPodManagement,
-			UpdateStrategy:      *updateStrategy,
+			ServiceName:                          mariadb.InternalServiceKey().Name,
+			Replicas:                             &mariadb.Spec.Replicas,
+			PodManagementPolicy:                  appsv1.ParallelPodManagement,
+			UpdateStrategy:                       *updateStrategy,
+			PersistentVolumeClaimRetentionPolicy: mariadb.Spec.Storage.PVCRetentionPolicy,
```
mmontes11: This has been introduced recently, it could potentially clash with older Kubernetes versions, resulting in the API server rejecting our `StatefulSet` resource due to an unknown field.

To avoid this, we need a discovery mechanism to check the Kubernetes version, and add the field if it is supported, see [this comment](https://github.com/mariadb-operator/mariadb-operator/issues/1347#issuecomment-3114032630).

We have users running in old Kubernetes vesions, we want to give them some time to migrate.

mmontes11 (changes requested):
One more request from my side.

majiayu000:
Hi, @mmontes11 , Thanks for the review! I’ve addressed the feedback:

- Avoided embedding Kubernetes types in the CR: added our own PVC retention policy types and
an adapter (`api/v1alpha1/kubernetes_statefulset_types.go`).
- Added discovery/version gating so the StatefulSet field is only set when supported (server
>= 1.32).
- Ran `make gen` and committed the generated artifacts (CRDs + docs), plus updated tests.

Please take another look.

api/v1alpha1/kubernetes_statefulset_types.go
```diff
@@ -0,0 +1,31 @@
+// nolint:lll
+package v1alpha1
+
+import appsv1 "k8s.io/api/apps/v1"
+
+// PersistentVolumeClaimRetentionPolicyType describes the lifecycle of persistent volume claims.
+// Refer to the Kubernetes docs: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#persistentvolumeclaimretentionpolicytype-v1-apps.
```
mmontes11:
```suggestion
// Refer to the Kubernetes docs: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#statefulsetpersistentvolumeclaimretentionpolicy-v1-apps.
```

hack/bump_version_examples.sh
```diff
@@ -11,9 +11,15 @@ fi
 
 echo "Updating examples in folder '$FOLDER' to '$IMAGE:$VERSION'"
 
+if sed --version >/dev/null 2>&1; then
```
mmontes11, line 15:
```suggestion
# Detect GNU vs BSD sed (sed --version only works in GNU)
if sed --version >/dev/null 2>&1; then
```

hack/kube_api_version.sh
```diff
@@ -10,10 +10,15 @@ fi
 
 echo "Updating API doc references to version: $KUBE_API_VERSION"
 ESCAPED_VERSION=$(echo "$KUBE_API_VERSION" | sed 's/\./\\./g')
+if sed --version >/dev/null 2>&1; then
```
mmontes11, line 14:
```suggestion
# Detect GNU vs BSD sed (sed --version only works in GNU)
if sed --version >/dev/null 2>&1; then
```

make/docs.mk
```diff
@@ -9,11 +9,12 @@ docs-api: crd-ref-docs ## Generate API reference docs.
 		--renderer=markdown \
 		--output-path=$(DOCS_API_REFERENCE) \
 		--max-depth=13
-	sed -i '/nolint:lll/d' $(DOCS_API_REFERENCE)
+	@sed -i '/nolint:lll/d' $(DOCS_API_REFERENCE) 2>/dev/null || \
```
mmontes11, line 13:
```suggestion
        # Detect GNU vs BSD sed (sed --version only works in GNU)
	@sed -i '/nolint:lll/d' $(DOCS_API_REFERENCE) 2>/dev/null || \
```

api/v1alpha1/mariadb_types.go
mmontes11: Addressed in https://github.com/mariadb-operator/mariadb-operator/pull/1554

PR #1565, opened 2026-01-05
perf: http.Transport created now has same defaults as http.DefaultTransport

Michaelpalacce:
This was causing goroutines to be created for each keep-alive idle connection. Now, they will timeout and will be pruned

### DIFF
--- a/pkg/http/client.go
+++ b/pkg/http/client.go
@@ -264,7 +264,7 @@ func (c *Client) getTransport(opts *Opts) (http.RoundTripper, error) {
 	}
 
 	// @PERF: We should not create a transport every time, but potentially pool them.
-	//        ref: net/http/transport.go > DefaultTransport
+	// Using defaults from DefaultTransport: https://pkg.go.dev/net/http#DefaultTransport
 	return &http.Transport{
 		Proxy: http.ProxyFromEnvironment,
 		TLSClientConfig: &tls.Config{


pkg/http/client.go
```diff
@@ -263,11 +263,18 @@ func (c *Client) getTransport(opts *Opts) (http.RoundTripper, error) {
 		return nil, fmt.Errorf("error parsing x509 keypair: %v", err)
 	}
 
+	// @PERF: We should not create a transport every time, but potentially pool them
```
mmontes11: Can we reference the `DefaultTransport` here for further reference about the defaults?

mmontes11 (changes requested):

pkg/http/client.go
Michaelpalacce: Done

pkg/http/client.go
```diff
@@ -263,11 +263,19 @@ func (c *Client) getTransport(opts *Opts) (http.RoundTripper, error) {
 		return nil, fmt.Errorf("error parsing x509 keypair: %v", err)
 	}
 
+	// @PERF: We should not create a transport every time, but potentially pool them.
+	//        ref: net/http/transport.go > DefaultTransport
```
mmontes11:
```suggestion
	// @PERF: We should not create a transport every time, but potentially pool them.
	// Using defaults from DefaultTransport: https://pkg.go.dev/net/http#DefaultTransport

PR #1567, opened 2026-01-05
make embed-entrypoint for 2025q4

grooverdan:


mmontes11:
Hey @grooverdan ! Thank you for looking into this.

I think the `make gen` is generating a `12.0`  directory, could you add it to the commit? If it was not intentional, or for some reason it shouldn't be added, could we discard it in `make gen`?

Thanks!

grooverdan:
> Hey @grooverdan ! Thank you for looking into this.
> 
> I think the `make gen` is generating a `12.0` directory, could you add it to the commit? If it was not intentional, or for some reason it shouldn't be added, could we discard it in `make gen`?
> 
> Thanks!

12.0 was discarded as its EOL, https://mariadb.org/about/#maintenance-policy, like all the other short term releases.

> could we discard it in `make gen`?
> 

ack, done. Should have read the output closer.

mmontes11 (approved):
LGTM!

Issue #1566, opened 2026-01-05
[Bug] Generate secret name for `rootPasswordSecretKeyRef` in helm chart instead of hardcoding it to `mariadb`


mikk150:
**Documentation**
- [x] I acknowledge that I have read the relevant [documentation](https://github.com/mariadb-operator/mariadb-operator/tree/main/docs).

**Describe the bug**
I would like to use `mariadb-cluster` helm chart in my chart as a dependency
I would like to be able to install multiple of these charts in a single namespace

**Expected behaviour**
Each release would generate their own root `Secret` with unique name and password for it (See #1265 )

**Steps to reproduce the bug**
1. install a release
2. install a new release
3. observe how they would use same secret for both releases (without actually even generating secret, but that is beyond the point)

**Debug information**
- Related object events:
```bash
kubectl get events --field-selector involvedObject.name=<mariadb-resource-name>
kubectl get events --field-selector involvedObject.name=<backup-resource-name>
kubectl get events --field-selector involvedObject.name=<restore-resource-name>
```
- `mariadb-operator` logs. Set the `--log-level` to `debug` if needed.

**Environment details**:
- Kubernetes version: 1.34.2
- Kubernetes distribution: K3S
- MariaDB Operator version: 25.10.2
- MariaDB topology: Galera
- MariaDB Server version: 11.8.2
- MaxScale version: [Version number]
- Install method: Helm

**Additional context**

mmontes11:
Hey there! 

> I would like to use mariadb-cluster helm chart in my chart as a dependency
I would like to be able to install multiple of these charts in a single namespace
Each release would generate their own root Secret with unique name and password for it 

All these requirements can be met by overriding the default values, for example:

instance1-values.yaml
```yaml
mariadb:
  rootPasswordSecretKeyRef:
    name: instance1
    key: root-password
    generate: true
  passwordSecretKeyRef:
    name: instance1
    key: root-password
    generate: true
```
instance2-values.yaml
```yaml
mariadb:
  rootPasswordSecretKeyRef:
    name: instance2
    key: root-password
    generate: true
  passwordSecretKeyRef:
    name: instance2
    key: root-password
    generate: true
```

```bash
helm upgrade --install instance1 mariadb-operator/mariadb-cluster -f instance1-values.yaml
helm upgrade --install instance2 mariadb-operator/mariadb-cluster -f instance2-values.yaml
```

[...]

PR #1572, opened 2026-01-07
Improve secref handling in mariadb-cluster chart

hedgieinsocks:
Fixes https://github.com/mariadb-operator/mariadb-operator/issues/1566

**Note:** this adds `generate: true` to the root secret reference to simplify testing for the newcomers. Thus, this change should be mentioned in the upgrade notes to allow old consumers to set it to false.

### DIFF
--- a/deploy/charts/mariadb-cluster/README.md
+++ b/deploy/charts/mariadb-cluster/README.md
@@ -32,7 +32,8 @@ Refer to the [helm documentation](https://github.com/mariadb-operator/mariadb-op
 | databases | list | `[]` | The list of Database CRs. The `.mariaDbRef` keys will be ignored. The `.name` keys are required to generate distinct CR names. The `.namespace` keys are allowed to override `.Release.Namespace`. https://github.com/mariadb-operator/mariadb-operator/blob/main/docs/api_reference.md#databasespec |
 | fullnameOverride | string | `""` |  |
 | grants | list | `[]` | The list of Grant CRs. The `.mariaDbRef` keys will be ignored. The `.name` keys are used to generate distinct CR names. The `.namespace` keys are allowed to override `.Release.Namespace`. https://github.com/mariadb-operator/mariadb-operator/blob/main/docs/api_reference.md#grantspec |
-| mariadb | object | `{"galera":{"enabled":true},"replicas":3,"rootPasswordSecretKeyRef":{"key":"root-password","name":"mariadb"},"storage":{"size":"1Gi"}}` | MariaDB CR https://github.com/mariadb-operator/mariadb-operator/blob/main/docs/api_reference.md#mariadbspec |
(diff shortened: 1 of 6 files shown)


mikk150:
it would be helpful to have a method to get a secret name for certain secretref

hedgieinsocks:
@mmontes11 sure thing, will do later today! 
@mikk150 could you elaborate a bit on your notion? I could still try to squeeze something useful here if need be.

mikk150:
> @mikk150 could you elaborate a bit on your notion? I could still try to squeeze something useful here if need be.

If I develop my own helm chart, and `mariadb-cluster` chart is a dependency for my chart, it would be really helpful if something like this could be done:
```
            - name: DB_HOST
              value: {{ include "mariadb-cluster.databaseHost" $ | quote }}
            - name: DB_PORT
              value: {{ include "mariadb-cluster.databasePort" $ | quote }}
            - name: DB_NAME
              value: {{ include "mariadb-cluster.databaseName" $ | quote }}
            - name: DB_USER
              value: {{ include "mariadb-cluster.databaseUser" $ | quote }}
            - name: DB_PASSWORD
              valueFrom:
                secretKeyRef:
                  name: {{ include "mariadb-cluster.secretKeyRefName" (dict "global" $ "secretKeyRef" $.mariadb-cluster.mariadb.passwordSecretKeyRef ) }}
                  key: {{ include "mariadb-cluster.secretKeyRefName" (dict "global" $ "secretKeyRef" $.mariadb-cluster.mariadb.passwordSecretKeyRef ) }}
```

hedgieinsocks:
```yaml
# example of quick dirty local dep declaration with alias in parent Chart.yaml
dependencies:
  - name: mariadb-cluster
    version: 25.10.3
    repository: "file://charts/mariadb-cluster"
    alias: mdb

# example of deployment.yaml in parent
env:
  - name: DB_PASSWORD
    valueFrom:
      secretKeyRef:
        name: {{ .Values.mdb.mariadb.rootPasswordSecretKeyRef.name }}
        key: {{ .Values.mdb.mariadb.rootPasswordSecretKeyRef.key }}
```
```sh
# template parent
❯ helm template foo .
---
# Source: foo/templates/deployment.yaml
env:
  - name: DB_PASSWORD
    valueFrom:
      secretKeyRef:
        name: foo-mdb-root
        key: root
---
# Source: foo/charts/mdb/templates/mariadb.yaml
kind: MariaDB
metadata:
  name: foo-mdb
  namespace: default
  labels:
    helm.sh/chart: mdb-25.10.3
    app.kubernetes.io/name: mdb
    app.kubernetes.io/instance: foo
    app.kubernetes.io/version: "0.0.0"
    app.kubernetes.io/managed-by: Helm
spec:
  galera:
    enabled: true
  replicas: 3
  rootPasswordSecretKeyRef:
    generate: true
    key: root
    name: foo-mdb-root
  storage:
    size: 1Gi
```

@mikk150 If I'm not raving, it looks like with this pr, the parent template picks up the value we set in the mariadb-cluster template, so you don't really need a function getter for that.

PR #1588, opened 2026-01-20
fix: use standard compression extensions and add magic bytes validation

### LINKED ISSUE #940: [Bug] Compressed backups use unconventional file extensions
**Documentation**

**Describe the bug**

Gzip compressed backups have the file extensions in the wrong order. They are named:

`backup.2024-10-13T02_00_00Z.gzip.sql`

Which means you cannot just decompress them, since gzip does not recognize the .sql extension:

```
$ gunzip backup.2024-10-13T02_00_00Z.gzip.sql      
gzip: backup.2024-10-13T02_00_00Z.gzip.sql: unknown suffix -- ignored
```

It would be better if the files are called `*.sql.gz` so they can be recognized by tools like gunzip by default.
 
**Expected behaviour**

1. Create a gzip compressed backup.
2. Check the created filename.

**Environment details**:
- Kubernetes version: 1.30.5+k3s1
- Kubernetes distribution: k3s
- mariadb-operator version: 0.34.0
- Galera data-plane version (init and agent): N/A
- Install method: helm

**Additional context**

yangminglintw:
## Summary
- Changes backup filename format from `backup.{timestamp}.{algo}.sql` to `backup.{timestamp}.sql.{ext}` (e.g., `.sql.gz` instead of `.gzip.sql`)
- Adds magic bytes validation during decompression to detect file format by content rather than relying solely on file extension:
  - gzip: validates `0x1f 0x8b` header
  - bzip2: validates `0x42 0x5a` ("BZ") header
- Maintains backward compatibility for reading legacy format files

## Changes
| File | Description |
|------|-------------|
| `pkg/command/backup.go` | Updated `newBackupFile()` to use standard extension format |
| `pkg/backup/processor.go` | Updated `ParseCompressionAlgorithm()` and `GetUncompressedBackupFile()` to support both new and legacy formats |
| `pkg/backup/processor_test.go` | Added test cases for new and legacy formats |
| `pkg/compression/compressor.go` | Added magic bytes constants and `validateMagicBytes()` function |
| `pkg/compression/compressor_test.go` | Added magic bytes validation tests |

## Test plan
- [x] Unit tests pass (`go test ./pkg/backup/... ./pkg/compression/... ./pkg/command/...`)
- [x] Lint passes (`make lint`)
- [x] Integration tests (CI will run)

Fixes #940

🤖 Generated with [Claude Code](https://claude.ai/code)


pkg/compression/compressor.go
```diff
@@ -13,6 +13,13 @@ import (
 	mariadbv1alpha1 "github.com/mariadb-operator/mariadb-operator/v25/api/v1alpha1"
 )
 
+// Magic bytes for compression format detection.
+// These are used to validate file content matches the expected compression algorithm.
+var (
```
mmontes11: The main point of introducing the magic bytes manipulation was to infer the compression type, but, we are doing that based on the extension, which is fine and now it is backward compatible, great!

However, I don't think we should introduce additional complexity to enforce the magic bytes, if something is wrong with the compression encoding, the gzip/bzip2 library will return an error.

Could we omit the magic bytes additional verification?

yangminglintw: Done! Removed the magic bytes validation. Thanks for the feedback - good lesson on keeping things simple.

leleobhz: @mmontes11 just 2 cents here about Magic: GZIP does handle extensions just for human conventions, as for example explained at https://cgit.git.savannah.gnu.org/cgit/gzip.git/tree/gzip.c#n1115 but for real file processing, magic is used instead: https://cgit.git.savannah.gnu.org/cgit/gzip.git/tree/gzip.c#n1435

The reason this is not overengineering is because magic is the way most of Unix software and culture relies. Linux (And other unices) tools rarely rely on extension for anything except for convention reasons. Using magic was my suggestion because magic usage allows the operator to just dont need to handle filename - keeping the old and new format compatibily more universal than start a filename parse list. In fact my suggestion was to do not handle filenames to restore or validate a backup: Rely on magic to read and create as the desired standard. 

About something wrong with compression encoding will fail gzip/bzip2/anycompresssystem anyways because data handling itself is based on magic.

PR #1601, opened 2026-01-26
Make volumes and volumeMounts mutable

### LINKED ISSUE #1587: [Bug] Volumes and VolumeMounts are immutable
I use openebs-lvmpv nvme-backed storage class for mariadb storage pvc. But the root partition on the node is a simple hard drive. So whenever mariadb writes to `/tmp` (when tmp table exceeds the configured buffer), it writes to the hard drive. I was thinking of doing something like below to avoid tweaking tmp folder location or raising buffers. 

```yaml
  volumes:
    - name: tmp
      ephemeral:
        volumeClaimTemplate:
          spec:
            storageClassName: openebs-lvmpv
            resources:
              requests:
                storage: 20Gi
            accessModes:
              - ReadWriteOnce

  volumeMounts:
    - name: tmp
      mountPath: /tmp
```

But it turned out that in the current API volumes and volumeMounts are immutable. 

Is it by design and there is a reason for that? Or is it some remnant from the past, because I remember some other fields were immutable without any reason for that, and since then were fixed.

hedgieinsocks:
Fixes https://github.com/mariadb-operator/mariadb-operator/issues/1587

### DIFF
--- a/api/v1alpha1/base_types.go
+++ b/api/v1alpha1/base_types.go
@@ -195,7 +195,7 @@ type Container struct {
 	// VolumeMounts to be used in the Container.
 	// +optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:advanced"}
-	VolumeMounts []VolumeMount `json:"volumeMounts,omitempty" webhook:"inmutable"`
+	VolumeMounts []VolumeMount `json:"volumeMounts,omitempty"`
 	// Resources describes the compute resource requirements.
 	// +optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:resourceRequirements"}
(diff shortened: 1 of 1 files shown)


mmontes11 (approved):

mmontes11:
Thanks for the contribution @hedgieinsocks !

hedgieinsocks:
@mmontes11 fyi it looks like I missed https://github.com/mariadb-operator/mariadb-operator/blob/6776f6b8940fa9cb5807b12ac6366619e251a207/api/v1alpha1/base_types.go#L116

mmontes11:
@hedgieinsocks  Added as [TODO for 26.06](https://github.com/mariadb-operator/mariadb-operator/pull/1715), planning to relese this week. Any other minor thing that we left behind?

hedgieinsocks:
Among minor things, you can check https://github.com/mariadb-operator/mariadb-operator/pull/1676 
Context: I run a daily per-db backup cron against the replication replica, and mysqldump command appears to cause some related transactions for in-memory tables (e.g. wordfence wordpress plugin), so the replica ends up with binary log entries that clean up in-memory tables during backup. And that causes failover to stumble if that replica is promoted.

mmontes11:
Thanks for flagging this. This has to be considered more carefully, as it has true implications for PITR and multi-cluster. Besides, we hae fully refactored that part as part of multi-cluster (now abastracted in `topology.go`), so it needs to be integrated. This will be evaluated for further released.

PR #1617, opened 2026-02-20
Test: Add Integration tests for PITR

Michaelpalacce:
This still may need more work, need to switch focus

### DIFF
--- a/internal/controller/mariadb_controller_pitr_test.go
+++ b/internal/controller/mariadb_controller_pitr_test.go
@@ -61,28 +61,8 @@ var _ = Describe("MariaDB PITR with Replication", Ordered, func() {
 			By("Expecting PhysicalBackup to be ready")
 			expectPhysicalBackupReady(physicalBackup)
 		})
-
-	})
-
-	PContext("With ABS Storage", func() {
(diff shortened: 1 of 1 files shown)


internal/controller/mariadb_controller_pitr.go
```diff
@@ -368,66 +368,13 @@ func (r *MariaDBReconciler) cleanupPITRJob(ctx context.Context, mariadb *mariadb
 }
 
 func (r *MariaDBReconciler) getS3Client(ctx context.Context, pitr *mariadbv1alpha1.PointInTimeRecovery) (*minio.Client, error) {
```
Michaelpalacce, line 432: I'm leaving this enhancement as I needed it for tests but I think it may be useful in the future

mmontes11:
@Michaelpalacce will take a look later. One observation: `test/feature-pitr-int-tests` does not trigger the full integration tests

Michaelpalacce:
> @Michaelpalacce will take a look later. One observation: `test/feature-pitr-int-tests` does not trigger the full integration tests

Hmm, I will run them all manually and post the results, I could close the PR, create a new branch with a different name if you want?

PR #1618, opened 2026-02-20
Add EphemeralVolumeSource to volume

hedgieinsocks:
Related to https://github.com/mariadb-operator/mariadb-operator/issues/1587

While https://github.com/mariadb-operator/mariadb-operator/pull/1601 made volumes mutable, I still cannot use `ephemeral` since it must be explicitly added to the codebase.

This pr is supposed to fix this.

### DIFF
--- a/api/v1alpha1/kubernetes_volume_types.go
+++ b/api/v1alpha1/kubernetes_volume_types.go
@@ -99,13 +99,22 @@ func (v PersistentVolumeClaimVolumeSource) ToKubernetesType() corev1.PersistentV
 
 // Refer to the Kubernetes docs: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#ephemeralvolumesource-v1-core.
 type EphemeralVolumeSource struct {
-	VolumeClaimTemplate *corev1.PersistentVolumeClaimTemplate `json:"volumeClaimTemplate,omitempty"`
+	VolumeClaimTemplate *VolumeClaimTemplate `json:"volumeClaimTemplate,omitempty"`
 }
 
 func (v EphemeralVolumeSource) ToKubernetesType() corev1.EphemeralVolumeSource {
-	return corev1.EphemeralVolumeSource{
-		VolumeClaimTemplate: v.VolumeClaimTemplate,
+	var ephemeral corev1.EphemeralVolumeSource
+	if v.VolumeClaimTemplate != nil {
+		meta := ptr.Deref(v.VolumeClaimTemplate.Metadata, Metadata{})
+		ephemeral.VolumeClaimTemplate = &corev1.PersistentVolumeClaimTemplate{
+			ObjectMeta: metav1.ObjectMeta{
+				Labels:      meta.Labels,
+				Annotations: meta.Annotations,
+			},
+			Spec: v.VolumeClaimTemplate.ToKubernetesType(),
+		}
 	}
+	return ephemeral
 }
 
 // Refer to the Kubernetes docs: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#secretvolumesource-v1-core.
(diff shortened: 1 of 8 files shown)


api/v1alpha1/kubernetes_volume_types.go
```diff
@@ -97,6 +97,17 @@ func (v PersistentVolumeClaimVolumeSource) ToKubernetesType() corev1.PersistentV
 	}
 }
 
+// Refer to the Kubernetes docs: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#ephemeralvolumesource-v1-core.
+type EphemeralVolumeSource struct {
+	VolumeClaimTemplate *corev1.PersistentVolumeClaimTemplate `json:"volumeClaimTemplate,omitempty"`
```
mmontes11: We should be using our own [VolumeClaimTemplate](https://github.com/mariadb-operator/mariadb-operator/blob/7df3be4b889d0351d60fefc0003964dcaaad9c08/api/v1alpha1/base_types.go#L598) here instead of corev1.

The reason is not htting the helm [1MB hard limit ](https://azure.github.io/azure-service-operator/design/adr-2023-02-helm-chart-size-limitations/), which would affect the CRD installation.

This change in particular has bumped the size from [706KB](https://github.com/mariadb-operator/mariadb-operator/actions/runs/22804612914/job/66151637310) to [857KB](https://github.com/mariadb-operator/mariadb-operator/actions/runs/22222128117/job/64279869655?pr=1618).

hedgieinsocks: oh, I see, even with our own it still adds 64KB
```sh
❯ make crd-size
Current CRD size: 770 KB
```

mmontes11: Yeah, still not ideal. Given that we are only interested in `MariaDB` in this context, I think it will be worth creating a new dedicated `MariaDBStorageVolumeSource` and use it only in `MariaDB`. `StorageVolumeSource` is reused across many CRDs and it is increasing the size a lot. What do you think?

`MariaDBStorageVolumeSource`  would contain the new ephemeral field, `StorageVolumeSource` won't, as it does not have a use case in the rest of the CRDs

hedgieinsocks: hm, not sure to be honest :thinking:  Should I try to just substitute `StorageVolumeSource` for `MariaDBStorageVolumeSource` in `VolumeSource` and call it a day since it seems to be used only there (gotta double-check)? Or also rename things e.g. `VolumeSource` to  `MariaDBVolumeSource` & `Volume` to `MariaDBVolume` etc? Or create a new `MariaDBVolume` etc. next to `Volume`? And where to house it then, in mariadb types file or k8s one?

mmontes11: Sorry, this is a bit trickier than I initially pictured, given the high degree of reustability and composition. Will book some time to work on this before releasing 26.03 this week. I will probably port your changes to a PR and add you as a co-author.

Right now, the only solution I'm thinking of creating dedicated types to `MariaDB` all the way down: `MariaDBPodTemplate` > `MariaDBVolume` > `MariaDBVolumeSource` (including the new `ephemeral`) . This would fix the size issue, give us room to grow CRDs, but will also make the maintenance worse, as we are duplicating types. 

FYI @Michaelpalacce in case you have any other ideas.

hedgieinsocks: It's nice that this can be solved but given how awkward it would be, I don't think we should go all in yet just to squeeze it into the next release. It's not urgent after all. We can try to tackle it later when inspiration or need kicks in.

mmontes11:
Ported changes and closed in favor of:
https://github.com/mariadb-operator/mariadb-operator/pull/1650

PR #1621, opened 2026-02-23
PITR support for Galera

mmontes11:
⚠️ Important considerations⚠️:
- Galera archivals will only happen in Pod 0, see [this comment](https://github.com/mariadb-operator/mariadb-operator/pull/1621#issuecomment-3950874920).

ToDos:

- [ ] Take a look at [this](https://github.com/mariadb-operator/mariadb-operator/pull/1694#discussion_r3039566448)


mmontes11:
@Michaelpalacce  Found a gotcha in the PITR Galera support: although all nodes have the same server Id and same GTID, they do not have the same binary logs. My plan is to archive the binlogs, not in the `currentPrimary`,  but in a Pod optionally defined by the user via `podArchiverIndex`. If not set, will default to 0. Archiver will only perform archival when located at `podArchiverIndex`, this will apply to both Galera and standalone topologies. `podArchiverIndex` will be immutable.

The issue arised when performing a primary failover mariadb-galera-0 -> mariadb-galera-1. The new primary (mariadb-galera-1) although it had all the tx, was still in mariadb-galera-bin.00001, but the previous primary (mariadb-galera-0) had already archived up until mariadb-galera-bin.00003 after performing a few FLUSH BINARY LOGS, so the archival process was blocked in mariadb-galera-1.

api/v1alpha1/pointintimerecovery_types.go
```diff
@@ -101,3 +100,7 @@ func (s *PointInTimeRecoveryStorage) Validate() error {
 
 	return nil
 }
+
+func init() {
```
Michaelpalacce: This is odd?

mmontes11:
@Michaelpalacce Could you solve conflicts and rebase with main?

PR #1627, opened 2026-03-02
Galera recovery: disable bootstrap on other pods before bootstrapping

infocusmodereal:
## What
During Galera recovery, ensure bootstrap is disabled on all **non-bootstrap** pods before enabling bootstrap on the chosen bootstrap pod.

## Why
Bootstrap is controlled by a persisted config file (`1-bootstrap.cnf` with `wsrep_new_cluster="ON"`) stored on the Galera config volume (PVC). If the operator enables bootstrap on the chosen pod and then fails to disable bootstrap on another pod (e.g. agent timeout / context deadline), there is a window where more than one pod can start with `wsrep_new_cluster=ON` (stale bootstrap), which can lead to multiple components / split brain.

This change makes recovery fail-safe: if any non-bootstrap pod cannot be reached / bootstrap cannot be cleared, the operator will return an error **before** enabling bootstrap anywhere.

## How
- Add a pre-flight loop in `restartPods()` to call `disableBootstrapInPod()` for all non-bootstrap pods.
- Then proceed with the existing bootstrap + sync flow.

## Testing
- `go test ./pkg/controller/galera`
- `make gen`

### DIFF
--- a/pkg/controller/galera/recovery.go
+++ b/pkg/controller/galera/recovery.go
@@ -175,36 +175,58 @@ func (r *GaleraReconciler) restartPods(ctx context.Context, mariadb *mariadbv1al
 		})
 	}
 
+	logger.Info("Ensuring bootstrap disabled in non-bootstrap Pods")
 	for _, podKey := range podKeys {
-		podLogger := logger.WithValues("pod", podKey.Name)
-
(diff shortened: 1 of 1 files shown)


infocusmodereal:
Superseded by #1631 .

PR #1635, opened 2026-03-04
Disable service links in MariaDB and MaxScale pod specs

### LINKED ISSUE #1634: Disable enableServiceLinks in pod specs to prevent env var explosion
## Problem

When running many MariaDB instances in a single namespace, Kubernetes injects service discovery environment variables for **every** Service into **every** pod. Each MariaDB instance creates primary, secondary, and metrics services, so the number of injected env vars grows quadratically with instance count.

At scale this causes hundreds of unnecessary environment variables per pod (e.g. `{INSTANCE}_PRIMARY_PORT_TCP`, `{INSTANCE}_SECONDARY_SERVICE_HOST`, etc.), wasting memory and making debugging harder.

### This is a hard limit, not just a nicety

At ~2000 services in a namespace, pods begin failing to start entirely with:

```
standard_init_linux.go:211: exec user process caused "argument list too long"
```

This happens because the combined size of all environment variables exceeds the Linux `ARG_MAX` limit for `execve()`. The kubelet injects ~3 env vars per service, so the ceiling is hit sooner than you might expect. This is a well-documented production failure mode:

- [kubernetes/kubernetes#84539](https://github.com/kubernetes/kubernetes/issues/84539)
- [kubernetes/kubernetes#121787 - proposal to make enableServiceLinks: false the default](https://github.com/kubernetes/kubernetes/issues/121787)
- [kubernetes/ingress-nginx#7968](https://github.com/kubernetes/ingress-nginx/issues/7968)
- [cloudnative-pg/cloudnative-pg#6234](https://github.com/cloudnative-pg/cloudnative-pg/issues/6234)

For operators managing many instances in a shared namespace, this is not theoretical -- it is a scaling wall.

### No user-facing workaround exists

The `PodTemplate` in the MariaDB CRD does not expose `enableServiceLinks`, and unlike some other operators (e.g. CNPG with its `cnpg.io/podPatch` annotation), there is no pod patching mechanism available. Manually patching the StatefulSet is not viable because the operator reconciles it back. There is currently no way for users to work around this.

## Fix

Set `EnableServiceLinks: ptr.To(false)` in the PodSpec construction in `pkg/builder/pod_builder.go`:

```go
Spec: corev1.PodSpec{
    AutomountServiceAccountToken: ptr.To(false),
    EnableServiceLinks:           ptr.To(false),  // <-- add this
```

This should be applied in both `mariadbPodTemplate` and `maxscalePodTemplate`.

MariaDB pods use direct service references and connection strings, not service link env vars, so there is no functional impact. This matches the direction Kubernetes itself is heading (kubernetes/kubernetes#121787).

usiegj00:
Fixes #1634

Sets `EnableServiceLinks: ptr.To(false)` in both `mariadbPodTemplate` and `maxscalePodTemplate` in `pkg/builder/pod_builder.go`.

### Why

Kubernetes injects service discovery env vars for every Service in a namespace into every pod. Each MariaDB instance creates multiple Services (primary, secondary, metrics), so in namespaces with many instances, pods accumulate hundreds of unnecessary env vars. At ~2000 services this hits the Linux `ARG_MAX` limit and pods fail to start with `exec user process caused "argument list too long"`.

MariaDB pods use direct service references and connection strings, not service link env vars. This is a two-line change that follows the same pattern already used for `AutomountServiceAccountToken`.

### References

- kubernetes/kubernetes#84539
- kubernetes/kubernetes#121787 (proposal to make `enableServiceLinks: false` the default)
- kubernetes/ingress-nginx#7968
- cloudnative-pg/cloudnative-pg#6234


pkg/builder/pod_builder.go
```diff
@@ -222,6 +222,7 @@ func (b *Builder) mariadbPodTemplate(mariadb *mariadbv1alpha1.MariaDB, opts ...m
 		ObjectMeta: objMeta,
 		Spec: corev1.PodSpec{
 			AutomountServiceAccountToken: ptr.To(false),
+			EnableServiceLinks:           ptr.To(false),
```
mmontes11: Although I completely agree with the reasons behind this PR, this would trigger an upgrade in all the `MariaDB` and `MaxScale` instances managed by the operator. Even if we add a notice in the release notes there will be users that upgrades without reading them, and we are aiming to be as stable as possible at this stage of the project.

For this reason I suggest we add `EnableServiceLinks` to both the MariaDB and MaxScale Pod templates, and map if here to the PodSpec:

```go
EnableServiceLinks: mariadb.Spec.EnableServiceLinks
```

usiegj00: Done — added `EnableServiceLinks` to both `PodTemplate` and `MaxScalePodTemplate`, mapped through to the PodSpec. No hardcoded default, so existing instances are unaffected.

PR #1606, opened 2026-02-04
Fix finalizer for ExternalMariDB

snaax:
Hello,

We spotted a problem using ExternalMariaDB. The finalizer doesn't trigger the cleanup of all CRDs (user/grants/database).

It seems it's coming from the fact that the finalizer is taking a MariaDB reference instead of an MariaDBObject reference (which handle both)

I think it's linked to this [issue](https://github.com/mariadb-operator/mariadb-operator/issues/1516)

Thanks

PR #1649, opened 2026-03-09
test: Added int tests for sql resources with external mariadb deletion

Michaelpalacce:
This is a continuation of PR #1606 to add INT tests


snaax:
Thank you very much @Michaelpalacce you should have told me in the other PR, I would have try to complete my duty :sweat_smile:

PR #1650, opened 2026-03-09
Support for ephemeral volumes in `MariaDB`

mmontes11:
Related to https://github.com/mariadb-operator/mariadb-operator/issues/1587

Related discussion: 
- https://github.com/mariadb-operator/mariadb-operator/pull/1618#discussion_r2900244284

As described in https://github.com/mariadb-operator/mariadb-operator/issues/1587, we are aiming to support  [generic ephemeral volumes](https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#generic-ephemeral-volumes) tied to the `Pod` lifecycle. For example , in https://github.com/mariadb-operator/mariadb-operator/issues/1587, the `/tmp` is mapped to a PVC that is tied to the `Pod` lifecycle. 

In [this discussion](https://github.com/mariadb-operator/mariadb-operator/pull/1618#discussion_r2900244284), an issue related to the CRD size arised, which has been resolved by using `MariaDBVolume` (with ephemeral support) instead of `Volume` in the `PodTemplate` struct. Turns out that `PodTemplate` was only being used by `MariaDB`, so it has been renamed to `MariaDBPodTemplate` and moved to `mariadb_types.go`, similarly to  [`MaxScalePodTemplate`](https://github.com/mariadb-operator/mariadb-operator/blob/7d126c32b9052d9fef289ce930a06e197229cc2b/api/v1alpha1/maxscale_types.go#L569).

The CRD after supporting ephemeral volumes is:

```bash
❯ make crd-size
Current CRD size: 714 KB
```

### DIFF
--- a/api/v1alpha1/mariadb_types.go
+++ b/api/v1alpha1/mariadb_types.go
@@ -485,7 +485,7 @@ func (v MariaDBVolume) ToKubernetesType() corev1.Volume {
 	}
 }
 
-// MariaDBPodTemplate defines a template to configure Container objects.
+// MariaDBPodTemplate defines a template for MariaDB Pods.
 type MariaDBPodTemplate struct {
 	// PodMetadata defines extra metadata for the Pod.
 	// +optional
--- a/api/v1alpha1/mariadb_types_test.go
+++ b/api/v1alpha1/mariadb_types_test.go
@@ -2321,3 +2321,53 @@ var _ = Describe("MariaDB types", func() {
 		)
 	})
 })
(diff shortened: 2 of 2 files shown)


mmontes11:
@hedgieinsocks, could you take a look?

hedgieinsocks:
lgtm, thanks for putting it all together!

mmontes11:
no prob @hedgieinsocks 🤝 , this will be released in 26.03 this week

PR #1651, opened 2026-03-10
feat: Latest controller-runtime. Webhook changes, New Events API

Michaelpalacce:
- Changes to the events recorders were needed because the linter was failing on deprecated methods. That required modifications to all the Events that were emitted.
  - Intentionally the only action for now is `Reconciling` to keep the review simple. Actions can be useful but from what I can tell they are purely for automation and machines to process
- The webhook changes were necessary due to breaking changes in the controller-runtime


internal/webhook/v1alpha1/backup_webhook.go
```diff
@@ -30,39 +28,27 @@ func SetupBackupWebhookWithManager(mgr ctrl.Manager) error {
 // when it is created, updated, or deleted.
 type BackupCustomValidator struct{}
 
-var _ webhook.CustomValidator = &BackupCustomValidator{}
+var _ admission.Validator[*v1alpha1.Backup] = &BackupCustomValidator{}
```
mmontes11, line 31: Generics, nice, but also the 2nd time they break the webhooks in the last year 😄

PR #1655, opened 2026-03-12
fix: accept compressed backup files in logical restore

voron:
## Summary
- `LogicalBackupProcessor.IsValidBackupFile()` rejected new-format compressed backups (`.sql.gz`, `.sql.bz2`) because it required files to end with `.sql`
- This was a regression from commit 4f538471 which changed compression format from `backup.{ts}.{algo}.sql` to `backup.{ts}.sql.{ext}`
- Changed `HasSuffix` to `Contains` so both compressed and uncompressed files pass the initial filter; the subsequent `ParseCompressionAlgorithm` and `parseDateInBackupFile` checks validate the full structure

## Test plan
- [x] Added test cases for new-format gzip (`.sql.gz`) and bzip2 (`.sql.bz2`) compressed files
- [x] All existing tests pass (31 suites)
- [x] Build succeeds

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/backup/processor.go
+++ b/pkg/backup/processor.go
@@ -83,7 +83,9 @@ func (p *LogicalBackupProcessor) GetOldBackupFiles(backupFileNames []string, max
 
 // IsValidBackupFile determines whether a backup file name is valid.
 func (p *LogicalBackupProcessor) IsValidBackupFile(fileName string) bool {
-	if !strings.HasPrefix(fileName, "backup.") || !strings.HasSuffix(fileName, ".sql") {
+	// Must start with "backup." and contain ".sql" either as suffix (uncompressed)
+	// or before the compression extension (e.g. ".sql.gz", ".sql.bz2").
+	if !strings.HasPrefix(fileName, "backup.") || !strings.Contains(fileName, ".sql") {
 		return false
 	}
 	_, err := p.ParseCompressionAlgorithm(fileName)
--- a/pkg/backup/processor_test.go
+++ b/pkg/backup/processor_test.go
@@ -330,8 +330,18 @@ func TestLogicalIsValidBackupFile(t *testing.T) {
 			wantValid:  true,
 		},
 		{
-			name:       "valid with compression",
+			name:       "valid with legacy compression",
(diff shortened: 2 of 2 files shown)


voron:
Hello,

> LGTM. Great catch! 🎖️

We hit it in dev env restore tests.

>May I ask you: How Claude Code is able to detect regressions even before releasing this code?

I suppose it analysed the whole repo code and commit history in local repo. Here is what it says:
```markdown
Great question! Here's how I traced the regression:

1. **Starting from the symptom**: You described "compressed logical backup restores fail name validation." That pointed directly at validation logic for backup file names.

2. **Reading `IsValidBackupFile`** in `pkg/backup/processor.go:86` — the `HasSuffix(fileName, ".sql")` check was immediately suspicious since compressed files end with `.gz` or `.bz2`, not `.sql`.

3. **Confirming via git history**: The commit `4f538471` ("fix: use standard compression extensions") changed the file naming convention from `backup.{ts}.{algo}.sql` (legacy, ends with `.sql`) to `backup.{ts}.sql.{ext}` (new, ends with `.gz`/`.bz2`). But `IsValidBackupFile` was **never updated** to match the new format.

4. **Verifying through the call chain**: `IsValidBackupFile` is called in `storage.go` during `List()` — both filesystem and S3 storage use it to **filter** which files are valid backups. Compressed files in the new format were silently filtered out, so the restore process saw zero valid backup files and failed.

5. **Cross-checking the tests**: The existing test suite only had a case for the **legacy** compressed format (`backup.2023-12-18T16:14:00Z.bzip2.sql`) — no test covered the new format. That's why the regression wasn't caught at the time of the original commit.

So the detection was essentially: **symptom → suspicious validation code → git blame to confirm the introducing commit → call chain analysis to confirm impact → test gap that explains why it slipped through**.
```

mmontes11:
@voron Thanks for sharing it! This slipped through our tests, as we don't have an integration test + compression + logical backup.

Luckily enough, you catch it just before release, I was literally writing the release notes:
https://github.com/mariadb-operator/mariadb-operator/pull/1517

Issue #1628, opened 2026-03-02
[Bug] Primary failover fails if primary is killed


lassebm:
**Documentation**
- [x] I acknowledge that I have read the relevant [documentation](https://github.com/mariadb-operator/mariadb-operator/tree/main/docs).

**Describe the bug**
If the node running the primary is killed, the primary failover fails due to unhealthy replicas. The replica readiness endpoint responds with `500 Internal Server Error: {"message":"could not determine replica lag"}`.

**Expected behaviour**
While the replicas has lost their primary, they should still be valid candidates for primary failover. This common scenario should be covered with default settings given all components are deployed with high availability.

**Steps to reproduce the bug**
1. Deploy kind cluster with 3 worker nodes
2. Deploy mariadb-operator with high availability on operator, webhook and cert-controller (node anti-affinity)
3. Deploy MariaDB instance with 2 replicas (node anti-affinity)
4. Identify node with current primary
5. Stop node with `docker stop kind-worker<x>`
6. Monitor primary failover

- Readiness probe 30s (default): Both replicas become unhealthy before the primary (50s). No primary candidates found when primary failover is started.
- Readiness probe 100s (failureThreshold=10): Primary become unhealthy (50s) before replicas. A new primary candidate is found and primary failover succeeds before replicas become unhealthy.

The primary failover is further delayed if the primary mariadb-operator pods are killed together with the primary, i.e. `failureThreshold` must be increased even more.

```bash
kind create cluster --config - <<EOF
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
- role: worker
- role: worker
- role: worker
EOF

helm repo add mariadb-operator https://helm.mariadb.com/mariadb-operator
helm install mariadb-operator-crds mariadb-operator/mariadb-operator-crds
helm upgrade mariadb-operator mariadb-operator/mariadb-operator --install --reset-values -f - <<EOF
logLevel: DEBUG
ha:
  enabled: true
  replicas: 2
affinity:
  podAntiAffinity:
    requiredDuringSchedulingIgnoredDuringExecution:
      - labelSelector:
          matchExpressions:
            - key: app.kubernetes.io/name
              operator: In
              values:
                - mariadb-operator
            - key: app.kubernetes.io/instance
              operator: In
              values:
                - mariadb-operator
        topologyKey: kubernetes.io/hostname
pdb:
  enabled: true
  maxUnavailable: 1
webhook:
  ha:
[...]

PR #1658, opened 2026-03-13
[WIP] Improve readiness & failover handling

hedgieinsocks:
- add `mariadb.spec.replication.primary.autoFailoverAllowUnknownLag`
- make `mariadb.spec.replication.standaloneProbes` affect readiness probes
- extend the logic of selecting failover candidates

Related: https://github.com/mariadb-operator/mariadb-operator/issues/1628

**Context:**

It seems to be possible to end up in a situation when the operator is deadlocked and cannot perform automatic primary failover because it sees no ready replicas:

- after the node hosting the primary goes down, the primary might become unhealthy only after replica is already marked not ready because k8s [1.35](https://github.com/kubernetes/kubernetes/blob/66452049f3d692768c39c797b21b793dce80314e/pkg/controller/nodelifecycle/config/v1alpha1/defaults.go#L40) gives a downed node ~50 seconds before it is marked not ready while mariadb cluster probes have a readiness threshold of ~30 seconds (failureThreshold: 3 * periodSeconds: 10) 

- or when `autoFailoverDelay` contributes to exceeding the readiness check  (my case)

- or when the operator goes down along with the primary, when the operator recoveres later, it is deadlocked since it is too late to the party and the replica is already marked not ready

This happens because current readiness probe is hardcoded to require `Seconds_Behind_Master`, but that gets nulled whenever the replica's IO thread is unable to reconnect to primary. 

**Notes:**

With this change, the user that enables `standaloneProbes` will not have their replicas marked as not ready because primary is not ok. But there will be still `Seconds_Behind_Master` check during candidate selection. And the new `autoFailoverAllowUnknownLag` will allow the user to select the strategy, risk or not. 

Yes, with `standaloneProbes` we will not be able to mark lagging replicas not ready and block read connections, but it is perfectly ok for those who don't split read/writes using routers.

`autoFailoverAllowUnknownLag` is set to true by default to not break current logic, as it's currently possible to have a ready replica but with already nulled `Seconds_Behind_Master`.


hedgieinsocks:
not stale

PR #1683, opened 2026-03-28
Improve cluster stability & recovery

YaCodesDevelopment:
# Summary

This PR fixes TLS handling for SQL connections, improves reconciliation flow, and expands controller event coverage.

# What changed

## SQL connection and TLS handling

* Refactored `getSqlOpts` in `connection_controller.go`:
  * Do not apply MariaDB TLS config when MaxScale is present but not using TLS.
  * Correctly handle client certificates when MaxScale uses TLS.
* Added unit tests in `connection_controller_unit_test.go` covering different MaxScale/MariaDB TLS scenarios.

## Reconciliation changes

* Added a **Primary failover** reconciliation phase in the MariaDB controller.
* SQL reconciliation now waits for the MariaDB writable service to be healthy before proceeding.
* User reconciliation now uses `waitForGrant` to block until the required `Grant` resource is ready.

## Controller events and RBAC

* Controller now watches labeled `PersistentVolumeClaim` changes and triggers reconciliation accordingly.
* RBAC updated:
  * Added `get` for `persistentvolumeclaims`
  * Added `databases` resource

## Other changes

* Added condition reason: `ConditionReasonReplicationNotConfigured`
* Small YAML and command fixes


YaCodesDevelopment:
@github-copilot please retry the review

mak011p:
We hit this in production on operator 26.3.0 — two separate `MariaDB` replication clusters in the same kube cluster.

**Reconcile error (matches your patched code path):**

```
Ready=False
message="Error reconciling Replica recovery: 1 error occurred:
  * error deleting StatefulSet: error getting StatefulSet:
    StatefulSet.apps \"<cluster>-mariadb\" not found"
```

**CR config (same shape on both clusters):**

```yaml
spec:
  replicas: 3
  replication:
    enabled: true
    gtidStrictMode: true
    primary:
      autoFailover: true
      autoFailoverDelay: 5m
    replica:
      gtid: CurrentPos
      recovery:
        enabled: true
        errorDurationThreshold: 30s
      bootstrapFrom:
        physicalBackupTemplateRef:
          name: <cluster>-mariadb-bootstrap-template   # storage.volumeSnapshot on csi-ceph-blockpool
```

Storage: Rook Ceph RBD.

**Two paths into the recovery loop.**

- Cluster A: rolling Talos worker reboot — replica lost primary connectivity >30s, `errorDurationThreshold: 30s` fired.
- Cluster B: manual primary switchover (`replication.primary.podIndex: 0 → 1`) flagged the old primary as a divergent-GTID broken replica.

Both reached `deleteStatefulSetLeavingOrphanPods`. STS got deleted. Next reconcile: `Get` returned NotFound, wedged on the wrapped error.

Your fix (`if apierrors.IsNotFound(err) { return nil }` in `deleteStatefulSetLeavingOrphanPods`) is correct — the desired state (no STS, orphan pods alive) is already met when the STS is absent.

**Mitigation on our side:** `replication.replica.recovery.enabled: false` cluster-wide + `spec.suspend: true` on the affected CRs. Holding the recovery flag off in prod until this PR ships in a release. Rough sense of timeline?

---

**Separate issue we hit as a cascade.** A raw `storage.volumeSnapshot` of a running mariadb isn't replication-consistent — GTID position in the data files lags InnoDB's committed state. A replica bootstrapped from such a snapshot replays binlog for rows it doesn't yet have; SQL thread aborts on `HA_ERR_KEY_NOT_FOUND` on high-write tables (for us, `magento.cron_schedule`).

On Cluster B the operator had already partially run the recovery loop before our `recovery.enabled: false` took effect — recreated the STS, bootstrapped pod -0 from the operator-created `<cluster>-mariadb-pb-recovery` snapshot, then got suspended mid-bootstrap. That replica is now permanently behind master.

Switching our bootstrap template to a `mariabackup`-based `PhysicalBackup` should fix it on our side. Is the new `mariadb_controller_replica_pvc.go` here intended to address the snapshot-consistency angle too, or is it worth a separate issue?

Happy to share full reconcile logs / status conditions if useful.

PR #1684, opened 2026-03-29
Add app.kubernetes.io/managed-by label to all operator-managed resources

### LINKED ISSUE #1186: [Feature]
I want to create cluster-wide network policies to ensure that operator pods always have access to the pods they create. Currently, the labels for all pods are either too specific (e.g. `app.kubernetes.io/instance: grafana-mariadb`) or too vague (e.g. `app.kubernetes.io/name: mariadb`):
```sh
❯ kubectl get pods -n grafana -o yaml grafana-mariadb-0 | yq eval '.metadata.labels'
app.kubernetes.io/instance: grafana-mariadb
app.kubernetes.io/name: mariadb
apps.kubernetes.io/pod-index: "0"
controller-revision-hash: grafana-mariadb-5975877476
k8s.mariadb.com/role: primary
statefulset.kubernetes.io/pod-name: grafana-mariadb-0

❯ kubectl get pods -n grafana -o yaml grafana-mariadb-metrics-6995f5c968-sfj9g | yq eval '.metadata.labels'
app.kubernetes.io/instance: grafana-mariadb-metrics
app.kubernetes.io/name: exporter
pod-template-hash: 6995f5c968

❯ kubectl get pods -n grafana -o yaml grafana-mariadb-maxscale-0 | yq eval '.metadata.labels'
app.kubernetes.io/instance: grafana-mariadb-maxscale
app.kubernetes.io/name: maxscale
apps.kubernetes.io/pod-index: "0"
controller-revision-hash: grafana-mariadb-maxscale-69d7997cf5
statefulset.kubernetes.io/pod-name: grafana-mariadb-maxscale-0

❯ kubectl get pods -n grafana -o yaml grafana-mariadb-maxscale-metrics-5545d78d5b-9759h | yq eval '.metadata.labels'
app.kubernetes.io/instance: grafana-mariadb-maxscale-metrics
app.kubernetes.io/name: exporte
```

I would like to have one static label for all pods created by the mariadb-operator to select only those pods. My proposal is to use one of the [recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels) as follows:
```yaml
app.kubernetes.io/managed-by: mariadb-operator
```

fvrk4n:
## Summary

Add the standard Kubernetes `app.kubernetes.io/managed-by: mariadb-operator` label to operator-managed objects built through `MetadataBuilder`. This enables users to target operator-created resources with a single label selector useful for NetworkPolicies, monitoring, and RBAC scoping.

Fixes #1186

## What changed

- Defined `KubernetesManagedByLabel` and `KubernetesManagedByValue`
- The label is enforced in `MetadataBuilder.Build()` so every resource gets it regardless of `inheritMetadata` configuration, this is an operator-owned label reflecting an operational fact, not a user preference (same approach as Prometheus Operator and etc.)

## What it does NOT touch

`LabelsBuilder` and `Selector.MatchLabels` are completely untouched. StatefulSet selectors are immutable after creation - adding a label there would break every existing deployment on upgrade.

## Upgrade impact

- Pod template labels change. With rolling update strategies this can trigger a one-time rollout on first reconciliation after upgrade; with non-rolling strategies, pods may require manual recreate/delete to pick up the label.
- Retroactive label backfill on already existing resources is currently reconciler-dependent. Services merge labels from the desired object today, while some other resources patch a narrower subset of fields. All newly created resources get the label immediately.
- VolumeClaimTemplate PVCs will not receive the label since they use `LabelsBuilder`, not `MetadataBuilder`.

## Follow-up

- Reconciler label drift fix: update StatefulSet, ConfigMap, Secret, and other reconcilers to patch `ObjectMeta.Labels` on existing resources (same pattern as `ServiceReconciler`)

### DIFF
--- a/pkg/builder/builder_test.go
+++ b/pkg/builder/builder_test.go
@@ -9,6 +9,7 @@ import (
 	mariadbv1alpha1 "github.com/mariadb-operator/mariadb-operator/v26/api/v1alpha1"
 	"github.com/mariadb-operator/mariadb-operator/v26/pkg/discovery"
 	"github.com/mariadb-operator/mariadb-operator/v26/pkg/environment"
+	mdbmetadata "github.com/mariadb-operator/mariadb-operator/v26/pkg/metadata"
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -52,6 +53,7 @@ func assertObjectMeta(t *testing.T, objMeta *metav1.ObjectMeta, wantLabels, want
 	if objMeta == nil {
 		t.Fatal("expecting object metadata to not be nil")
 	}
+	wantLabels = withManagedByLabel(wantLabels)
 	if wantLabels != nil && !reflect.DeepEqual(wantLabels, objMeta.Labels) {
 		t.Errorf("unexpected labels, want: %v  got: %v", wantLabels, objMeta.Labels)
 	}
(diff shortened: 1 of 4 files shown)


fvrk4n:
This PR adds the managed-by label at resource creation time. On upgrade, some existing resources won't receive it because their reconcilers only patch specific fields. For example, the ConfigMap reconciler patches Data but not ObjectMeta.Labels, while the Service reconciler already merges labels and picks up the change.

I plan to follow up with a PR that extends label patching to the remaining reconcilers so the label is consistently backfilled on upgrade.

PR #1688, opened 2026-03-31
Maintenance mode

mmontes11:
### Notes
- Avoid adding the cordon label if `spec.maintenance.enabled=false` 
- ~~`spec.maintenance.enabled` implies suspension to avoid overlapping operations such as cluster/replica recovery which may interfere. To simplify, just changed the condition in the `Suspend` stage, rather than defaulting `spec.suspend`.~~ We need the reconciliation to reach the `Maintenance` stage to drain conns and set readonly, we can't suspend. If we want to avoid reconciliation, set `spec.suspend=true`. We could do this for `MaxScale`, as we don't drain conns not set readonly there, but better having a consistent `maintenance` functionality.
- Continiously reconcile `readOnly` based on the desired state. This help not only for maintenance but also for fixing difts, which we do have with the replica Pods in the replication topology.


mmontes11:
Closed in favor of https://github.com/mariadb-operator/mariadb-operator/pull/1695

PR #1694, opened 2026-04-06
Feature/root password rotation

Michaelpalacce:
## TODO
- [x] Documentation 

## ⚠️ BREAKING CHANGES ⚠️
- The agent sidecar needs to be updated


mmontes11:
# PR Review: Feature/root password rotation

Great work on implementing root password rotation! This is a valuable feature for MariaDB operator. Here's my detailed feedback:

## ✅ Strengths

- **Well-architected**: The separation of concerns between internal secret management and password rotation is clean
- **Rollback mechanism**: The `restoreFunc` for rolling back failed password changes is excellent for safety
- **Comprehensive test coverage**: The integration test for replication scenarios is thorough

## ⚠️ Backwards Compatibility Concerns

**This PR is not fully backwards compatible without careful rollout.**

For existing Galera clusters:
1. Galera currently uses `MARIADB_ROOT_PASSWORD` env var for SST authentication
2. The new internal secret (`internal-{name}`) will be created by the reconciler
3. However, `ensureRootPassword()` only runs when `IsHAEnabled()`, and there's a risk of inconsistency between the internal secret and the env var used by Galera

**Recommended rollout strategy:**
- Phase 1: Deploy this version to create internal secrets without enabling rotation
- Phase 2: Once internal secrets are in place across all clusters, enable password rotation
- Consider adding logic to ensure `MARIADB_ROOT_PASSWORD` env var stays in sync with the internal secret for Galera compatibility

## 📝 Requested Changes

1. **Typo fix**: `apiHandlder` → `apiHandler` in `pkg/agent/router/router.go:67`

2. **Rollback logic**: The `restoreFunc` uses the same `sqlClient` that may be broken if password change failed. Consider creating a fresh connection with the old password for the rollback.

3. **Test coverage**: Add tests for:
   - Single instance (non-HA)
   - Galera cluster
   - Rollback scenarios

4. **Timeout values**: Consider using constants instead of hardcoded values (15s, 5s, 10s)

## 🚀 Overall

This is a solid implementation with good safety mechanisms. With the backwards compatibility considerations addressed and the minor fixes, this PR is ready to merge. The rollback capability is particularly well done and should give users confidence in the feature.

internal/controller/mariadb_controller.go
```diff
@@ -909,6 +944,254 @@ func (r *MariaDBReconciler) reconcileUsers(ctx context.Context, mariadb *mariadb
 			return result, err
 		}
 	}
+
+	return r.reconcileRootPassword(ctx, mariadb)
+}
+
+// reconcileRootPassword will ensure the root password in the mariadb CR is up to date
+// @NOTE: `root@localhost` and `root@%` are modified as both are created when MariaDB is first configured
+func (r *MariaDBReconciler) reconcileRootPassword(ctx context.Context, mariadb *mariadbv1alpha1.MariaDB) (ctrl.Result, error) {
```
mmontes11: `mariadb_controller.go` is getting a bit large and unmanageable, how about moving root password related logic to `mariadb_controller_root_password.go`?

internal/controller/mariadb_controller.go
```diff
@@ -909,6 +944,254 @@ func (r *MariaDBReconciler) reconcileUsers(ctx context.Context, mariadb *mariadb
...
+		health.WithEndpointPolicy(health.EndpointPolicyAll),
+	)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("error checking MariaDB health: %v", err)
+	}
+	if !healthy {
+		return ctrl.Result{RequeueAfter: time.Second}, nil
+	}
+
+	var jobList batchv1.JobList
```
mmontes11: Why this check? I pressume it may be for checking in-flight physical backup jobs? If so, maybe we can add a helper `ListMariaDBJobs` to `pkg/job.go`, similar to `ListJobs`. The idea is to potentially reuse it in other guard functions. We are currently not checking this in other guards, but maybe we should from now on when relevant.

internal/controller/mariadb_controller.go
```diff
@@ -909,6 +944,254 @@ func (r *MariaDBReconciler) reconcileUsers(ctx context.Context, mariadb *mariadb
...
+	rootPassLogger.Info("Root password updated successfully. Requeuing...")
+
+	return ctrl.Result{RequeueAfter: time.Second}, nil
+}
+
+// ensureRootPassword updates the root password based on the current value of `RootPasswordSecretKeyRef`
+// - Inside the agents for the probes
+// - If galera is enabled, updates `wsrep_sst_auth`
+func (r *MariaDBReconciler) ensureRootPassword(ctx context.Context, mariadb *mariadbv1alpha1.MariaDB) (ctrl.Result, error) {
+	if !mariadb.IsHAEnabled() {
```
mmontes11: Note: This will introduce the agent for the standalone topology:
- https://github.com/mariadb-operator/mariadb-operator/pull/1621

Can we add a TODO in that PR body to take this into account?

internal/controller/mariadb_controller.go
```diff
@@ -909,6 +944,254 @@ func (r *MariaDBReconciler) reconcileUsers(ctx context.Context, mariadb *mariadb
...
+			rootPassLogger.V(1).Info("Internal root password secret not found, requeuing")
+			return ctrl.Result{RequeueAfter: 1 * time.Second}, nil
+		}
+		return ctrl.Result{}, fmt.Errorf("error getting internal root password secret: %v", err)
+	}
+	internalRootPassword := internalRootPasswordSecret.Data[mariadb.Spec.RootPasswordSecretKeyRef.Key]
+
+	if newRootPassword == string(internalRootPassword) {
+		// Ensure root password is set as expected
+		return r.ensureRootPassword(ctx, mariadb)
```
mmontes11: This creates some reconciliation overhead, which is acceptable and preferred over having drifts between the agents/Galera configuration and the actual root password (we made this trade-off already in the codebase). However, my major concern here is that continuously reconciling means continuously sending the root password over the network. If TLS is not enabled, this can be very concerning.

In order to minimize this, I suggest that we keep a hash of the root password in the status, and only reconcile when it changes, something like:

```go
type MariaDBStatus struct {
	// RootPasswordHash is the hash of the root password used to track if the password has been reconciled.
	// +optional
	// +operator-sdk:csv:customresourcedefinitions:type=status
	RootPasswordHash *string `json:"rootPasswordHash,omitempty"`
}


if newRootPassword == string(internalRootPassword) {
	rootPasswordHash := hash.Hash(newRootPassword)
	if mariadb.Status.RootPasswordHash != nil && *mariadb.Status.RootPasswordHash == rootPasswordHash {
		// Hash matches, ensureRootPassword has already been reconciled
		return ctrl.Result{}, nil
	}
	// Hash doesn't match or not set, run ensureRootPassword
	if result, err := r.ensureRootPassword(ctx, mariadb); !result.IsZero() || err != nil {
		return result, err
	}
	// Update the hash after successful reconciliation
	return ctrl.Result{}, r.patchStatus(ctx, mariadb, func(status *mariadbv1alpha1.MariaDBStatus) error {
		status.RootPasswordHash = &rootPasswordHash
		return nil
	})
}
```

The `RootPasswordHash` acts as a marker to only reconcile `ensureRootPassword` once per password change. This significantly minimizes the security concern and reduces reconciliation overhead (less important). The hash will automatically be invalidated when the password changes, ensuring the password is still properly synchronized. For existing installations, the field will be nil initially and `ensureRootPassword` will run once to populate it, making this change backward compatible.

internal/controller/mariadb_controller.go
```diff
@@ -909,6 +944,254 @@ func (r *MariaDBReconciler) reconcileUsers(ctx context.Context, mariadb *mariadb
...
+	}
+
+	rootPassLogger.Info("Root password changed. Updating.")
+
+	sqlClient, err := mariadbsql.NewClientWithMariaDB(
+		ctx,
+		mariadb,
+		r.RefResolver,
+		mariadbsql.WithPassword(string(internalRootPassword)), // Using the internal password
+		mariadbsql.WithHost(mariadb.GetHost()),
```
mmontes11: `NewClientWithMariaDB` already uses `mariadb.GetHost()`, we can omit this option believe

internal/controller/mariadb_controller.go
mmontes11: I see that you have hooked it into `reconcileUsers`, which logically makes sense. Given its relevance, and as we are treating it like a different phase of the lifecycle, we can create a dedicated "Root password" phase for it? Easier to find/follow I believe.

pkg/agent/handler/environment.go
```diff
@@ -0,0 +1,82 @@
...
+	for i := 0; i < v.NumField(); i++ {
+		field := t.Field(i)
+		tag := field.Tag.Get("env") // this is the struct tag
+		envKey := strings.Split(tag, ",")[0]
+
+		if envKey == req.Key {
+			fieldValue := v.Field(i)
+			if fieldValue.CanSet() {
+				if fieldValue.Kind() == reflect.String {
+					fieldValue.SetString(req.Value)
```
mmontes11, line 71: Neat! We need to be mindful about sharing the same `PodEnvironment` reference, as there is nothing preventing us from creating a new hanlder receiving a copy of the `PodEnvironment`. Can't think of any way to prevent this other than leaving a comment in the setup in `main.go` . Thoughts?

internal/controller/mariadb_controller.go
mmontes11: I see that you have hooked it into reconcileUsers, which logically makes sense. Given its relevance, and as we are treating it like a different phase of the lifecycle, we can create a dedicated "Root password" phase for it and hook the `reconcileRootPassword` there? Easier to find/follow I believe.

internal/controller/mariadb_controller.go
```diff
@@ -909,6 +944,254 @@ func (r *MariaDBReconciler) reconcileUsers(ctx context.Context, mariadb *mariadb
...
+
+	rootPassLogger.Info("Root password updated successfully. Requeuing...")
+
+	return ctrl.Result{RequeueAfter: time.Second}, nil
+}
+
+// ensureRootPassword updates the root password based on the current value of `RootPasswordSecretKeyRef`
+// - Inside the agents for the probes
+// - If galera is enabled, updates `wsrep_sst_auth`
+func (r *MariaDBReconciler) ensureRootPassword(ctx context.Context, mariadb *mariadbv1alpha1.MariaDB) (ctrl.Result, error) {
```
mmontes11: How about `ensureRootPasswordInDataPlane`? Bit longer but you know what to expect by reading it.

mmontes11 (changes requested):
Great work! Works flawlessly! 

Left some suggestions. Will assign you an item of documentation to cover this more towards the end of the release. If you want to work on it, please create the `release-26.6.0` release and add it there to not create confusion with unreleased docs in the `main` branch.

pkg/agent/router/router.go
```diff
@@ -64,7 +64,7 @@ type ProbeHandler interface {
 	Readiness(w http.ResponseWriter, r *http.Request)
 }
 
-func NewRouter(apiHandlder RouteHandler, k8sClient ctrlclient.Client, logger logr.Logger, opts ...Option) http.Handler {
+func NewRouter(apiHandlders []RouteHandler, k8sClient ctrlclient.Client, logger logr.Logger, opts ...Option) http.Handler {
```
mmontes11: There is a typo here

mmontes11 (changes requested):

mmontes11:
Please note: I have updated [the expression](https://github.com/mariadb-operator/mariadb-operator/commit/edeb125b180eeb593b90eff00393c9e9dd933b5d) to run the full integration tests in branches like this one.

internal/controller/mariadb_controller.go
```diff
@@ -909,6 +944,254 @@ func (r *MariaDBReconciler) reconcileUsers(ctx context.Context, mariadb *mariadb
...
+	internalRootPasswordSecret.Data[mariadb.Spec.RootPasswordSecretKeyRef.Key] = []byte(newRootPassword)
+
+	if err := r.Patch(ctx, &internalRootPasswordSecret, patch); err != nil {
+		return ctrl.Result{}, fmt.Errorf(
+			"error while patching internal root password secret: %v",
+			multierror.Append(err, restoreFunc(err)),
+		)
+	}
+
+	rootPassLogger.Info("Root password updated successfully. Requeuing...")
```
mmontes11: How about emitting an event here?

internal/controller/mariadb_controller.go
Michaelpalacce: Agreed, moving it

internal/controller/mariadb_controller.go
Michaelpalacce: Will move it :+1: 

And yes, that's precisely why. Although it is not foolproof, race conditions can happen... this will detect any ongoing backups that are taking a long time for example, and if a backup is started the exact moment after this check, then it may fail, but that's acceptable

internal/controller/mariadb_controller.go
Michaelpalacce: I've considered this quite a bit actually and went back and forth on the implementation a few times.. I was considering the solution you described, but the situation I was worried about is the case where the `root` password was changed manually. The current solution will fix that, ensure the correct state, it is more simple and as you said we've made this decision before. 

As an alternative, what I was considering is:

```sql
SELECT Password
FROM mysql.user
WHERE User = 'root'
  AND Host = '%';
```
(and for 'root'@'localhost') and based on that make a decision.

internal/controller/mariadb_controller.go
Michaelpalacce: I love naming things :smile:

pkg/agent/handler/environment.go
Michaelpalacce: Outside of a major refactor... no. I will add the comment tho

internal/controller/mariadb_controller.go
Michaelpalacce: Alternatively, if security is important, we can say that TLS should be enabled. It'll keep the logic simpler

mmontes11:
@Michaelpalacce Something I've just noticed while testing:
```bash
2026-04-08T15:57:06.871+0200    ERROR   Reconciler error        {"controller": "mariadb", "controllerGroup": "k8s.mariadb.com", "controllerKind": "MariaDB", "MariaDB": {"name":"mariadb-repl","namespace":"default"}, "namespace": "default", "name": "mariadb-repl", "reconcileID": "a411f0a6-efe7-4409-8eae-36eb45c89114", "error": "error reconciling Root Password: 1 error occurred:\n\t* error setting root password in agent environment: error decoding body into error: json: cannot unmarshal number into Go value of type errors.APIError\n\n"}
```

This happened because I was running with an old agent sidecar (from another branch). Upgrading the agent sidecar is required when upgrading the operator. In this cases, I normally leave  __⚠️ BREAKING CHANGES ⚠️__ section in the PR body, to facilitate writing the release notes and the upgrade guide. Otherwise, we will likely forget this when releasing 26.06

internal/controller/mariadb_controller.go
mmontes11: > Alternatively, if security is important, we can say that TLS should be enabled. 

Yes, the documentation should detail the flow for rotating the root password and have a warning banner to enable TLS.

I still feel like, although we should be safe with TLS, the logs I am attaching will raise security concerns in an audit and about the product in general. This is targeting users with very high security and compliance standards, someone that will probably have recurrent security audits, and sending the root password over the network, will certainly not help given the sensitive nature of the root password. 

When provisioning (see [provision.log](https://github.com/user-attachments/files/26571870/provision.log)) we are making an HTTP call for setting the environment, but this can be avoided. When erroring the situation gets a bit worse, and we continously send the HTTP call (see [error.log](https://github.com/user-attachments/files/26571868/error.log)).

The feature is driven by a security need, therefore, I think in this case although the logic is a bit more complicated, it is something we should opt for. Let me suggest something to minimize the network calls with the root password, and at the same time make the function a bit more readable and encapsulated:

```go
	if newRootPassword == string(internalRootPassword) {
		if result, err := r.reconcileRootPasswordInDataPlane(ctx, mariadb, newRootPassword); !result.IsZero() || err != nil {
			return result, err
		}
		return ctrl.Result{}, nil
	}
	
	
	func (r *MariaDBReconciler) reconcileRootPasswordInDataPlane(ctx context.Context, mariadb *mariadbv1alpha1.MariaDB,
	newRootPassword string) (ctrl.Result, error) {
	rootPasswordHash := hash.Hash(newRootPassword)

	if mariadb.Status.RootPasswordHash == nil {
		// No need to call ensureRootPasswordInDataPlane the first time, during bootstrap
		return ctrl.Result{}, r.patchStatus(ctx, mariadb, func(status *mariadbv1alpha1.MariaDBStatus) error {
			status.RootPasswordHash = &rootPasswordHash
			return nil
		})
	}
	if *mariadb.Status.RootPasswordHash == rootPasswordHash {
		// Hash matches, ensureRootPasswordInDataPlane has already been reconciled
		return ctrl.Result{}, nil
	}
	// Hash doesn't match or not set, run ensureRootPassword
	if result, err := r.ensureRootPasswordInDataPlane(ctx, mariadb); !result.IsZero() || err != nil {
		return result, err
	}
	// Update the hash after successful reconciliation
	return ctrl.Result{}, r.patchStatus(ctx, mariadb, func(status *mariadbv1alpha1.MariaDBStatus) error {
		status.RootPasswordHash = &rootPasswordHash
		return nil
	})
}
```
This way, there is no network call during the bootstrap nor the reconciliation, only when the root password is rotated. The logic is slightly more elaborated but it is encapsulated in its own function, but by no means would be the most complex logic of the codebase 😄. 

```sql
SELECT Password
FROM mysql.user
WHERE User = 'root'
  AND Host = '%';
```
In this case, the password is continously sent over the network anyway, using a TCP protocol (MariaDB) instead of the HTTP used by the agent, so the root problem remains. Besides, 2 x queries are required: one for `root@%` and another one for `root@localhost`.

> I was worried about is the case where the root password was changed manually

In my opinion, there should be just one documented way of rotating the root password, which should be the Kubernetes-native given that we are an operator. Although we could undo manual actions done by the user, we should in general discourage them specially for cases like this one. To me, security is a more important priority for this use case.

mmontes11 (changes requested):

internal/controller/mariadb_controller.go
mmontes11: Wait a second, the signature could be:

```go
func (r *MariaDBReconciler) reconcileRootPasswordInDataPlane(ctx context.Context, mariadb *mariadbv1alpha1.MariaDB,
	newRootPassword string) (error) {
```

docs/api_reference.md
```diff
@@ -1213,6 +1218,45 @@ _Appears in:_
 | `values` _string array_ |  |  |  |
 
 
+#### Lifecycle
```
mmontes11, line 1221: @Michaelpalacce Thanks for taking care of the docs. Can we port them to the release branch? That way we will not be confusing users with unreleased features, as the docs are in `main`?

internal/controller/mariadb_controller_root_password.go
```diff
@@ -0,0 +1,287 @@
...
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			rootPassLogger.V(1).Info("Internal root password secret not found, requeuing")
+			return ctrl.Result{RequeueAfter: 1 * time.Second}, nil
+		}
+		return ctrl.Result{}, fmt.Errorf("error getting internal root password secret: %v", err)
+	}
+	internalRootPassword := internalRootPasswordSecret.Data[mariadb.Spec.RootPasswordSecretKeyRef.Key]
+
+	newRootPasswordHash := hash.Hash(newRootPassword)
```
mmontes11, line 70: We can avoid sending the root password over the network (2x agent + SQL) during provisioning and save a `ptr.Deref`:

```bash
	newRootPasswordHash := hash.Hash(newRootPassword)
	if newRootPassword == string(internalRootPassword) {
		if mariadb.Status.RootPasswordHash == nil {
			return ctrl.Result{}, r.patchRootPasswordHash(ctx, mariadb, newRootPasswordHash)
		}
		if *mariadb.Status.RootPasswordHash == newRootPasswordHash {
			return ctrl.Result{}, nil
		}
		if result, err := r.reconcileRootPasswordInDataPlane(ctx, mariadb); !result.IsZero() || err != nil {
			return result, err
		}

		return ctrl.Result{}, r.patchRootPasswordHash(ctx, mariadb, newRootPasswordHash)
	}
```

mmontes11 (changes requested):
I have created a `release-26.06` branch, can we port the docs changes there? I have also left a suggestion. Other than that, LGTM

mmontes11 (approved):
LGTM

PR #1695, opened 2026-04-07
Maintenance mode

mmontes11:
## TODO
- [ ] Documentation. Mention the use of the new label and that it should not be used by operators.

## Notes
- Avoid adding the cordon label if `spec.maintenance.enabled=false` 
- ~~`spec.maintenance.enabled` implies suspension to avoid overlapping operations such as cluster/replica recovery which may interfere. To simplify, just changed the condition in the `Suspend` stage, rather than defaulting `spec.suspend`.~~ We need the reconciliation to reach the `Maintenance` stage to drain conns and set readonly, we can't suspend. If we want to avoid reconciliation, set `spec.suspend=true`. We could do this for `MaxScale`, as we don't drain conns not set readonly there, but better having a consistent `maintenance` functionality.
- Continiously reconcile `readOnly` based on the desired state. This help not only for maintenance but also for fixing difts, which we do have with the replica Pods in the replication topology. 

## Proof of work

```bash
LAST SEEN   TYPE     REASON        OBJECT                 MESSAGE
7m39s       Normal   Maintenance   MariaDB/mariadb-repl   Draining process (id=149,command=Query,time=36)
0s          Normal   Maintenance   MariaDB/mariadb-repl   Draining process (id=458,command=Query,time=38)
0s          Normal   Maintenance   MariaDB/mariadb-repl   Enabling readonly in Pod mariadb-repl-0
0s          Normal   Maintenance   MariaDB/mariadb-repl   Disabling readonly in Pod mariadb-repl-0
0s (x2 over 40s)   Normal   Maintenance   MariaDB/mariadb-repl   Disabling readonly in Pod mariadb-repl-0
0s                 Normal   Maintenance   MariaDB/mariadb-repl   Enabling readonly in Pod mariadb-repl-0
0s                 Normal   Maintenance   MariaDB/mariadb-repl   Disabling readonly in Pod mariadb-repl-0
0s (x2 over 50s)   Normal   Maintenance   MariaDB/mariadb-repl   Disabling readonly in Pod mariadb-repl-0
0s                 Normal   Maintenance   MariaDB/mariadb-repl   Draining process (id=726,command=Query,time=36)
```


make/net.mk
```diff
@@ -23,6 +23,7 @@ host-mdb-test: ## Add MariaDB test hosts to /etc/hosts.
 .PHONY: host-mdb-emulated-external-test
 host-mdb-emulated-external-test: ## Add MariaDB test hosts to /etc/hosts.
 	@./hack/add_host.sh 47 mdb-emulate-external-test.default.svc.cluster.local
+	@./hack/add_host.sh 48 mdb-emulate-external-test-0.mdb-emulate-external-test-internal.default.svc.cluster.local
```
mmontes11, line 26: With the introduction of maintenance mode, connections via the internal server are performed and needed to complete the reconciliation successfully

api/v1alpha1/event_types.go
```diff
@@ -82,4 +82,7 @@ const (
 
 	// ReasonBinlogTimelineInvalid indicates that an error has occurred when validating the binlog timeline to be used for point-in-time recovery.
 	ReasonBinlogTimelineInvalid = "BinlogTimelineInvalid"
+
+	// ReasonMaintenance indicates that an action related to maintenance has been performed.
+	ReasonMaintenance = "Maintenance"
```
Michaelpalacce, line 87: This seems unused

pkg/controller/maintenance/read_only.go
```diff
@@ -0,0 +1,83 @@
...
+			if err := client.DisableReadOnly(ctx); err != nil {
+				return ctrl.Result{}, fmt.Errorf("error disabling readonly in Pod %s: %v", podName, err)
+			}
+		}
+	}
+	return ctrl.Result{}, nil
+}
+
+// getReadOnlyDesiredPodState returns a map with the desired readOnly state indexed by Pod index.
+func (r *MaintenanceReconciler) getReadOnlyDesiredPodState(mariadb *mariadbv1alpha1.MariaDB) map[int]bool {
```
mariadb-stefangenov: For Galera, if ReadOnly is disabled, we should set `read_only` to false everywhere

mariadb-stefangenov (approved):
LGTM outside a minor leftover Reason

Michaelpalacce (dismissed):
LGTM outside a minor leftover Reason

PR #1697, opened 2026-04-07
[multi-cluster] Multi-cluster topology

mmontes11:
Close https://github.com/mariadb-operator/mariadb-operator/issues/1543
Close https://github.com/mariadb-operator/mariadb-operator/issues/1538
Close https://github.com/mariadb-operator/mariadb-operator/issues/1530
Close https://github.com/mariadb-operator/mariadb-operator/issues/1504
Close https://github.com/mariadb-operator/mariadb-operator/issues/1218
Close https://github.com/mariadb-operator/mariadb-operator/issues/1214
Close https://github.com/mariadb-operator/mariadb-operator/issues/929
Close https://github.com/mariadb-operator/mariadb-operator/issues/878
Close https://github.com/mariadb-operator/mariadb-operator/issues/591
Close https://github.com/mariadb-operator/mariadb-operator/issues/220
Close https://github.com/mariadb-operator/mariadb-operator/issues/85
Close https://github.com/mariadb-operator/mariadb-operator/issues/76

## PRs
- https://github.com/mariadb-operator/mariadb-operator/pull/1695
- https://github.com/mariadb-operator/mariadb-operator/pull/1698
- https://github.com/mariadb-operator/mariadb-operator/pull/1720
- https://github.com/mariadb-operator/mariadb-operator/pull/1731
- https://github.com/mariadb-operator/mariadb-operator/pull/1740
- https://github.com/mariadb-operator/mariadb-operator/pull/1744

## TODO
- [ ] [Test GTID with multiple domain IDs](https://github.com/mariadb-operator/mariadb-operator/pull/1698/changes/879a5e46a631944da444d63ef4122580bbc604dc)

Issue #13, opened 2022-11-22
Helm repo down ?


krptg0:
Hi,

Seems like the helm repo charts.mmontes-dev.duckdns.org is down, it times out after some time.

Cheers,

mmontes11:
sorry about this @krptg0, the repo is back online!

Issue #12, opened 2022-11-03
Support for `MARIADB_PASSWORD_HASH` and `MARIADB_ROOT_PASSWORD_HASH`


mmontes11:
The oficial image of MariaDB now supports passing the passwords as a hash, check the environment variables documentation:
- https://hub.docker.com/_/mariadb/

This implies that the user would provide a hash instead of the actual password in the `Secret` , so he basically will need to run this command beforehand in another MariaDB instance:
> SELECT PASSWORD('thepassword')

This could be a security enhancement, since the Kubernetes secrets are just base64 encoded. However, it also requires the user to do extra work, so we should support providing the passwords as we are doing at the moment.

Issue #1, opened 2022-06-19
Perform upgrades in `MariaDB` CRD


mmontes11:
Some fields of the `StatefulSet` are inmutable, for example, upgrading the `volumeClaimTemplates` will result in the following error:
```
# statefulsets.apps "mariadb" was not valid:
# * spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', 'updateStrategy' and 'minReadySeconds' are forbidden
```
As an alternative, the operator could perform a blue-green deployment by provisioning the new StatefulSet first and decomissioning the old one afterwards.

PR #1698, opened 2026-04-07
[multi-cluster] Replication provisioning

mmontes11:
Related to:
- https://github.com/mariadb-operator/mariadb-operator/pull/1697

## ⚠️ BREAKING changes

- [Fix regression in MaxScale TLS](https://github.com/mariadb-operator/mariadb-operator/pull/1698/changes/5aa9e8b8a5fd43c314e9b071cf6a6a6798922402)

## Experiments

- https://github.com/mariadb-operator/mariadb-operator/pull/1717


mmontes11:
# PR #1698 Review: "[multi-cluster] Replication provisioning"

## Summary
Large PR (54 files, +3407/-501, 58 commits) introducing multi-cluster replication topology. Base branch is `feature-multi-cluster`. Self-declared breaking change in MaxScale TLS.

---

## Correctness

### ✅ Good
- **Topology abstraction** (`topology.go`) is clean — `Topology` interface with `SingleClusterTopology` and `multiClusterTopology` implementations properly separates concerns
- **GTID domain isolation** — multi-cluster uses distinct `gtid_domain_id` per cluster, preventing GTID conflicts across clusters
- **`log_slave_updates=ON`** correctly enabled for multi-cluster replicas to relay binary events across clusters
- **Switchover phases** are well-structured and ordered: lock → read_only → wait sync → configure new primary → connect replicas → change old primary to replica
- **Stale switchover recovery** properly cleans up lock/read_only and resets status
- **MaxScale auth provisioning** now gates on MariaDB readiness — prevents race conditions during provisioning and PITR

### ⚠️ Issues

**1. `configureSecondaryReplica` missing remote replication channel** (`topology.go:422-463`)

Secondary replicas in multi-cluster only configure the *local* master connection via `singleCluster.changeMaster()`, but never set up the remote replication channel to the primary cluster. The `StartSlave` at line 459 only starts the *local* slave. The primary replica (pod 0 in the replica cluster) gets `changeMasterPrimaryInPrimaryReplica` which sets up the remote channel, but secondary replicas within the same replica cluster don't.

**2. `ResetSlave` error handling in `configureSecondaryReplica`** (`topology.go:434-442`)

The remote slave reset tolerates `SQLConnectionNotExists` error, but the local slave reset at line 440 does not. If a fresh secondary replica has never had a local slave configured, `ResetSlave` could fail unexpectedly.

**3. `shouldReconcileSwitchover` blocks MaxScale switchover** (`switchover.go:34-35`)

Switchover is disabled when MaxScale is enabled. This seems overly restrictive — MaxScale should be able to handle primary switches transparently. If intentional, needs documentation.

**4. `configureNewPrimary` switchover path** (`switchover.go:309`)

Calls `topology.ConfigurePrimary()` which for multi-cluster primary replica will call `configurePrimaryReplica`, stopping all slaves and resetting the local slave. Need to verify the topology selection is correct for all switchover scenarios across both clusters.

---

## Safety

### ✅ Good
- **Switchover blocking conditions** are comprehensive: MaxScale, restoring backup, resizing storage, no configured replica
- **Automatic failover** respects same blocking conditions plus suspension state
- **Primary readiness check** before switchover phases — gracefully skips steps if primary is unavailable
- **`ErrDelayAutomaticFailover`** mechanism with configurable delay prevents premature failover

### ⚠️ Issues

**5. Automatic failover patches spec directly** (`pod_replication_controller.go:120-128`)

The `patch` method writes to `mdb.Spec.Replication.Primary.PodIndex`, mutating the spec in-place. If the spec patch fails, the status patch still runs with a mutated object. Consider using a copy for the status patch.

**6. No GTID lag threshold for failover candidate selection**

`FurthestAdvancedReplica` will promote the "least bad" replica even if all replicas are severely lagging. Worth considering a GTID lag threshold.

---

## Pitfall Detection

**7. `ReplicationRolePrimaryReplica` may not be assigned** (`mariadb_replication_types.go:481`)

The `PrimaryReplica` role is defined but the role assignment logic in the status update path doesn't appear to set it. The `shouldSkipPrimaryReconciliation` function checks for it, but without the role being assigned, multi-cluster primary replicas would fall through to the single-cluster path.

**8. `log_slave_updates=ON` on primary cluster replicas**

Enabling `log_slave_updates` on the primary cluster's replicas means they'll assign GTIDs from their own domain to relayed events. Only the replica cluster's nodes should need this — verify the primary cluster isn't affected.

**9. Server ID collision risk across clusters**

`serverIdStartIndex` defaults to 10. If two clusters use the same default without setting `GtidDomainID`, server IDs could collide. No validation enforces `GtidDomainID` when multi-cluster is enabled.

---

## Backwards Compatibility

### Breaking Change — MaxScale TLS

**10. MaxScale TLS now requires server cert refs** (`maxscale_controller_tls.go`)

Rename from `TLSAdminCASecretKey()` to `TLSListenerCASecretKey()`. Existing MaxScale resources with TLS enabled but without `serverCASecretRef` and `serverCertSecretRef` will fail to reconcile. Acceptable if documented in release notes.

### Non-breaking additions
- `MultiCluster` field is optional and nil by default
- `GtidDomainID` and `ServerIDStartIndex` are optional — defaults to server defaults
- `ServerCertAdditionalNames` in TLS is additive
- CRD changes are all additive
- `zz_generated.deepcopy.go` properly generated for all new types

---

## Code Quality

### ✅ Good
- Clean separation of topology strategies
- Well-named switchover phases with good logging
- Proper use of `errgroup` for parallel replica operations
- Comprehensive integration test in `mariadb_controller_multicluster_test.go`

### ⚠️ Issues

**11. `config.go` renamed purpose**

`ReplicationConfigClient` was removed and logic moved to `topology.go`. The file now only contains `NewReplicationConfig` (cnf template) and helpers. Consider renaming to reflect new purpose.

**12. Missing webhook validation for new fields**

No changes to `mariadb_webhook.go`. New `GtidDomainID` and `ServerIDStartIndex` fields have `webhook:"inmutable"` tags but no webhook enforces immutability or validates `GtidDomainID` is set when multi-cluster is enabled.

**13. `MultiClusterReplicaConnectionName` hardcoded** (`topology.go:20`)

The connection name `"replica"` is hardcoded. Won't scale if users need multiple remote connections.

**14. Test coupling**

773-line test file tightly coupled to specific cluster names and IP addresses. Consider parameterizing for reuse.

---

## Verdict

**LGTM with concerns.** The architecture is sound and the topology abstraction is well-designed.

| Priority | Issue | Action |
|----------|-------|--------|
| **P0** | Secondary replica missing remote replication channel (#1) | Fix before merge |
| **P0** | Missing webhook validation (#12) | Fix before merge |
| **P1** | No `GtidDomainID` validation for multi-cluster (#9) | Fix before merge |
| **P2** | Switchover blocked for MaxScale (#3) | Discuss/document |
| **P2** | Spec mutation in failover (#5) | Fix |
| **P3** | Hardcoded connection name (#13) | Future improvement |
| **P3** | No GTID lag threshold (#6) | Future improvement |

mmontes11:
@Michaelpalacce Please take a look. I'm still iterating the [agentic support ](https://github.com/mariadb-operator/mariadb-operator/pull/1718), here [the output](https://github.com/mariadb-operator/mariadb-operator/pull/1698#issuecomment-4342177093) in case you find it useful.

> Secondary replica missing remote replication channel (https://github.com/mariadb-operator/mariadb-operator/issues/1)

This is intended, secondary replicas connect to the primary replica, they don't have a remote connection.

> Missing webhook validation (https://github.com/mariadb-operator/mariadb-operator/issues/12)

To be added in upcoming multi-cluster PRs, when all the model is available i.e. Galera is supported.

mmontes11:
Here we go again... 

@Michaelpalacce Seems like we are getting Pods in Pending state when running the full tests, looking into it, last thing I want is introduce test unstability.

Multi-cluster tests were focused and were working here:
- https://github.com/mariadb-operator/mariadb-operator/pull/1698/changes/7da2cc5ed296f2c1e03c753f0b8fed41bca7f84f
- https://github.com/mariadb-operator/mariadb-operator/actions/runs/25097274794/job/73537130706

mmontes11:
@Michaelpalacce Multi-cluster __[tests are passing on my machine](https://www.reddit.com/r/ProgrammerHumor/comments/cw58z7/it_works_on_my_machine/)__:
```bash
❯ make test-int-multi-cluster
RELATED_IMAGE_MARIADB=docker-registry1.mariadb.com/library/mariadb:11.8.5 RELATED_IMAGE_MARIADB_NAME=docker-registry1.mariadb.com/library/mariadb RELATED_IMAGE_MAXSCALE=docker-registry2.mariadb.com/mariadb/maxscale:23.08.5 RELATED_IMAGE_EXPORTER=prom/mysqld-exporter:v0.15.1 RELATED_IMAGE_EXPORTER_MAXSCALE=docker-registry2.mariadb.com/mariadb/maxscale-prometheus-exporter-ubi:v0.0.1 MARIADB_GALERA_LIB_PATH=/usr/lib/galera/libgalera_smm.so MARIADB_OPERATOR_IMAGE=docker-registry3.mariadb.com/mariadb-operator/mariadb-operator:26.6.0-dev MARIADB_OPERATOR_NAME=mariadb-operator MARIADB_OPERATOR_NAMESPACE=default MARIADB_OPERATOR_SA_PATH=/tmp/mariadb-operator/token MARIADB_DEFAULT_VERSION=11.8 WATCH_NAMESPACE="" KUBEBUILDER_ASSETS="/home/martin/.local/share/kubebuilder-envtest/k8s/1.35.0-linux-amd64" /home/martin/go/src/github.com/mariadb-operator/mariadb-operator/bin/ginkgo --coverprofile=cover.out --timeout 1h10m  --label-filter="multi-cluster" ./internal/controller/...
Running Suite: Controller Suite - /home/martin/go/src/github.com/mariadb-operator/mariadb-operator/internal/controller
==========================================================================================================
============
Random Seed: 1777460516

Will run 13 of 171 specs
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS••••••••SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS•••••SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS

Ran 13 of 171 Specs in 464.139 seconds
SUCCESS! -- 13 Passed | 0 Failed | 0 Pending | 158 Skipped
PASS
coverage: 43.6% of statements
composite coverage: 43.6% of statements

Ginkgo ran 1 suite in 7m46.297635321s
```

I've made some changes for running them as part of the CI, let me know your thoughts

mmontes11:
All green 🟢 :
https://github.com/mariadb-operator/mariadb-operator/actions/runs/25112139776/job/73601361442?pr=1698

.github/workflows/ci.yml
```diff
@@ -141,17 +151,55 @@ jobs:
...
         run: make test-int-basic
 
       - name: Post Test Debugging (on failure)
         if: failure()
         run: make dump
 
+  # Multi-cluster integration tests are slow and create a significant amount of Pods (compute pressure), causing other tests to flake and/or fail.
+  # Using the default GitHub action workers: https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories
+  # It results in Pods ending up in Pending state, causing the Eventually blocks to fail: https://github.com/mariadb-operator/mariadb-operator/actions/runs/25092999123/job/73523204752?pr=1698
+  integration-test-multi-cluster:
```
Michaelpalacce, line 168: Why can't we reuse the cluster already created in `integration-test` stage? Can't we just put:

```yml
      - name: Run integration tests
        run: make test-int-multi-cluster
```
After the other tests with an if to check if it needs to be ran? Also, perhaps we can only run these tests when there is a push to `main`? That way we'll save a lot of compute (mainly thinking here about the enterprise operator CI) and time. The only downside here is that there may be a regression introduced, but we'll catch it on merge and can address it (we can revert the merge commit to `main` for one and re-do)

cmd/backup/main.go
```diff
@@ -318,6 +318,7 @@ func handleBackupMeta(ctx context.Context, backupLogger logr.Logger) error {
 	if err != nil {
 		return fmt.Errorf("error getting backup GTID: %v", err)
 	}
+	// TODO: test GTID with multiple domain IDs
```
Michaelpalacce, line 321: Did you intend to do this in this PR?

pkg/controller/replication/topology.go
```diff
@@ -0,0 +1,463 @@
...
+	}
+	if isReplica {
+		if err := client.StopSlave(ctx); err != nil {
+			return fmt.Errorf("error stopping slaves: %v", err)
+		}
+		if err := client.ResetSlave(ctx); err != nil {
+			return fmt.Errorf("error resetting slave: %v", err)
+		}
+		if err := client.ResetGtidSlavePos(ctx); err != nil {
+			// This error could happen when log_slave_updates=0N (multi-cluster, PITR),
```
Michaelpalacce, line 137: In order to fix this I was removing the redundant domain id entirely. For example when performing a cluster switchover, the new primary cluster no longer needs to track the position of the old primary cluster (the new replica) so I'd reset it. This was done in the multicluster controller. Not sure if this is helpful

Michaelpalacce: ~~But this is confusing. This is the `singleClusterTopology`, not the `multiClusterTopology`. In which cases is this relevant in regards to single cluster? Perhaps a case where we switched over already and decided to remove multi cluster and change the topology of the new primary cluster to not being part of multi cluster replication?~~

Nevermind, I saw that it's being reused by multicluster topology. I think it's fine so we don't have code duplication, but perhaps this may be case where code duplication is better for readability?

internal/controller/maxscale_controller.go
```diff
@@ -698,8 +698,31 @@ type maxscaleAuthReconcileItem struct {
...
+	mariadb, err := r.RefResolver.MariaDB(ctx, mxs.Spec.MariaDBRef, mxs.Namespace)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("error getting MariaDB: %v", err)
+	}
+	// This prevents:
+	// - MaxScale and MariaDB provisioning from happening in parallel.
+	// - Performing writes in mysql.user and  mysql.maxscale_config  (related to auth and monitor respectively)
+	// before MariaDB is fully configured and restored.
+	// - Provision MaxScale in the middle of a point-in-time restoration.
+	if (mariadb.IsReplicationEnabled() && !mariadb.HasConfiguredReplication()) ||
```
Michaelpalacce:
```go
	if (mariadb.IsReplicationEnabled() && !mariadb.HasConfiguredReplication()) ||
		(mariadb.IsGaleraEnabled() && !mariadb.HasGaleraConfiguredCondition()) ||
```
This seems like it can be moved to  a helper func in MariaDB

```go
mariadb.HasConfiguredHATopology()
```

pkg/controller/replication/controller.go
```diff
@@ -335,3 +335,16 @@ func (r *ReplicationReconciler) patchStatus(ctx context.Context, mariadb *mariad
 	patcher(&mariadb.Status)
 	return r.Status().Patch(ctx, mariadb, patch)
 }
+
+func shouldSkipPrimaryReconciliation(mariadb *mariadbv1alpha1.MariaDB, replRoles map[string]mariadbv1alpha1.ReplicationRole,
```
Michaelpalacce, line 339: One thing I was thinking when writing this was that we are very dependent on the status. In cases where the connection is not as stable we will continuously try to do ConfigurePrimary, performing a lot of resets. I wonder if in the future we should think of debouncing this *after* the first configuration

Michaelpalacce:
The integration tests for multicluster are failing for me, odd... They are passing for you and the runner.

```


2026-05-04T13:14:47.958+0300	ERROR	Reconciler error	{"controller": "mariadb", "controllerGroup": "k8s.mariadb.com", "controllerKind": "MariaDB", "MariaDB": {"name":"mar            iadb-eu-central","namespace":"default"}, "namespace": "default", "name": "mariadb-eu-central", "reconcileID": "bacc3760-3cb6-49ce-9b3f-281b778a732a", "error": "error reconciling Init: 1 error occurred:\n\t* error ensuring replication configured: error configuring Pod mariadb-eu-central-1: context deadline exceeded\n\n"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler
	/home/stef/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.23.3/pkg/internal/controller/controller.go:495
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem
	/home/stef/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.23.3/pkg/internal/controller/controller.go:438
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func1.1
	/home/stef/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.23.3/pkg/internal/controller/controller.go:313

```

Michaelpalacce:
I've noticed that the multicluster status cannot be checked with `kubectl get mdb` should we add it (maybe as an expanded argument)?

Michaelpalacce:
Cluster switchover is not part of this PR, correct?

mmontes11:
@Michaelpalacce yes, cluster switchover out of the scope of this PR, will be added in following PRs.

> I've noticed that the multicluster status cannot be checked with kubectl get mdb should we add it (maybe as an expanded argument)?

As far as I understood, not needed for provisioning. Should we leave it for the cluster switchover PR?

> The integration tests for multicluster are failing for me, odd... They are passing for you and the runner.

Did you find the underlying reason? Struggled a lot with Pods in pending state, and therefore MariaDB ready sttus not becoming ready. It happened in the runnet, when multiple Pods (6) were running at the same time, and therefore scheduler refused.

Michaelpalacce:
@mmontes11 

> As far as I understood, not needed for provisioning. Should we leave it for the cluster switchover PR?
Yeah, we can do it in a later PR.

> Did you find the underlying reason? Struggled a lot with Pods in pending state, and therefore MariaDB ready sttus not becoming ready. It happened in the runnet, when multiple Pods (6) were running at the same time, and therefore scheduler refused.

Must have been some sort of resource contention. It happened a few times back when I was originally writing it... Cluster re-createion didn't help, but after a system reboot, it got fixed. Odd.

mmontes11:
@Michaelpalacce Tests are crucial at this point, having to manually orchestrate the whole manual multi-cluster setup is time-consuming. If you keep seeing unstability, please let me know. 

The motivation to run them separately was precisely to achieve predictability via better isolation. I'm concious of the overhead of this, but I think it is much better than run by hand. If predictability is still not achieved, then the overhead would not be justified IMHO. Worth investing some time if needed, to deliver and refactor this with confidence

Michaelpalacce (changes requested):
Well separated from the other functionality made it easier to review 😄 I left a few suggestions

.github/workflows/ci.yml
mmontes11: The initial intention was to not affect the feedback loop on the rest of the functionality, but I guess we can take a look at the CI logs if we introduce another step.

For community operator, we can afford the compute overhead, probably for enterprise operator we don't, so we will need to diverge in the CI here. To align a bit more, I suggest incorporating the step above, but running it on all branches for community operator to catch regressions early. For enterprise operator, we'll evaluate when this functionality gets ported, probably running only in `main` would make sense, we can discuss in the PR.

Thoughts?

cmd/backup/main.go
mmontes11: Initially yes, but it was too ambitious. Added it as a TODO in the main PR, the points to be adapted are now identified, but this will be out of the scope for the first version.

pkg/controller/replication/topology.go
mmontes11: > This was done in the multicluster controller. Not sure if this is helpful
> but perhaps this may be case where code duplication is better for readability?

Yes, it was, I'm aware that additional tweaks will be required for the multi-cluster switchover. Not affecting the single cluster topology will be crucial, multi-cluster topology will be available but existing (and probably most of users) will be on single cluster topology. For this reason I think:
-  Additional logic specific to multi-cluster should be introduced in either the multi-cluster topology or the  the multi-cluster controller when it gets introduced.
- A bit of duplication may be acceptable in exchange for  better readability and backwards compatibility in the single cluster. Also, by having `ConfigurePrimary` and `configurePrimaryReplica` (with some instructions duplicated, but not too concerning IMHO), we could introduce future tweaks specific to multi-cluster or single cluster. For example. we may decide that `ResetGtidSlavePos` is acceptable for multi-cluster, but it might not for single-cluster, which is the motivation of introducing different methods, yet, being able to use composition to re-utilize methods from single cluster in multi cluster.

internal/controller/maxscale_controller.go
mmontes11: Good suggestion, agreed.

pkg/controller/replication/controller.go
mmontes11: There is certainly room for improvement here, but I think we have enough in our plate right now. I'd rather release in late may and catch any potential issues introduced by multi-luster.

> debouncing this after the first configuration

You mean, perform only the continous reconciliation after the first configuration? Continous reconciliation is crucial for fixing breaking replication, but at the same time it could end up in a _primary reset loop_. 

The classic _primary reset loop_ happens when there are no replicas connected to the primary, and that could be because of:
1. Replicas broken, therefore not connecting
2. Primary broken and therefore refusing connections

I think this function should account for `2.`, and avoid primary reset loops when `1.`, it will be crucial to use statuses more precisely, or even introduce them if needed. But as I said, this would need to be studied separately, not as part of multi-cluster.

.github/workflows/ci.yml
mmontes11: This has been addressed [as part of the Galera PR,](https://github.com/mariadb-operator/mariadb-operator/pull/1720) see:
https://github.com/mariadb-operator/mariadb-operator/blob/feature-multi-cluster-galera-prov/.github/workflows/ci.yml

api/v1alpha1/mariadb_multi_cluster_types.go
```diff
@@ -0,0 +1,51 @@
...
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:booleanSwitch"}
+	Enabled bool `json:"enabled,omitempty"`
+}
+
+type MultiClusterSpec struct {
+	// Primary is the name of the primary cluster. It refers to a member in the 'members' field, containing its full specification.
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:advanced"}
+	Primary string `json:"primary,omitempty"`
+	// Replicas is the name of all replica clusters. They refer to a member in the 'members' field, containing its full specification.
```
mariadb-stefangenov, line 22:
```suggestion
	// Replicas are the names of all replica clusters part of the topology. They each refer to a member in the 'members' field, containing its full specification.
```

mariadb-stefangenov (approved):
LGTM! One minor typo, nothing too crazy. Tested it and works still

Michaelpalacce (dismissed):

api/v1alpha1/mariadb_multi_cluster_types.go
mmontes11: This field was removed. With `members` and `primary` should be sufficient, no need to explicitely define replicas, at least not for now with the currently supported topologies.

PR #1711, opened 2026-04-16
Fix `PhysicalBackup` deadlock and add dynamic pod affinity

mmontes11:
- Fix waitForRunningJobs() to skip failed Jobs instead of blocking new schedules
- Add FailedJobsHistoryLimit to allow cleanup of failed Jobs
- Replace static nodeSelector with dynamic podAffinity for backup Jobs
- This prevents permanent deadlock when Jobs fail due to node replacement or failover
- Add debug logging to track failed Job handling


mmontes11:
Closed in favor of:
https://github.com/mariadb-operator/mariadb-operator/pull/1712

PR #1712, opened 2026-04-16
Fix `PhysicalBackup` deadlock and add dynamic pod affinity

mmontes11:
- Fix waitForRunningJobs() to skip failed Jobs instead of blocking new schedules
- Add FailedJobsHistoryLimit to allow cleanup of failed Jobs
- Replace static nodeSelector with dynamic podAffinity for backup Jobs
- This prevents permanent deadlock when Jobs fail due to node replacement or failover
- Add debug logging to track failed Job handling


internal/controller/physicalbackup_controller_job.go
```diff
@@ -183,28 +183,46 @@ func (r *PhysicalBackupReconciler) cleanupJobs(ctx context.Context, backup *mari
...
+			if err != nil && !apierrors.IsNotFound(err) {
+				return fmt.Errorf("error deleting Job \"%s\": %v", job.Name, err)
+			}
+			logger.V(1).Info("Deleted old Job", "job", job.Name, "physicalbackup", backup.Name)
+		}
 	}
 
-	for i := maxHistory; i < len(completeJobs); i++ {
-		job := completeJobs[i]
+	maxFailedHistory := int(ptr.Deref(backup.Spec.FailedJobsHistoryLimit, 0))
```
Michaelpalacce: Better to default to 5 here, no?

internal/controller/physicalbackup_controller_job.go
```diff
@@ -183,28 +183,46 @@ func (r *PhysicalBackupReconciler) cleanupJobs(ctx context.Context, backup *mari
...
+	if len(failedJobs) > maxFailedHistory {
```
Michaelpalacce, line 206: I wonder if we should extract this logic to a separate func in `pkg/job/job.go`? Or in this package

pkg/builder/batch_builder.go
```diff
@@ -260,20 +261,34 @@ func (b *Builder) BuildPhysicalBackupJob(key types.NamespacedName, backup *maria
...
+	if ptr.Deref(backup.Spec.PodAffinity, true) {
+		// Use dynamic pod affinity to schedule the Job on the same node as the MariaDB Pod.
+		// This allows the Job to be rescheduled if the MariaDB Pod moves to a different node
+		// (e.g., due to node replacement or eviction).
+		// Required for ReadWriteOnce storage.
+		affinity = &corev1.Affinity{
+			PodAffinity: &corev1.PodAffinity{
+				RequiredDuringSchedulingIgnoredDuringExecution: []corev1.PodAffinityTerm{
+					{
+						LabelSelector: &metav1.LabelSelector{
```
Michaelpalacce, line 279: Shouldn't we use the Labels builder? That way we don't have to expose `InstanceLabel` and `StatefulSetPodName`

Michaelpalacce (changes requested):
Overall, looks good. Left some minor suggestions

mmontes11:
@Michaelpalacce  Done! All three suggestions addressed:

1. `FailedJobsHistoryLimit` now defaults to `5` instead of `0`
2. Extracted the deletion logic into a `deleteOldJobs` helper function in `physicalbackup_controller.go`
3. Using `LabelsBuilder.WithInstance()` and `WithPodName()` instead of directly accessing the label constants. `InstanceLabel` and `StatefulSetPodName` have been reverted to private.

mmontes11:
@Michaelpalacce Just realized something while testing the PR. We export the job's `backoffLimit` in the `PhysicalBackup` resource. It is important to tune it accordingly to determine how many job container restarts are required to mark the job (and therefore the `PhysicalBackup` also) as "Failed". By default it is 5, meaning that 5 restarts will be required to be considered as "Failed" and therefore, proceed to schedule a new job.

Michaelpalacce (approved):
LGTM 👍

PR #1716, opened 2026-04-23
Add replication role label to metrics

hedgieinsocks:
This PR adds an optional `role` label to metrics that can be either `primary` or `replica`.

This can be used by consumers to fine tune custom alerts.


mmontes11 (changes requested):
LGTM! Thank you very much for the contribution. Would you mind adding a small unit test to verify that the `RelabelConfigs` are effectively added? 🙏🏻

mmontes11 (approved):
LGTM! Thanks for your contribution 🦔

PR #1720, opened 2026-04-29
[multi-cluster] Galera provisioning

mmontes11:
Related to:
- https://github.com/mariadb-operator/mariadb-operator/pull/1697

Depends on ⚠️ : 
- https://github.com/mariadb-operator/mariadb-operator/pull/1698


api/v1alpha1/mariadb_galera_types.go
```diff
@@ -373,6 +373,25 @@ type GaleraSpec struct {
 	// +optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:advanced"}
 	Config GaleraConfig `json:"config,omitempty"`
+	// GtidDomainID is the domain ID to be used in GTID mode, enabled when the multi-cluster topology is used.
```
Michaelpalacce, line 376: For `GtidDomainID` and `ServerID` we say:

>	// Make sure it has a different value on each the member of a multi-cluster topology.

But actually for gtid_domain_id we need to have a different value on each node in each cluster, without any overlaps (so for 2 clusters with 3 nodes each, it'll potentiall be 1,2,3 and 4,5,6). And for `server_id` we need it the same on each node in a cluster, but different between clusters (same example as above would be: 1,1,1 and 2,2,2)

internal/controller/mariadb_controller_status.go
```diff
@@ -171,36 +168,51 @@ func (r *MariaDBReconciler) getReplicationRole(ctx context.Context, mdb *mariadb
 	return role, nil
 }
 
-func shouldSkipReplicaStatusForPod(mdb *mariadbv1alpha1.MariaDB, podIndex int) bool {
-	isPrimary := podIndex == *mdb.Status.CurrentPrimaryPodIndex
+func (r *MariaDBReconciler) isPrimaryReplica(ctx context.Context, mdb *mariadbv1alpha1.MariaDB, client *sql.Client,
```
Michaelpalacce, line 171: Perhaps the name should reflect the fact that this is regarding MultiCluster? Additionally it may be good to check that if multicluster is enabled even here?

internal/controller/mariadb_controller_multicluster_test.go
Michaelpalacce, line 1: One thing I noticed, while the tests do pass, they are not explicit enough. What I managed to find out on my environment is that you cannot create a multi cluster environment with a galera cluster that has writes in it. I tested by:

1. `kubectl apply -f examples/manifests/multi-cluster/galera/eu-south.yaml`
2. `kubectl apply -f examples/manifests/multi-cluster/galera/eu-south-backup.yaml`
3. Seed data:

```sql
CREATE DATABASE test;
USE test;
CREATE TABLE test_sync (
    id INT PRIMARY KEY AUTO_INCREMENT,
    test_data VARCHAR(100)
);
INSERT INTO test_sync (test_data) VALUES ('Data from Cluster 1 Master: 1');
INSERT INTO test_sync (test_data) VALUES ('Data from Cluster 1 Master: 2');
INSERT INTO test_sync (test_data) VALUES ('Data from Cluster 1 Master: 3');
INSERT INTO test_sync (test_data) VALUES ('Data from Cluster 1 Master: 4');
INSERT INTO test_sync (test_data) VALUES ('Data from Cluster 1 Master: 5');
INSERT INTO test_sync (test_data) VALUES ('Data from Cluster 1 Master: 6');
INSERT INTO test_sync (test_data) VALUES ('Data from Cluster 1 Master: 7');
INSERT INTO test_sync (test_data) VALUES ('Data from Cluster 1 Master: 8');
INSERT INTO test_sync (test_data) VALUES ('Data from Cluster 1 Master: 9');
SELECT * FROM test.test_sync;
```

4. `kubectl apply -f examples/manifests/multi-cluster/galera/eu-central.yaml`

I think the same applies when you flip 2. and 3.

Michaelpalacce: The above is not necessarily a problem with the tests, but it should be something we are supporting I believe.

pkg/galera/config/config.go
```diff
@@ -204,6 +229,25 @@ func (c *ConfigFile) getProviderOptions(env *environment.PodEnvironment, options
 	return providerOpts.marshal(), nil
 }
 
+// gtidDomainID can be used to get a new 'gtid_domain_id' from the pod name.
+// We are relying on the natural incrementation of pod naming in statefulsets.
+// The 'offset' is added since 'gtid_domain_id' and 'wsrep_gtid_domain_id' must never match and they will for the first pod ("-0")
+// See: https://mariadb.com/docs/galera-cluster/high-availability/using-mariadb-replication-with-mariadb-galera-cluster/configuring-mariadb-replication-between-two-mariadb-galera-clusters
+func (c *ConfigFile) gtidDomainID(podName string, offset int) (*int, error) {
```
Michaelpalacce, line 236: I think the offset can be a constant here

Michaelpalacce (changes requested):
Outside of one issue with the creation, the rest are pretty minor!

internal/controller/mariadb_controller_multicluster_test.go
mmontes11: Unable to reproduce following these steps in the current `feature-multi-cluster` branch:
1. `kubectl apply -f examples/manifests/multi-cluster/galera/eu-south.yaml`
2. `kubectl apply -f examples/manifests/multi-cluster/galera/eu-south-backup.yaml`
3. `make client` changing the `host` flag and configuring `ca-cert` with `mariadb-server-ca`. It should be equivalent to your SQL above:
```bash
❯ make client
go run hack/client/*.go --host=mariadb-eu-south-primary.default.svc.cluster.local --port=3306 --username=root --password=MariaDB11! --database=test --table=test --timeout=3s --ca-cert=ca.crt --mode=insert --insert-interval=1s
2026/05/22 11:31:28 inserted at 1779442288
2026/05/22 11:31:29 inserted at 1779442289
2026/05/22 11:31:30 inserted at 1779442290
2026/05/22 11:31:31 inserted at 1779442291
2026/05/22 11:31:32 inserted at 1779442292
2026/05/22 11:31:33 inserted at 1779442293
2026/05/22 11:31:34 inserted at 1779442294
2026/05/22 11:31:35 inserted at 1779442295
...
```
4. In primary cluster:
```bash
MariaDB [(none)]>  SELECT @@global.hostname, COUNT(*) FROM test.test;
+--------------------+----------+
| @@global.hostname  | COUNT(*) |
+--------------------+----------+
| mariadb-eu-south-0 |      42 |
+--------------------+----------+
```
5. `kubectl apply -f examples/manifests/multi-cluster/galera/eu-central.yaml`
6. In replica cluster:
```bash
MariaDB [(none)]> SELECT @@global.hostname,@@global.gtid_current_pos, COUNT(*) FROM test.test;
+----------------------+---------------------------+----------+
| @@global.hostname    | @@global.gtid_current_pos | COUNT(*) |
+----------------------+---------------------------+----------+
| mariadb-eu-central-0 | 0-1-717,10-2-6            |      712 |
+----------------------+---------------------------+----------+
1 row in set (0.000 sec)
```
```bash
MariaDB [(none)]> SELECT @@global.hostname,@@global.gtid_current_pos, COUNT(*) FROM test.test;
+----------------------+---------------------------+----------+
| @@global.hostname    | @@global.gtid_current_pos | COUNT(*) |
+----------------------+---------------------------+----------+
| mariadb-eu-central-1 | 0-1-745,10-2-6            |      740 |
+----------------------+---------------------------+----------+
1 row in set (0.000 sec)
```

api/v1alpha1/mariadb_galera_types.go
mmontes11: Exactly, and I believe we are handling this the right way internally:

`mariadb-eu-south-0`:
```bash
# Multi-cluster
log-bin
log_slave_updates=ON
wsrep_gtid_mode=ON
wsrep_gtid_domain_id=0
gtid_domain_id=1
server_id=1
```
`mariadb-eu-south-1`:
```bash
# Multi-cluster
log-bin
log_slave_updates=ON
wsrep_gtid_mode=ON
wsrep_gtid_domain_id=0
gtid_domain_id=2
server_id=1
```
`mariadb-eu-central-0`:
```bash
# Multi-cluster
log-bin
log_slave_updates=ON
wsrep_gtid_mode=ON
wsrep_gtid_domain_id=10
gtid_domain_id=11
server_id=2
```
`mariadb-eu-central-1`:
```bash
# Multi-cluster
log-bin
log_slave_updates=ON
wsrep_gtid_mode=ON
wsrep_gtid_domain_id=10
gtid_domain_id=12
server_id=2
```

The comment on the field refers to the actual value the user interacts with, most of them won't care about the internals. If they want to get into specifics, they can check the linked docs or the `/etc/mysql/mariadb.conf.d/0-galera.cnf`

Thoughts?

pkg/galera/config/config.go
mmontes11: Addresed by https://github.com/mariadb-operator/mariadb-operator/pull/1697/changes/1873b982fa901f4a7e76dbb64d0d17ddbf435124

internal/controller/mariadb_controller_status.go
mmontes11: Addressed by:
https://github.com/mariadb-operator/mariadb-operator/pull/1697/changes/68fbe02a4056203a0b58b2727b5dc0f8a158cb73
https://github.com/mariadb-operator/mariadb-operator/pull/1697/changes/6f5a95d6292446e862c946f19118ba8b734325a0

mmontes11:
@Michaelpalacce can you take another look?

PR #1733, opened 2026-05-13
fix: refuse switchover when target replica has fatal replication error

xcompass:
## Summary

Adds an early guard in `reconcileSwitchover` that refuses to start a switchover when the target replica's last-observed `Status.Replication.Replicas[pod]` indicates a fatal replication state (non-zero `LastSQLErrno`/`LastIOErrno`, or `Slave_SQL_Running=No` / `Slave_IO_Running=No` with errno zero).

Without this guard the switchover state machine repeatedly locked the still-healthy current primary read-only and then timed out in `MASTER_GTID_WAIT` for as long as the wedge persisted — observed on **moodle-education-stg-db** and **moodle-education-prod-db** within a week of each other, both stuck for 5+ days with `errno 1062` (Duplicate entry on `moodle.mdl_task_log`) until manual rollback of `spec.replication.primary.podIndex`.

## Why read cached status, not live SQL

In the prod incident the target's `mariadb` container was in `CrashLoopBackOff` with 2376 restarts. A live `ReplicaStatus` call would itself have failed with `invalid connection` and masked the actual cause. `MariaDB.Status.Replication.Replicas` is populated by the status reconciler in earlier loops and carries the last good observation across the bouncing container.

## Failure modes the guard catches

| Check | Field-incident example |
|---|---|
| `LastSQLErrno != 0` | `1062` Duplicate entry on `mdl_task_log` (this PR's seed) |
| `LastIOErrno != 0` | `1236` master is missing the GTID |
| `Slave_SQL_Running=No` (errno=0) | `slave_skip_errors` aftermath, manual `STOP SLAVE SQL_THREAD` |
| `Slave_IO_Running=No` (errno=0) | network drop or manual stop |

`SlaveSQLRunning`/`SlaveIORunning` default to `true` via `ptr.Deref(_, true)` so an uninitialized replica (status never observed yet) does not block the very first switchover; the guard fires only after the status reconciler has populated an explicit `false`.

## Explicit non-goal

This does **not** unblock auto-recovery on the wedged target — that path remains gated by `shouldReconcileReplicaRecovery`'s `IsSwitchingPrimary` / `IsReplicationSwitchoverRequired` exclusion. Plumbing recovery to run concurrently with a wedged switchover is a larger behavioural change and tracked as a follow-up. This PR narrowly addresses the *infinite-retry-that-keeps-degrading-the-healthy-primary* half.

## Test plan

- [x] Unit: 10 table-driven cases for `targetReplicaError` (`TestTargetReplicaError_GuardsAgainstWedgedTarget`)
  - nil status, target == current primary, healthy target
  - SQL errno 1062, IO errno 1236
  - `Slave_SQL_Running=No` with errno=0
  - `Slave_IO_Running=No` with errno=0
  - nil thread-running fields (uninitialized replica)
  - target absent from status map, unset spec PodIndex
- [x] `go test ./pkg/controller/replication/... -race` clean
- [x] `gofmt`, `go vet`, full `go build ./...` clean
- [x] End-to-end coverage of the `reconcileSwitchover` call site follows existing convention in this file (integration-test-only with a documented design record)
- [ ] Field validation on staging cluster after merge


xcompass:
Sorry, closing this. wrong repo

PR #1790, opened 2026-06-08
feat(password): improve password generation

mmontes11:
This PR introduces multiple fixes:

## Password generation to match MariaDB `simple_password_check` policy 

The go-password library does not guarantee at least one uppercase letter, causing Error 1819 on MariaDB password policy. Add pkg/password module that validates and retries generation to ensure all character classes are represented.

Since the `simple_password_check` policy is configurable via:

```bash
simple_password_check_digits
simple_password_check_letters_same_case
simple_password_check_minimal_length
simple_password_check_other_characters
```

The ideal scenario would be a password generator aware of these variables. This is the first towards that, but not a final solution.


pkg/password/password.go
```diff
@@ -0,0 +1,85 @@
+package password
```
mariadb-stefangenov, line 1: To be honest, at this point isn't it better to write our own password generator:

```go
package main

import (
	"crypto/rand"
	"fmt"
	"math/big"
)

const (
	lowerChars  = "abcdefghijklmnopqrstuvwxyz"
	upperChars  = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
	digitChars  = "0123456789"
	symbolChars = "!@#$%^&*()-_=+[]{}|;:,.<>/?"
)

// At least one lower, one upper, 4 digits, 2 symbols
var charGroups = []string{
	lowerChars,
	upperChars,
	digitChars,
	digitChars,
	digitChars,
	digitChars,
	symbolChars,
	symbolChars,
}

const allChars = lowerChars + upperChars + digitChars + symbolChars

func GeneratePassword(length int) (string, error) {
	if length < len(charGroups) {
		return "", fmt.Errorf("length must be at least %d", len(charGroups))
	}

	password := make([]byte, 0, length)

	for _, group := range charGroups {
		ch, err := randomChar(group)
		if err != nil {
			return "", err
		}
		password = append(password, ch)
	}

	for len(password) < length {
		ch, err := randomChar(allChars)
		if err != nil {
			return "", err
		}
		password = append(password, ch)
	}

	if err := shuffle(password); err != nil {
		return "", err
	}

	return string(password), nil
}

func randomChar(chars string) (byte, error) {
	if len(chars) == 0 {
		return 0, fmt.Errorf("character set cannot be empty")
	}

	n, err := rand.Int(rand.Reader, big.NewInt(int64(len(chars))))
	if err != nil {
		return 0, err
	}

	return chars[n.Int64()], nil
}

func shuffle(items []byte) error {
	for i := len(items) - 1; i > 0; i-- {
		j, err := secureRandInt(i + 1)
		if err != nil {
			return err
		}

		items[i], items[j] = items[j], items[i]
	}

	return nil
}

func secureRandInt(max int) (int, error) {
	n, err := rand.Int(rand.Reader, big.NewInt(int64(max)))
	if err != nil {
		return 0, err
	}

	return int(n.Int64()), nil
}

func main() {
	password, err := GeneratePassword(32)
	if err != nil {
		fmt.Println("Error:", err)
		return
	}

	fmt.Println(password)
}
```

- We can make it more sophisticated, of course, but I think this is good enough

mariadb-stefangenov (changes requested):
I think we don't need an extra dependency for something like this.

mariadb-stefangenov (approved):
I'd have moved `isCompliant` probably to the `_test.go` file, but either way looks good

PR #1823, opened 2026-07-03
Add extra objects templating to mariadb-cluster chart

hedgieinsocks:
Allow users to template custom objects without the need to use external raw charts as a dependency. 

This will be of use for argocd + helmfile setups.

An example of application is to declare a configmap with custom shell scripts, so one could then reference it and mount it into mariadb pods.


mmontes11 (approved):
Neat! LGTM! Thank you very much for this contribution. This way any resource could be managed by `mariadb-cluster` helm chart in one place, and be garbage collected when uninstalling the chart.

PR #1827, opened 2026-07-08
Fix: replica cluster primary should be read-only

### LINKED ISSUE #1799: [Bug] primary replica in a multi-cluster setup is not read_only
I am experimenting with a multi-cluster setup (binlog based).

I observe that in the replica cluster, the primary node is unconditionally set to writeable (read_only=0), while my expectation would be that only the primary node of the primary cluster is writable.

The reason I think is clear (writes to non-primary nodes break replication eventually, data divergence, data loss, etc).

I think this is just hardcoded in pkg/controller/replication/topology.go unless I am missing something.

So my question is: is this intentional? Should that just be inverse, i.e. unconditionally to enable read_only?

Environment:

* mariadb-operator v26.6.0
* Replication-based multiCluster: one primary cluster + one replica cluster, each replicas: 2.

Observed via:

`SELECT @@global.read_only` returns 0 on the replica cluster's PrimaryReplica pod (-0), while on the secondary replica (-1) it returns 1.

mooo112:
Fixes #1799


mooo112:
I think that by accident `DisableReadOnly()` instead of `EnableReadOnly()` was used in `configurePrimaryReplica()`. We stumbled across this issue in a scenario (it not only happens there though), where we strictly require to only have a single writer per instance and thus added additional fencing (setting `read_only`explicitly and additionally for all non-writers) and the operator kept setting `read_only` to `OFF` after a planned site failover.

mooo112:
The proposed fix is not complete. Closing for now.

PR #1831, opened 2026-07-09
Fix: Compose multi-domain GTIDs on multiCluster demote

### LINKED ISSUE #1830: [Bug] multiCluster demote fails on multi-domain GTID position
**Description**
multiCluster demote fails when the external primary's @@gtid_binlog_pos spans multiple GTID domains (normal after a cross-site failover with log_slave_updates). composeGtids (internal/controller/mariadb_controller_multicluster.go ~L299) parses both positions with the single-GTID replication.ParseGtid, which rejects comma separated multi-domain lists. Demote reconcile then errors forever and status.currentMultiClusterPrimary never converges as expected.

**The error**
```
error composing GTIDs: error parsing external GTID 1-10-216,2-211-18359:
invalid GTID format: expected 3 parts (domain-server-sequence), got 5
```

**Expected behavior/proposed fix**
Parse both with ParseAllGtids (already used ~L208) and merge per domain (external wins). Single-domain behaviour remains unchanged. I'm gonna open a PR soon.

**Additional context**
We are running the operator on v26.6.0, multiCluster replication, MariaDB 11.4.
We are experimenting with additional multiCluster functionality on top: Site failover, Strict single writer enforcement beyond it's "nature" and related to this "write-site-awareness" for our clients.

mooo112:
Fixes #1830 

Fixes the multiCluster demote path when the external primary's `gtid_binlog_pos` spans more than one GTID domain.

We hit this during a planned multi-site failover in our experiments: after promoting the remote site and letting the old site's history replicate over (`log_slave_updates`), the external primary's `gtid_binlog_pos` legitimately became multi-domain (e.g. 1-10-216,2-211-18359). On the next demote, `composeGtids()` errored out because it parsed both positions with the single-GTID ParseGtid, which rejects the comma-separated multi-domain list, so the reconcile failed on every attempt and `status.currentMultiClusterPrimary` never converged as we would expect.

### DIFF
--- a/internal/controller/mariadb_controller_multicluster.go
+++ b/internal/controller/mariadb_controller_multicluster.go
@@ -297,13 +297,32 @@ func (r *MariaDBReconciler) shouldReconcileMultiCluster(ctx context.Context, mdb
 }
 
 func composeGtids(rawGtid, rawExternalGtid string) (string, error) {
-	gtid, err := replication.ParseGtid(rawGtid)
+	gtids, err := parseGtids(rawGtid)
 	if err != nil {
 		return "", fmt.Errorf("error parsing GTID %s: %v", rawGtid, err)
 	}
-	externalGtid, err := replication.ParseGtid(rawExternalGtid)
+	externalGtids, err := parseGtids(rawExternalGtid)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


mooo112:
In depth testing of this proposed fix revealed that this is caused by a problem on another level, which instead has to be addressed. Updated PR is following soon.

Issue #556, opened 2024-04-12
[Feature] Support for ignoring `mysql.global_priv` natively in `Backup`


mmontes11:
**Is your feature request related to a problem? Please describe.**
Galera only replicates the tables with InnoDB engine:
- https://galeracluster.com/library/kb/user-changes.html

Something which unfortunately does not include `mysql.global_priv`.  This results in some Galera `Pods` restarting after restoring a backup:
- https://github.com/mariadb-operator/mariadb-operator/issues/525

 When backing `mysql.global_priv`, a `DELETE` statement is generated in the backup, which makes all the replicas to delete the table, but the data won't be replicated, leading to authentication errors and replica `Pods` restarting due to failed liveness probes.

**Describe the solution you'd like**
The purpose of logical backups is being able to seed data in a `MariaDB` instance without depending on its physical topology i.e. we are able to port data from a `MariaDB` standalone to a `MariaDB` with Galera enabled.

For archieving this, we would need a `ignoreGlobalPriv` set to `false` by default in the `Backup` resource. When backing up a `MariaDB` with Galera enabled `mysql.global_priv` will always be ignored.

This means that non Galera `MariaDB`'s will continue to operate as usual, but we are enabling the possibility of migrating from a non Galera `MariaDB` to a `MariaDB` with Galera by enabling `ignoreGlobalPriv = true`. 

Note that, when `ignoreGlobalPriv = false`, the `mysql.global_priv` will not be backed up and therefore nor the users and its permissions. This will be handled by the `MariaDB` image entrypoint, that will take into account the `spec.rootPasswordSecretKeyRef`, `spec.username` and `spec.passwordSecretKeyRef`. In addition, users and grants created via the `User` and `Grant` CRs are continously reconciled and therefore they will be recreated.

**Describe alternatives you've considered**
Continue working as we currently do and not ignoring the `mysql.global_priv`, which results in very curbersome misbehaviour with Galera after restoring a backup.

PR #1836, opened 2026-07-12
fix: restore Galera logical backups and preserve repeatable backup args

### LINKED ISSUE #1691: [Bug] args field on backup resource ignores multiple --ignore-* options
According to the documentation of the mariadb-dump command, the `--ignore-databases`, `--ignore-tables` and `--ignore-table-data` options require multiple entries when specifying to ignore multiple databases, tables or table data.

This is no longer possible with mariadb-operator 25.10.4

**Documentation**
- [ x] I acknowledge that I have read the relevant [documentation](https://github.com/mariadb-operator/mariadb-operator/tree/main/docs).

**Describe the bug**
When specifying "duplicate" --ignore-* options in the args resource of a backup job, only one ends up in the actual job. For example, when specifying to ignore two tables, will end up ignoring only one.
```
args:
  - --single-transaction
  - --all-databases
  - --ignore-table=mysql.help_category
  - --ignore-table=mysql.help_relation
```

**Expected behaviour**
The ability to ignore multiple databases, tables or table-data when specified.

Create a backup job with multiple --ignore key/values pairs. The created job wil only contain one.

**Environment details**:
- Kubernetes version: 1.32
- Kubernetes distribution: Vanilla Kubernetes
- MariaDB Operator version: 10.25.4
- MariaDB topology: Replication
- MariaDB Server version: 11.4.10
- Install method: Helm

**Additional context**
I believe this bug is introduced by #1546 with mariadb-operator 25.10.4

In my case it happened with the `--ignore-table-data` option, but theoretically it should be the same with the `--ignore-databases` and `--ignore-tables` options too.

### LINKED ISSUE #1757: Logical backups of Galera clusters are not restorable: dump includes mysql.wsrep_* tables and restore fails with "DROP command denied to 'root'"
**Documentation**

**Describe the bug**

A logical `Backup` of a Galera cluster cannot be restored. The generated
`mariadb-dump` runs with `--all-databases` and excludes only
`--ignore-table=mysql.global_priv`. It therefore includes the Galera system
tables in the `mysql` database:

- `mysql.wsrep_cluster`
- `mysql.wsrep_cluster_members`
- `mysql.wsrep_streaming_log`
- `mysql.wsrep_allowlist`

On restore, the dump issues `DROP TABLE IF EXISTS` for these tables, which is
denied even to `root`, and the restore aborts:

```
💾 Restoring backup: /backup/.../backup.2026-06-02T08:00:00Z.sql
DROP TABLE IF EXISTS `wsrep_streaming_log`
ERROR 1142 (42000) at line 2646: DROP command denied to user 'root'@'10.42.8.68' for table `mysql`.`wsrep_streaming_log`
```

The restore Job retries and fails every time; the bootstrapping MariaDB stays in
`Restoring backup` and never becomes ready. In other words, **disaster recovery
from a default logical backup of a Galera cluster is impossible**.

The operator already has Galera awareness in the logical-dump path (it injects
`--skip-add-locks` and `--ignore-table=mysql.global_priv` for Galera, cf. #556),
so this looks like an incomplete fix — the remaining `mysql.wsrep_*` tables are
not excluded.

Note: a **physical** `PhysicalBackup` of the same cluster restores cleanly
(file-level `copy-back`, no SQL `DROP`/`CREATE`), so the issue is specific to the
logical path.

**Expected behaviour**

A logical backup of a Galera cluster should be restorable out of the box. The
dump should exclude the Galera-managed `mysql.wsrep_*` system tables (the same
way `mysql.global_priv` is already excluded), so that the restore does not
attempt a denied `DROP TABLE` on them.

1. Run a Galera `MariaDB` (e.g. 3 replicas).
2. Create a default logical `Backup` (no `spec.args`) targeting S3.
3. Bootstrap a fresh Galera `MariaDB` from it:

   ```yaml
   spec:
     bootstrapFrom:
       s3:
         bucket: ...
         prefix: ...   # the logical backup prefix
   ```

4. The restore Job fails:

   ```
   DROP TABLE IF EXISTS `wsrep_streaming_log`
   ERROR 1142 (42000) at line 2646: DROP command denied to user 'root'@'...' for table `mysql`.`wsrep_streaming_log`
   ```

   The MariaDB stays in `Restoring backup` and never becomes ready.

The generated `mariadb-dump` command can be inspected on the backup Job and
shows only `--ignore-table=mysql.global_priv`:

```bash
kubectl get job <backup-job> -o jsonpath='{.spec.template.spec.initContainers[0].args}'
```

**Debug information**
- Restore Job logs:

```bash
kubectl logs <mariadb>-restore-xxxxx --all-containers
```

**Environment details**:
- Kubernetes version: v1.36.1+rke2r1
- Kubernetes distribution: Rancher (RKE2)
- MariaDB Operator version: 26.3.0
- MariaDB topology: Galera
- MariaDB Server version: 12.2.2
- MaxScale version: n/a
- Install method: Helm

**Additional context**

A natural workaround would be to exclude the `wsrep_*` tables via `spec.args`
(multiple `--ignore-table=...`), but that is currently blocked by a separate bug
where `spec.args` silently drops all but one repeated same-name flag
(`UniqueArgs` flag-name collapse).

### LINKED ISSUE #1758: spec.args on Backup/PhysicalBackup silently drops all but one of repeated same-name flags (UniqueArgs)
**Documentation**

**Describe the bug**

When `spec.args` contains more than one entry using the *same flag name*, all but
one are silently dropped from the generated `mariadb-dump` / `mariadb-backup`
command.

Root cause: `UniqueArgs()` in `pkg/datastructures/datastructures.go`. `spec.args`
is appended as `ExtraOpts` and the merged argument list is passed through
`UniqueArgs()` (`mariadbDumpArgs` / `mariadbBackupArgs` in
`pkg/command/backup.go`). `UniqueArgs` groups arguments by flag name
(`extractFlagName` = everything before the first `=`) and keeps only **one**
occurrence per flag name — the **last** one when values differ. The unit test
`multiple value overrides keep last` documents this as intended, but it is
incorrect for *repeatable* flags such as `--ignore-table` and
`--databases-exclude`.

This has two harmful effects:
1. Repeatable flags collapse to a single value.
2. User-supplied args silently override operator-injected defaults that share the
   same flag name.

**Expected behaviour**

All `spec.args` entries are forwarded verbatim, preserving repeated flags.
Deduplication should only remove true exact-string duplicates, never collapse
distinct values of a repeatable flag.

*Logical `Backup` — repeated `--ignore-table` collapse to one:*

1. Create a `Backup` with `ignoreGlobalPriv: true` and three `--ignore-table` args:

   ```yaml
   spec:
     ignoreGlobalPriv: true
     args:
       - --ignore-table=wordpress.argtest_one
       - --ignore-table=wordpress.argtest_two
       - --ignore-table=wordpress.argtest_three
   ```

2. Inspect the generated command:

   ```bash
   kubectl get job <backup-job> -o jsonpath='{.spec.template.spec.initContainers[0].args}'
   ```

3. Observed `mariadb-dump` (trimmed):

   ```
   mariadb-dump ... --all-databases --skip-add-locks ... --ignore-table=wordpress.argtest_three > ...
   ```

   Only the last entry survives. `argtest_one` and `argtest_two` are gone — and
   the `--ignore-table=mysql.global_priv` that `ignoreGlobalPriv: true` injects is
   *also* dropped, because it shares the `--ignore-table` flag name. So combining
   `ignoreGlobalPriv` with any custom `--ignore-table` silently disables the
   global_priv exclusion.

*`PhysicalBackup` — user `--databases-exclude` drops the built-in `lost+found`:*

1. Create a `PhysicalBackup` with:

   ```yaml
   spec:
     args:
       - --databases-exclude=mysql
   ```

2. Observed `mariadb-backup` (trimmed):

   ```
   mariadb-backup ... --backup --stream=xbstream --databases-exclude=mysql ...
   ```

   The hardcoded `--databases-exclude='lost+found'` has been silently removed.

**Debug information**

```bash
kubectl get job <backup-job> -o jsonpath='{.spec.template.spec.initContainers[0].args}'
```

**Environment details**:
- Kubernetes version: v1.36.1+rke2r1
- Kubernetes distribution: Rancher (RKE2)
- MariaDB Operator version: 26.3.0
- MariaDB topology: Galera
- MariaDB Server version: 12.2.2
- MaxScale version: n/a
- Install method: Helm

**Additional context**

`UniqueArgs` / `extractFlagName`: `pkg/datastructures/datastructures.go`, reached
via `ds.UniqueArgs(ds.Merge(...))` in `mariadbDumpArgs`, `mariadbBackupArgs` and
`mariadbArgs` (`pkg/command/backup.go`). This bug also blocks the obvious
workaround for #1757 (logical Galera backups are not restorable): excluding the
`mysql.wsrep_*` tables would require multiple `--ignore-table` args, which this
bug collapses to one.

ioanalytica:
Close MDB-17

## Summary

Fixes two coupled bugs that make disaster recovery from a logical `Backup` of a Galera cluster impossible.

### #1757 — Galera logical backups are not restorable

A logical dump runs with `--all-databases` and includes the Galera-managed `mysql.wsrep_*` system tables. On restore the dump issues `DROP TABLE IF EXISTS` for them, which Galera denies even to `root`, so the restore aborts and the `MariaDB` never becomes ready:

```text
ERROR 1142 (42000) at line 2646: DROP command denied to user 'root'@'...' for table `mysql`.`wsrep_streaming_log`
```

We now exclude `mysql.wsrep_cluster`, `mysql.wsrep_cluster_members`, `mysql.wsrep_streaming_log` and `mysql.wsrep_allowlist` from the dump — the same way `mysql.global_priv` already is (#556). Galera recreates these tables on bootstrap, so nothing is lost.

### #1758 / #1691 — `spec.args` dropped repeated same-name flags

`UniqueArgs` collapsed repeated same-name flags to a single value, so multiple `--ignore-table` args (and the operator's own injected excludes) were silently dropped — which also blocked the fix for #1757. `UniqueArgs` now preserves every distinct occurrence of the repeatable `mariadb-dump` flags (`--ignore-table`, `--ignore-table-data`), collapsing only exact duplicates; scalar flags keep their last-wins override. This builds on the approach proposed by @SAY-5 in #1721 (credited as co-author).

Additionally, `mariadb-backup`'s `--databases-exclude` is a single-value, last-wins flag (accepting a space-separated list), so a user-supplied `--databases-exclude` silently dropped the built-in `lost+found` exclusion. Operator default and user values (in both `--databases-exclude=<v>` and `--databases-exclude <v>` forms) are now merged into a single, shell-escaped flag.

Closes #1757.
Closes #1758.
Closes #1691.

Supersedes #1721.

## Test plan

- New/updated unit tests in `pkg/datastructures` (repeatable flags preserved, operator + user `--ignore-table` coexist) and `pkg/command` (Galera `wsrep_*` excludes, `--databases-exclude` merge incl. space form and single-quote escaping).
- `go test ./pkg/command/... ./pkg/datastructures/...`, `go vet`, `gofmt`, `go build ./...` — all green.
- No API/CRD changes, so no codegen diff.

### DIFF
--- a/docs/logical_backup.md
+++ b/docs/logical_backup.md
@@ -545,6 +545,18 @@ Also, to avoid situations where `mysql.global_priv` is unreplicated, all the ent
 - Rely on the [`User`](https://github.com/mariadb-operator/mariadb-operator/blob/main/examples/manifests/user.yaml) and [`Grant`](https://github.com/mariadb-operator/mariadb-operator/blob/main/examples/manifests/grant.yaml) CRs to create additional users and grants. Refer to the [SQL resource documentation](./sql_resources.md) for further detail.
 
 
+#### `mysql.wsrep_*`
+
+Galera manages its own set of system tables in the `mysql` database: `mysql.wsrep_cluster`, `mysql.wsrep_cluster_members`, `mysql.wsrep_streaming_log` and `mysql.wsrep_allowlist`. These tables hold Galera cluster runtime state and are recreated by Galera when a node bootstraps.
+
+A logical dump taken with `--all-databases` includes these tables, so the backup file contains a `DROP TABLE IF EXISTS` statement for each of them. On restore, Galera denies `DROP` on the `wsrep_*` tables even to `root`, so the restore aborts:
+
+```text
+ERROR 1142 (42000) at line 2646: DROP command denied to user 'root'@'...' for table `mysql`.`wsrep_streaming_log`
+```
+
+This makes disaster recovery from a default logical backup of a Galera cluster impossible. To address this, when backing up `MariaDB` instances with Galera enabled, the operator automatically excludes these tables from the dump using the `--ignore-table` option with `mariadb-dump`. Since Galera recreates them on bootstrap, no cluster state is lost.
+
 #### `LOCK TABLES` 
 
 Galera is not compatible with the `LOCK TABLES` statement:
@@ -566,7 +578,7 @@ mariadb-dump --user=${MARIADB_USER} --password=${MARIADB_PASSWORD} --host=${MARI
 > If you are using Galera or planning to migrate to a Galera instance, make sure you understand the [Galera backup limitations](#galera-backup-limitations) and use the following command instead:
 
 ```bash
-mariadb-dump --user=${MARIADB_USER} --password=${MARIADB_PASSWORD} --host=${MARIADB_HOST} --single-transaction --events --routines --all-databases --skip-add-locks --ignore-table=mysql.global_priv > backup.2024-08-26T12:24:34Z.sql
+mariadb-dump --user=${MARIADB_USER} --password=${MARIADB_PASSWORD} --host=${MARIADB_HOST} --single-transaction --events --routines --all-databases --skip-add-locks --ignore-table=mysql.global_priv --ignore-table=mysql.wsrep_cluster --ignore-table=mysql.wsrep_cluster_members --ignore-table=mysql.wsrep_streaming_log --ignore-table=mysql.wsrep_allowlist > backup.2024-08-26T12:24:34Z.sql
 ```
 
 2. Ensure that your backup file is named in the following format: `backup.2024-08-26T12:24:34Z.sql`. If the file name does not follow this format, it will be ignored by the operator.
--- a/docs/logical_backup.md
+++ b/docs/logical_backup.md
@@ -557,6 +557,9 @@ ERROR 1142 (42000) at line 2646: DROP command denied to user 'root'@'...' for ta
 
 This makes disaster recovery from a default logical backup of a Galera cluster impossible. To address this, when backing up `MariaDB` instances with Galera enabled, the operator automatically excludes these tables from the dump using the `--ignore-table` option with `mariadb-dump`. Since Galera recreates them on bootstrap, no cluster state is lost.
 
+> [!IMPORTANT]
+> Galera logical backups taken *before* this fix still contain the `mysql.wsrep_*` tables and remain unrestorable. Retake them once the operator includes this change.
+
 #### `LOCK TABLES` 
 
 Galera is not compatible with the `LOCK TABLES` statement:
--- a/pkg/command/backup.go
+++ b/pkg/command/backup.go
@@ -18,6 +18,16 @@ import (
 	"k8s.io/utils/ptr"
 )
 
+// galeraSystemTables are the mysql.wsrep_* system tables managed by Galera. They are
+// excluded from logical dumps because Galera denies DROP on them even to root, which
+// would otherwise make a restore fail. See mariadbDumpArgs and issue #1757.
+var galeraSystemTables = []string{
+	"mysql.wsrep_cluster",
+	"mysql.wsrep_cluster_members",
+	"mysql.wsrep_streaming_log",
+	"mysql.wsrep_allowlist",
+}
+
 type BackupOpts struct {
 	CommandOpts
 	Path                 string
@@ -572,6 +582,15 @@ func (b *BackupCommand) mariadbDumpArgs(backup *mariadbv1alpha1.Backup, mariadb 
 	// LOCK TABLES is not compatible with Galera: https://mariadb.com/kb/en/lock-tables/#limitations
 	if mariadb.IsGaleraEnabled() {
 		args = append(args, "--skip-add-locks")
+		// Galera manages the mysql.wsrep_* system tables and denies DROP on them even to
+		// root. A logical dump emits DROP TABLE IF EXISTS for them, so the restore aborts
+		// with "DROP command denied to user 'root'" and the MariaDB never becomes ready.
+		// Exclude them so a logical backup is restorable out of the box; Galera recreates
+		// these tables on bootstrap, so nothing is lost.
+		// See: https://github.com/mariadb-operator/mariadb-operator/issues/1757
+		for _, table := range galeraSystemTables {
+			args = append(args, fmt.Sprintf("--ignore-table=%s", table))
+		}
 	}
 	// Galera only replicates InnoDB tables and mysql.global_priv uses the MyISAM engine.
 	// Ignoring this table enables a clean restore without replicas getting restarted
@@ -629,12 +648,25 @@ func (b *BackupCommand) mariadbBackupArgs(mariadb *mariadbv1alpha1.MariaDB, targ
 	backupOpts := make([]string, len(b.ExtraOpts))
 	copy(backupOpts, b.ExtraOpts)
 
+	// mariadb-backup's --databases-exclude takes a single space-separated value and is
+	// last-wins (not repeatable), so a user-supplied --databases-exclude would silently
+	// drop the built-in lost+found exclusion. Merge both into a single flag instead (#1758).
+	// The ext4 filesystem creates a lost+found directory by default, which causes
+	// mariadb-backup to include it in the backup file as a database.
+	databasesExclude := []string{"lost+found"}
+	for _, opt := range backupOpts {
+		if isDatabasesExcludeOpt(opt) {
+			if value := databasesExcludeValue(opt); value != "" {
+				databasesExclude = append(databasesExclude, value)
+			}
+		}
+	}
+	backupOpts = ds.Remove(backupOpts, isDatabasesExcludeOpt)
+
 	args := []string{
 		"--backup",
 		"--stream=xbstream",
-		// The ext4 filesystem creates a lost+found directory by default,
-		// which causes mariadb-backup to include it in the backup file as a database.
-		"--databases-exclude='lost+found'",
+		fmt.Sprintf("--databases-exclude=%s", shellSingleQuote(strings.Join(databasesExclude, " "))),
 	}
 	if mariadb.IsTLSEnabled() {
 		args = append(args, b.tlsArgs(mariadb)...)
@@ -648,6 +680,31 @@ func (b *BackupCommand) mariadbBackupArgs(mariadb *mariadbv1alpha1.MariaDB, targ
 	}
 
 	return ds.UniqueArgs(ds.Merge(args, backupOpts)...)
+}
+
+// isDatabasesExcludeOpt reports whether opt is a --databases-exclude flag carrying a
+// value, in either the --databases-exclude=<value> or space-separated
+// --databases-exclude <value> form (mariadb-backup accepts both). A bare
+// --databases-exclude with the value in a separate argv element is intentionally not
+// matched, so it is left untouched rather than orphaning its value.
+func isDatabasesExcludeOpt(opt string) bool {
+	opt = strings.TrimSpace(opt)
+	return strings.HasPrefix(opt, "--databases-exclude=") || strings.HasPrefix(opt, "--databases-exclude ")
+}
+
+// databasesExcludeValue extracts the value of a --databases-exclude option, supporting
+// both the --databases-exclude=<value> and space-separated --databases-exclude <value>
+// forms and stripping any surrounding single or double quotes.
+func databasesExcludeValue(opt string) string {
+	rest := strings.TrimSpace(strings.TrimPrefix(strings.TrimSpace(opt), "--databases-exclude"))
+	rest = strings.TrimSpace(strings.TrimPrefix(rest, "="))
+	return strings.Trim(rest, `'"`)
+}
+
+// shellSingleQuote wraps s in single quotes safe for a POSIX shell, escaping any embedded
+// single quotes so user-provided values cannot break out of the generated bash command.
+func shellSingleQuote(s string) string {
+	return "'" + strings.ReplaceAll(s, "'", `'\''`) + "'"
 }
 
 func (b *BackupCommand) mariadbRestoreArgs(restore *mariadbv1alpha1.Restore, mariadb interfaces.TLSProvider) []string {
--- a/pkg/command/backup.go
+++ b/pkg/command/backup.go
@@ -18,6 +18,16 @@ import (
 	"k8s.io/utils/ptr"
 )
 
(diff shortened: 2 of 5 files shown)


ioanalytica:
Hi @mmontes11 — a friendly nudge on this one when you have a moment. It fixes a critical disaster-recovery bug: logical backups of Galera clusters currently can't be restored (#1757 — the dump includes the `mysql.wsrep_*` tables, so the restore hits `DROP command denied to 'root'` and the MariaDB never becomes ready). It also closes #1758 and #1691 and supersedes the idle #1721 (with @SAY-5 credited as co-author). The CodeRabbit review is clean and unit tests are green; it touches only `pkg/command` and `pkg/datastructures` — no API/CRD changes. Could you enable CI and take a look? Happy to adjust anything.

mmontes11:
## 🔍 PR Review

Fixes two coupled backup bugs: (1) logical dumps of Galera clusters include the Galera-managed `mysql.wsrep_*` tables, so restore always aborts with `DROP command denied` (#1757); (2) `UniqueArgs` collapsed repeated `--ignore-table` flags, and a user-supplied `--databases-exclude` silently replaced the built-in `lost+found` exclusion (#1758/#1691).

### Verdict at a glance

| Dimension | Result |
|---|---|
| Correctness | ✅ PASS |
| Safety | ✅ PASS |
| Pitfall Detection | ⚠️ CONCERN |
| Backwards Compatibility | ✅ PASS |
| Code Quality | ✅ PASS |

**Overall:** 📝 Approve with notes — **Risk:** 🟡 MEDIUM

---

### ✅ Correctness

The arg-building logic is sound across all three paths; the wsrep exclusions mirror the existing `global_priv` handling, and the quoting is correct for the bash execution path.

### ✅ Safety

Only backup-job command construction changes — no StatefulSet pod template, no HA orchestration, no credential handling; shell-escaping user-supplied `--databases-exclude` values is an improvement over the previous raw pass-through.

### ⚠️ Pitfall Detection

One residual edge case in the `--databases-exclude` merge: a value passed as **two separate argv elements** (`--databases-exclude` followed by `mysql`) does not match `isDatabasesExcludeOpt` (pkg/command/backup.go:691-693), so it survives `ds.Remove` and `UniqueArgs` then keeps the bare user flag (last-wins) and drops the merged `--databases-exclude='lost+found ...'` (pkg/command/backup.go:682). Net effect: `lost+found` silently re-enters the backup — the same silent-drop class this PR fixes. This is pre-PR behavior for that input shape and it's rare (most users write the list as a single `spec.args` entry), so it's non-blocking, but worth handling or explicitly documenting.

Otherwise clean: unit tests cover all new behaviors and pass locally (`go test ./pkg/command/... ./pkg/datastructures/...`, `go vet`, `gofmt`, `go build ./...` all green), the docs update sits under the existing Galera backup limitations section, and there are no CRD/RBAC/codegen concerns.

### ✅ Backwards Compatibility

All changes are intended behavioral changes, none breaking:

- **Behavioral:** Galera logical dumps no longer include the `mysql.wsrep_*` tables (fixes #1757).
- **Behavioral:** Distinct repeated `--ignore-table`/`--ignore-table-data` values are preserved instead of collapsing to the last one (fixes #1691/#1758).
- **Behavioral:** User `--databases-exclude` is merged with the built-in `lost+found` exclusion instead of overriding it (fixes #1758).

No API/CRD changes; existing CRs are unaffected on upgrade.

### ✅ Code Quality

Minimal, pattern-following change (the wsrep list mirrors the #556 `global_priv` handling), small well-commented helpers with issue links, and table-driven tests matching the existing suite style. `shellSingleQuote` is the right call over adding a shell-escaping dependency.

---

### 📋 Suggested before merge

- [ ] Handle or explicitly document the two-element `--databases-exclude <value>` form (bare flag + value as separate `spec.args` entries) in `isDatabasesExcludeOpt` (pkg/command/backup.go:691-693) — as-is, `lost+found` silently drops out of the exclusion in that case.
- [ ] Optional: a one-liner in `docs/logical_backup.md` noting that Galera logical backups taken **before** this fix still contain the `wsrep_*` tables and remain unrestorable, so they should be retaken.

Note: CI has not run on this fork yet (`mergeable_state: blocked`); the branch merges cleanly into current `main` (verified via `git merge-tree`).

ioanalytica:
Thanks for the thorough review, @mmontes11! I've addressed both suggestions from your list:

- **Two-element `--databases-exclude` form:** the bare `--databases-exclude` with its value as a separate `spec.args` entry is now merged as well, so `lost+found` no longer silently drops out in that case. Excluded database names are also deduplicated (order-preserving) so redundant entries don't grow the command.
- **Docs:** added a note that Galera logical backups taken *before* this fix still contain the `mysql.wsrep_*` tables and should be retaken.

I also rebased onto current `main` (the branch was behind the zstd compression change) — it's now up to date and conflict-free. Ready for another look whenever you have a moment.

mmontes11:
First, thanks to @ioanalytica for spotting this and putting together the fix — the #1757 report and the PR's dump-exclusion approach are exactly the right response to the server-side restriction described below.

## Findings: the MariaDB server change (10.11 reference)

MariaDB **10.11** introduced a hard-coded guard in the server's `DROP TABLE` handler that unconditionally refuses to drop `mysql.wsrep_streaming_log` while Galera is active — regardless of privileges. Root with `ALL PRIVILEGES` still gets a **fake access-denied error** (1142), which is what makes it look like a permissions problem.

The guard, in the `SQLCOM_DROP_TABLE` path — [sql/sql_parse.cc#L5067-L5079](https://github.com/MariaDB/server/blob/10.11/sql/sql_parse.cc#L5067-L5079):

```c 
#ifdef WITH_WSREP
      /* In Galera do not allow dropping mysql.wsrep_streaming_log
         because it would make streaming replication to fail.
      */
      if (WSREP(thd) && wsrep_is_streaming_log(all_tables))
      {
        my_error(ER_TABLEACCESS_DENIED_ERROR, MYF(0), "DROP", ...);
        goto error;
      }
#endif /* WITH_WSREP */
```

Supporting pieces in 10.11:

• `wsrep_is_streaming_log()` matches exactly one table — `mysql.wsrep_streaming_log` — [sql/sql_parse.cc#L1155-L1166](https://github.com/MariaDB/server/blob/10.11/sql/sql_parse.cc#L1155-L1166)

• `WSREP(thd)` is true only when the Galera provider is loaded **and** `wsrep_on=ON` — [sql/wsrep_on.h#L31-L35](https://github.com/MariaDB/server/blob/10.11/sql/wsrep_on.h#L31-L35)

Consequences:

• The check runs *after* `check_table_access()`, so privilege grants are irrelevant — no grant combination can bypass it.

• Only `mysql.wsrep_streaming_log` is protected; the other `mysql.wsrep_*` tables can be dropped even on a live Galera cluster.

• Prior to 10.11 this block did not exist, which is why the same `DROP TABLE IF EXISTS mysql.wsrep_streaming_log;` statement in a dump succeeded on older targets and aborts restores on 10.11+ targets.

## Test results

**1) Live DROP test on 10.11.19** (single-node Galera, `wsrep_ready=ON`, provider `Galera`, connected as root):

Statement | Result on 10.11.19

--- | ---

`DROP TABLE mysql.wsrep_cluster` | OK

`DROP TABLE mysql.wsrep_cluster_members` | OK

`DROP TABLE mysql.wsrep_allowlist` | OK

`DROP TABLE mysql.wsrep_streaming_log` | **ERROR 1142 (42000): DROP command denied to user 'root'@'localhost'**

**2) End-to-end KIND reproduction of #1757 on MariaDB 10.11** (KIND, operator built from source, 3-replica Galera clusters, all running **10.11.19**):

• **Bug (current main `ca6d54f9`)**: `Backup` of a Galera cluster produces a dump containing `DROP TABLE IF EXISTS` + `CREATE TABLE` for all four `mysql.wsrep_*` tables. Restoring that dump into a fresh Galera cluster aborts in the restore job:
```
ERROR 1142 (42000) at line 4680: DROP command denied to user 'root'@'10.244.0.65'
for table `mysql`.`wsrep_streaming_log`
```
The restore job goes to Error/BackOff and the restore never completes.

• **Fix (PR head `720d003c`)**: dump now contains **zero** `wsrep_*` DDL (the only "wsrep" occurrences left are help-text strings in `mysql.help_*` rows); restore job completes, the fresh cluster reaches `Ready=True`, `appdb.orders` is queryable intact, and all four `mysql.wsrep_*` tables remain present on the restored nodes.

## How PR #1836 addresses it

The PR excludes the four Galera-managed system tables (`wsrep_cluster`, `wsrep_cluster_members`, `wsrep_streaming_log`, `wsrep_allowlist`) from Galera logical dumps via injected `--ignore-table` flags — the same pattern as the existing `mysql.global_priv` exclusion (#556). Since the server will never allow these tables back in via DDL on 10.11+, keeping them out of dumps is the correct and durable fix; the KIND reproduction above proves it end-to-end against the real failure mode.

One operational note (already covered by the PR's `docs/logical_backup.md` addition): dumps taken with pre-fix operator versions still contain the wsrep DDL and must be retaken or scrubbed.

mmontes11:
@ioanalytica please fix the linter and we should be good to go

ioanalytica:
Pushed a fix for the **Lint** failure — a `misspell` nit (British "honour" in a comment).

The **Artifacts** failure looks unrelated to this PR. `make gen` regenerates `docs/api_reference.md` with `compressionThreads` / the `zstd` enum and `semiSyncWaitNoSlave` — all from already-merged features whose fields aren't reflected in the committed `docs/api_reference.md` on `main` (the Go types have them, the doc doesn't; likely last regenerated under a `-dev` VERSION, which skips the docs target). So any branch rebased onto current `main` trips this check.

I've left it out of this PR to avoid bundling unrelated generated-doc changes here — probably cleanest to regenerate `docs/api_reference.md` on `main` directly. But I'm happy to include the regenerated file in this PR instead if you'd prefer; just let me know.

mmontes11 (approved):
LGTM. Thank you very much for this contribution! 🙏🏻 🦭

PR #1851, opened 2026-07-21
refactor: Removed `event_actions` altogether

Michaelpalacce:
Event Actions are used to aggregate events. Using the same event action leads to very little events actually showing. This should fix that.


mmontes11:
## 🔍 PR Review

Removes `api/v1alpha1/event_actions.go` (the `ActionReconciling` / `ActionArchiving` constants) and replaces every `action` argument in `Eventf(...)` calls with the same `Reason*` constant already passed for `reason`. Motivated as a fix for event over-aggregation. Porting is clean — no dangling references remain and no generated artifacts are implicated.

### Verdict at a glance

| Dimension | Result |
|---|---|
| Correctness | ⚠️ CONCERN |
| Safety | ✅ PASS |
| Pitfall Detection | ✅ PASS |
| Backwards Compatibility | ⚠️ CONCERN |
| Code Quality | ✅ PASS |

**Overall:** 🔧 Request changes — **Risk:** 🟢 LOW

---

### ⚠️ Correctness

The refactor is clean and complete (no remaining `ActionReconciling`/`ActionArchiving`/`event_actions` references), but **it does not achieve its stated goal** of reducing event over-aggregation.

Event deduplication in `k8s.io/client-go/tools/events` (the package this repo uses) keys the isomorphic event cache on `getKey()`, whose `eventKey` struct already includes **both** `action` **and** `reason` (plus `eventType`, `reportingController`, `regarding`, `related`). The `note` (message) is **not** part of the key.

Consequences:
- Since `reason` already varied per call site, and the key already contained `reason`, setting `action = reason` adds **no distinguishing power** — the set of unique keys is identical before and after. Aggregation behavior is unchanged.
- The actual symptom described ("very little events showing") is caused by events that share `reason` + `regarding` but differ only in `note` collapsing into a single series (e.g. `ReasonPrimarySwitching` "from 0 to 1" vs "from 1 to 2" in `pkg/controller/galera/controller.go`, or the two `ReasonMaintenance` sites in `pkg/controller/maintenance/read_only.go`). Because `note` is not in the key at all, this PR — which only touches `action` — cannot separate them.

Failure scenario: an operator merges this expecting distinct switchover/maintenance events to stop collapsing; after deploying, `kubectl get events` shows the same single series-with-count as before. The reported problem persists.

If the goal really is to surface distinct-message events separately, `action` is the wrong lever — the message would need to become part of the key, which this API deliberately does not support. As a pure cleanup (dropping two rarely-meaningful constants) the change is harmless, but it should not be framed or merged as a bug fix.

### ✅ Safety

Observational-only change; touches no HA sequencing, storage, or Pod template.

### ✅ Pitfall Detection

No gotchas tripped — no CRD/RBAC/StatefulSet-template surface involved.

### ⚠️ Backwards Compatibility

- **Behavioral (low risk):** the `action` field value on emitted Events changes from `Reconciling`/`Archiving` to the reason string. Kubernetes Events are not a stable API contract, but any user alerting/filtering on `action == "Reconciling"` would be affected. Worth a line in release notes.
- **Additive/neutral:** removing the `Action*` constants is safe — they are internal Go identifiers, not CRD fields, so no generated artifacts or stored-object compatibility are involved.

### ✅ Code Quality

Consistent with surrounding call sites; removing the file is a reasonable simplification. AGENTS.md is updated to match the new convention.

---

### 📋 Suggested before merge

- [ ] Re-verify the over-aggregation symptom against client-go's `getKey` (`eventKey` already includes `reason`) — confirm whether this change measurably changes `kubectl get events` output. It appears to be a no-op for aggregation.
- [ ] If it is a no-op, reframe the PR as a cleanup/simplification rather than a fix, and update the title/description accordingly.
- [ ] Note the `action`-field value change in release notes for anyone filtering Events on `action`.
- [ ] Consider whether the `action` dimension carried any intended value (distinct from `reason`) that is now lost by making them identical.

mmontes11 (approved):

Issue #1499, opened 2025-11-02
[Bug] Getting stuck on `Switching primary to`


hydazz:
**Documentation**
- [x] I acknowledge that I have read the relevant [documentation](https://github.com/mariadb-operator/mariadb-operator/tree/main/docs).

**Describe the bug**

Out of the blue, twice my mariadb replica has gotten stuck in `Switching primary to 'mariadb-x'`. Only way I could find to resolve it is to destroy and restore from backup, unsure if I'm using replicas incorrectly, but it worked fine with galera. No pods were deleted or anything that would case a switch over, so unsure why its trying to switch over in the first place.

**Expected behaviour**

No switchover/switchover actually switches over...

**Debug information**
```
NAMESPACE   NAME      READY   STATUS                             PRIMARY     UPDATES                    AGE
databases   mariadb   False   Switching primary to 'mariadb-1'   mariadb-0   ReplicasFirstPrimaryLast   36h
```

```
Name:         mariadb
Namespace:    databases
Labels:       kustomize.toolkit.fluxcd.io/name=mariadb-cluster
              kustomize.toolkit.fluxcd.io/namespace=databases
Annotations:  <none>
API Version:  k8s.mariadb.com/v1alpha1
Kind:         MariaDB
Metadata:
  Creation Timestamp:  2025-11-01T00:58:17Z
  Generation:          5
  Resource Version:    84366639
  UID:                 a1abd8e8-febc-439e-a57f-f99cffd85568
Spec:
  Bootstrap From:
    Backup Content Type:  Physical
    s3:
      Access Key Id Secret Key Ref:
        Key:     B2_ACCESS_KEY_ID
        Name:    mariadb-secret
      Bucket:    b2backup-k8s
      Endpoint:  s3.us-east-005.backblazeb2.com
      Prefix:    mariadb
      Secret Access Key Secret Key Ref:
        Key:   B2_SECRET_ACCESS_KEY
        Name:  mariadb-secret
      Tls:
        Enabled:  true
    Staging Storage:
      Persistent Volume Claim:
        Access Modes:
          ReadWriteOnce
        Resources:
          Requests:
            Storage:         4Gi
        Storage Class Name:  openebs-hostpath
    Volume:
      Persistent Volume Claim:
        Claim Name:  mariadb-physicalbackup-staging
[...]

Issue #1719, opened 2026-04-29
[Bug] Unexpected read_only=0 on a node demoted from primary to replica after failover


rgambelli:
After the initial deployment, the cluster behaved as expected:

Primary node → read_only = 0
Replica nodes → read_only = 1

However, after some event which I'm not able to explain what:

The operator promoted one of the replicas to primary.
The former primary was reconfigured as a replica.

Replication is now working correctly and the topology is consistent (i.e., SHOW SLAVE STATUS shows proper replication on the demoted node).

Issue

The node that was originally primary and then demoted to replica still has:

SELECT @@read_only;

returning:

0

So it is now functioning as a replica, but it remains writable.

Expected behavior

After a failover where a primary is demoted to replica, I would expect the operator to:

Set read_only = 1 on the demoted node
Ensure replicas are not writable by default
Actual behavior
The demoted node is correctly configured as a replica.
Replication is running normally.
read_only remains 0.
Questions
Is this the intended behavior?
Should the operator enforce read_only=1 on replicas after demotion?
Is there a recommended way to ensure replicas are always read-only in this setup?

From a safety perspective, leaving a replica writable could potentially allow accidental writes and cause data divergence.

Thanks for your help!

**Environment details**:
- MariaDB version: 10.x
- Deployment: Kubernetes
- Operator: mariadb-operator
- Topology: 1 primary + 2 replicas

**Additional context**

hedgieinsocks:
Readonly is being enabled [here](https://github.com/mariadb-operator/mariadb-operator/blob/84fbf4e47da650b229adb539612ae902f89460fd/pkg/controller/replication/config.go#L128) which is run during switchover [here](https://github.com/mariadb-operator/mariadb-operator/blob/84fbf4e47da650b229adb539612ae902f89460fd/pkg/controller/replication/switchover.go#L370) for replicas and [here](https://github.com/mariadb-operator/mariadb-operator/blob/84fbf4e47da650b229adb539612ae902f89460fd/pkg/controller/replication/switchover.go#L426) for demoted master. And per my check it seems to work as expected.

The issue occurs whenever a replica container is restarted. In other words, after a swithover, new replicas are set to be readonly as they should. But if you delete a replica pod at any point, it will come back without readonly.

[...]

PR #1853, opened 2026-07-21
fix: prevent replication switchover from deadlocking with the primary locked

### LINKED ISSUE #1852: [Bug] Switchover deadlocks holding the primary read-only forever when a replica is wired to the not-yet-promoted new primary (no overall deadline, no abort)
**Documentation**

**Describe the bug**

During an automated `ReplicasFirstPrimaryLast` update (triggered by the operator's own TLS certificate renewal), the pending switchover (`spec.replication.primary.podIndex: 0`, current primary `mariadb-1`) deadlocked for ~3 hours and held the current primary **read-only the whole time** — a full write outage for the application — until we manually intervened (`spec.suspend: true`, `SET GLOBAL read_only=OFF`, manual reseed of the stuck replica, unsuspend).

The deadlock mechanism, confirmed from server state:

- Replica `mariadb-2` was configured to replicate from **`mariadb-0` — the switchover *target*, which had not been promoted yet** (`Master_Host: mariadb-0.mariadb-internal...`).
- `mariadb-0` (still a plain replica of `mariadb-1` at that point) had an **empty `gtid_binlog_pos`** — fresh pod, and replicas don't run with `log_slave_updates` — so it had nothing to serve. `mariadb-2` therefore reported `Slave_IO_Running: Yes`, `Slave_SQL_Running: Yes`, `Seconds_Behind_Master: 0`, **no errors**, while being permanently stale (stuck at a GTID far behind the primary).
- `waitForReplicaSync` (switchover phase "Wait sync") waits for **every** replica to reach the current primary's `gtid_binlog_pos`. `mariadb-2` could never receive those events over its replication connection, so every reconcile failed with `Error waiting for GTID '0-11-…' in replica '2': timeout waiting for replica to be synced` and was retried — every ~35 min, indefinitely.
- On each retry, the earlier phases re-run first: "Lock primary with read lock" and "Set read_only in primary" re-apply the read-only state to the current primary. There is **no overall switchover deadline and no abort/rollback path** that releases the primary when the sync wait cannot make progress, so the primary stays read-only forever.

Two contributing design issues, either of which would have prevented the outage:

1. **The sync wait doesn't verify what the replica is actually replicating from.** It waits for the current primary's GTID on a connection that may point elsewhere (here: at the unpromoted switchover target). Checking `Master_Host`/connection health against the *expected* source before waiting — or treating "replica connected to the wrong host" as a reconfigure-first condition — would surface the misconfiguration instead of waiting on it.
2. **`read_only` on the primary is held across unbounded retries.** A switchover that cannot complete within some overall deadline should release the read lock / read_only (abort and surface an error condition) rather than keep the application down. The per-replica `syncTimeout` (10s) only bounds a single attempt, not the overall operation.

How `mariadb-2` came to point at the unpromoted target: the pod was recreated ~3 minutes before the switchover began (same rolling update; `ReplicaRecovered=True` was also set, so a replica recovery ran the same night). Whichever path configured its replication during the *pending* switchover wired it to the **desired** primary (spec `podIndex`) instead of the **current** one (`status.currentPrimaryPodIndex`). We could not pin the exact reconciler from the outside, but the resulting state was stable and observable for hours.

This looks closely related to #1499 ("Getting stuck on `Switching primary to`"), which was closed by the stale-bot without a fix — notably, a commenter there also observed a replica replicating from an unexpected host during the stuck switchover.

**Expected behaviour**

- A switchover that cannot complete does not hold the primary read-only indefinitely; it aborts after a bounded time, restores writability, and surfaces a status condition/event.
- The replica-sync wait detects a replica whose replication connection does not point at the current primary (or is connected but receiving nothing) and reconfigures/fails fast instead of waiting on a GTID that can never arrive.

1. Replication cluster (3 nodes, semi-sync, `gtidStrictMode`), `updateStrategy: ReplicasFirstPrimaryLast`, pending primary switchover (`spec.replication.primary.podIndex` != current primary) as part of a rolling update.
2. Arrange for a replica to be (re)configured during the pending switchover so that it replicates from the switchover target instead of the current primary (in our case this happened organically when the replica pod was recreated by the same rolling update).
3. Observe: `ReplicaSync` events repeating with `ReplicaSyncErr … timeout waiting for replica to be synced`, current primary read-only, replica reporting healthy/0-behind while permanently stale. State persists indefinitely.

**Debug information**

- mariadb-operator version: 25.10.4 (Helm chart), MariaDB 11.8.2, 3-node semi-sync replication, `gtidStrictMode: true`, `syncTimeout: 10s`, operator-issued TLS. The switchover phase logic at current `main` (`pkg/controller/replication/switchover.go`) appears unchanged with respect to the behaviours described above.
- Event sequence (repeating, ~every 35 min for ~3 h):
  ```
  Normal   PrimaryReadonly  Enabling readonly mode in primary
  Normal   ReplicaSync      Waiting for replicas to be synced with primary
  Warning  ReplicaSyncErr   Error waiting for GTID '0-11-2749535' in replica '2': timeout waiting for replica to be synced
  Normal   PrimaryLock      Locking primary with read lock
  ```
- Stuck replica (`mariadb-2`) at the time:
  ```
  Master_Host: mariadb-0.mariadb-internal.<ns>.svc.cluster.local   # switchover target, not yet promoted
  Slave_IO_Running: Yes / Slave_SQL_Running: Yes / Seconds_Behind_Master: 0 / no errors
  Gtid_IO_Pos: 0-10-2440029                                        # far behind; never advances
  ```
- Switchover target (`mariadb-0`) at the time: healthy replica of the current primary, fully synced (`Gtid_IO_Pos: 0-11-2749538`), `@@gtid_binlog_pos` **empty**.
- After manually reseeding/repointing the replica at the current primary and resuming reconciliation, the same switchover completed cleanly in ~10 seconds of read-only.

colans:
Closes #1852

## Problem

During a pending switchover, a replica whose replication connection points at a host other than the current primary — in our production incident, the new primary *before* promotion, wired there while a rolling update recreated the pod — reports `Slave_IO_Running: Yes`, `Seconds_Behind_Master: 0` and no errors, yet can never reach the GTID that the `Wait sync` phase waits for: the unpromoted target has nothing to relay (no `log_slave_updates`). Every reconciliation then times out in `waitForReplicaSync`, and the retry re-applies the read lock and `read_only` to the current primary. Net effect: the application loses writes indefinitely (2h20m in our case, until manual intervention) while the cluster looks healthy by every per-replica metric. #1499 appears to be the same failure mode.

## Fix

Two independent changes, either of which is sufficient to end the write outage; together they make the switchover both convergent and bounded:

1. **`Wait sync` verifies each replica's replication source first.** Before waiting for the primary's GTID, the replica's `Master_Host` is compared against the current primary's FQDN; on mismatch (including replication not configured), the replica is reconnected to the current primary via the existing `topology.ConfigureReplica` path (`WithResetMaster(false)`, GTID position untouched) and a `ReplicaSourceFix` warning event is emitted. The wait then converges instead of blocking forever. The sync requirement itself is unchanged — replicas must still fully reach the primary's GTID before promotion.

2. **Failed pre-promotion phases release the primary.** If a phase fails before `Configure new primary`, the read lock and `read_only` are released (reusing the unlock logic extracted from the stale-switchover path) and a `SwitchoverRetry` warning event is emitted. The next reconciliation re-locks and retries, so an unable-to-complete switchover costs one `syncTimeout` window of read-only per attempt instead of a sustained outage. No promotion-ordering change: unlocking only happens on failure paths that run strictly before the new primary is configured. With semi-sync, replicas are near-synced when the lock is taken, so convergence within a single attempt remains the normal case.

Also adds `sql.Client.ReplicaMasterHost` (reads `Master_Host` from `SHOW REPLICA STATUS`; empty string when replication is not configured) rather than extending `ReplicaStatusVars`, to avoid touching the CRD status API.

## Validation

- Production incident on 25.10.4 (detailed in #1852): the deadlock held the primary read-only for ~3h; after manually reconnecting the stale replica to the current primary — exactly what change 1 automates — the same switchover completed in ~10 seconds.
- `make build`, `make lint` (0 issues), `make helm-crds && make test` (all suites pass).
- No API/CRD changes, so no generated-file diffs.

### DIFF
--- a/api/v1alpha1/event_types.go
+++ b/api/v1alpha1/event_types.go
@@ -11,6 +11,11 @@ const (
 	ReasonReplicationReplicaSync = "ReplicaSync"
 	// ReasonReplicationReplicaSyncErr indicates that an error has happened while replicas were being synced with primary.
 	ReasonReplicationReplicaSyncErr = "ReplicaSyncErr"
+	// Indicates that a replica replicating from a wrong source is being reconnected to the current primary.
+	ReasonReplicationReplicaSourceFix = "ReplicaSourceFix"
+	// Indicates that a switchover attempt failed before promoting the new primary
+	// and the current primary has been unlocked until the next attempt.
+	ReasonReplicationSwitchoverRetry = "SwitchoverRetry"
 	// ReasonReplicationPrimaryNewSync indicates that the new primary is being synced.
 	ReasonReplicationPrimaryNewSync = "PrimaryNewSync"
 	// ReasonReplicationPrimaryNewSync indicates that an error has happened while the new primary was being synced.
--- a/pkg/controller/replication/switchover.go
+++ b/pkg/controller/replication/switchover.go
@@ -24,6 +24,9 @@ import (
 type switchoverPhase struct {
 	name      string
 	reconcile func(context.Context, *ReconcileRequest, logr.Logger) error
+	// Phases running before the new primary is configured: if one fails, the current primary
+	// can be safely unlocked until the next reconciliation.
+	prePromotion bool
 }
 
 func isSwitchoverStale(mdb *mariadbv1alpha1.MariaDB) bool {
@@ -70,16 +73,19 @@ func (r *ReplicationReconciler) reconcileSwitchover(ctx context.Context, req *Re
 
 	phases := []switchoverPhase{
 		{
-			name:      "Lock primary with read lock",
-			reconcile: r.lockPrimaryWithReadLock,
+			name:         "Lock primary with read lock",
+			reconcile:    r.lockPrimaryWithReadLock,
+			prePromotion: true,
 		},
 		{
-			name:      "Set read_only in primary",
-			reconcile: r.setPrimaryReadOnly,
+			name:         "Set read_only in primary",
+			reconcile:    r.setPrimaryReadOnly,
+			prePromotion: true,
 		},
 		{
-			name:      "Wait sync",
-			reconcile: r.waitSync,
+			name:         "Wait sync",
+			reconcile:    r.waitSync,
+			prePromotion: true,
 		},
 		{
 			name:      "Configure new primary",
@@ -99,6 +105,17 @@ func (r *ReplicationReconciler) reconcileSwitchover(ctx context.Context, req *Re
 		if err := p.reconcile(ctx, req, logger.WithValues("phase", p.name)); err != nil {
 			if apierrors.IsNotFound(err) {
 				return err
+			}
+			// Keeping the primary locked across reconciliations would leave applications without
+			// writes indefinitely if the switchover is unable to complete.
+			if p.prePromotion && req.currentPrimaryReady {
+				if unlockErr := r.unlockPrimary(ctx, req, logger); unlockErr != nil {
+					logger.Error(unlockErr, "Error unlocking primary after switchover phase error", "phase", p.name)
+				} else {
+					r.recorder.Eventf(req.mariadb, nil, corev1.EventTypeWarning, mariadbv1alpha1.ReasonReplicationSwitchoverRetry,
+						mariadbv1alpha1.ReasonReplicationSwitchoverRetry,
+						"Switchover phase '%s' failed. Primary unlocked until next attempt", p.name)
+				}
 			}
 			return fmt.Errorf("error in %s switchover reconcile phase: %v", p.name, err)
 		}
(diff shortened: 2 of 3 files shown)


colans:
Refreshed this PR for the 26.6.1 planning discussed in #1719 — it's up to date with `main` and pre-adapted for #1883:

- **Merged current `main`** (including 2aa29a9, the unconditional `read_only` reconciliation).
- **Dropped the explicit `WithResetMaster(false)`** from the wait-sync reconnection, so the heal path now just follows whatever binlog policy `ConfigureReplica` applies. That means no conflict with #1883, which removes that option along with `RESET MASTER` itself — this branch compiles and behaves sensibly on either side of it.
- `make helm-crds && make test` (36 suites) and `make lint` pass on the merged result.

Worth noting for the 26.6.1 scope: the planned changes (always-reconcile `read_only`, `RESET MASTER` removal, `MASTER_DEMOTE_TO_SLAVE` on dirty `gtid_binlog_pos`) don't cover the failure mode in #1852. That deadlock is on the *pre-promotion* side: `waitForReplicaSync` waits for the current primary's GTID on a replica whose replication connection points elsewhere (in our production case, at the not-yet-promoted target, which has nothing to serve it), and re-applies the read lock + `read_only` on every retry with no overall bound — a sustained write outage that looks healthy on every per-replica metric. This PR fixes both halves: verify/reconnect each replica's actual replication source before waiting on it, and release the primary when a pre-promotion phase fails so retries cost seconds of read-only rather than holding writes down indefinitely.

Since 26.6.1 is touching exactly these files, it would be a natural window to get this in — happy to adjust to whatever direction you'd prefer, including rebasing onto #1883 once it lands.

PR #1868, opened 2026-08-16
docs: document WATCH_NAMESPACE for watching specific namespaces

### LINKED ISSUE #567: [Docs] Add documentation about `WATCH_NAMESPACE` environment variable
I would like the operator to manage only specific namespaces. This can be done with environment variable `WATCH_NAMESPACE`, which I had trouble finding because it is not documented anywhere.

vshulcz:
Close MDB-4

Fixes #567.

`WATCH_NAMESPACE` is read in `pkg/environment` and feeds `mgrOpts.Cache.DefaultNamespaces` in `cmd/controller/main.go`, and it accepts a comma-separated list. The helm docs only describe the two extremes — cluster-wide, and `currentNamespaceOnly=true` for the release namespace — so scoping the operator to a specific set of namespaces is not discoverable, which is what the issue reports.

Adds a "Specific namespaces" subsection to the deployment modes in `docs/helm.md`: how to set the variable through `extraEnv`, the startup log line that confirms which namespaces were picked up, and a note that `currentNamespaceOnly=true` is the same thing scoped to the release namespace.

Docs only, no code changes.

### DIFF
--- a/docs/helm.md
+++ b/docs/helm.md
@@ -124,6 +124,26 @@ helm install mariadb-operator \
   mariadb-operator/mariadb-operator
 ```
 
+#### Specific namespaces
+
+To watch a set of namespaces rather than all of them or just the operator's own, set the `WATCH_NAMESPACE` environment variable to a comma-separated list. RBAC stays cluster-wide, so leave `currentNamespaceOnly` at its default:
+
+```bash
+helm install mariadb-operator \
+  -n databases --create-namespace \
+  --set 'extraEnv[0].name=WATCH_NAMESPACE' \
+  --set 'extraEnv[0].value=app-a\,app-b' \
+  mariadb-operator/mariadb-operator
+```
+
+The operator logs the namespaces it ends up watching on startup:
+
+```
+Watching namespaces {"namespaces": ["app-a", "app-b"]}
+```
+
+Setting `currentNamespaceOnly=true` is equivalent to setting `WATCH_NAMESPACE` to the release namespace.
+
 ## Updates
 
 > [!IMPORTANT]  
--- a/docs/helm.md
+++ b/docs/helm.md
@@ -138,9 +138,11 @@ helm install mariadb-operator \
 
 The operator logs the namespaces it ends up watching on startup:
 
+```json
+{"level":"info","ts":1756742400.123,"logger":"setup","msg":"Watching namespaces","namespaces":["app-a","app-b"]}
 ```
-Watching namespaces {"namespaces": ["app-a", "app-b"]}
-```
+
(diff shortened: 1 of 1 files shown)


mmontes11 (changes requested):
Thanks for the PR, could you make the adjustments below?

---
## 🔍 PR Review
Docs-only PR adding a "Specific namespaces" deployment mode to `docs/helm.md`, documenting `WATCH_NAMESPACE` (comma-separated list) via `extraEnv` to fix the discoverability gap reported in #567. All load-bearing claims verified against the code and chart.

### Verdict at a glance

| Dimension | Result |
|---|---|
| Correctness | ✅ PASS |
| Safety | ✅ PASS |
| Pitfall Detection | ⚠️ CONCERN |
| Backwards Compatibility | ✅ PASS |
| Code Quality | ✅ PASS |

**Overall:** 📝 Approve with notes — **Risk:** 🟢 LOW

---

### ✅ Correctness
The documented mechanism matches the implementation (env parsing, cache scoping, `extraEnv` wiring, and the `currentNamespaceOnly` equivalence).

### ✅ Safety
Docs-only change; no behavioral impact.

### ⚠️ Pitfall Detection
One minor accuracy nit: the sample startup log line in `docs/helm.md` ("Specific namespaces" section) does not match the actual output format. The operator logs via `setupLog.Info("Watching namespaces", "namespaces", namespaces)` (`cmd/controller/main.go`) with the default zap **JSON** encoder (non-development mode, `pkg/log/log.go`), so the real line looks like:

```
{"level":"info","ts":1755440000.123,"logger":"setup","msg":"Watching namespaces","namespaces":["app-a","app-b"]}
```

not `Watching namespaces {"namespaces": ["app-a", "app-b"]}`. The message string and field name are right, so the guidance is still actionable (grep for `Watching namespaces`), but a reader following the doc verbatim won't find that exact line. Suggest either showing the actual JSON line or rewording to "the operator logs a `Watching namespaces` message on startup".

### ✅ Backwards Compatibility
Additive documentation only; nothing to classify.

### ✅ Code Quality
Placement under Deployment modes is consistent with the existing `Cluster-wide` / `Single namespace` subsections (h4, so no TOC change needed), and the escaped comma in the `helm --set` example is correct.

---

### 📋 Suggested before merge
- [ ] Fix the sample log line to match the actual JSON output, or reword it to reference the `Watching namespaces` message only.

vshulcz:
fixed, thanks for catching it.

checked the defaults while i was in there: `--log-dev` is `false` and `--log-time-encoder` is `epoch` (`cmd/controller/main.go`), so the json encoder with a float ts is what you actually get. the doc now shows that line and points at grepping for `Watching namespaces` instead of the whole thing, since the shape moves with those two flags.

mmontes11 (approved):
LGTM. Thanks for your contributions.

Issue #1573, opened 2026-01-07
[Feature]  Generic ephemeral volume for stagingStorage


vaclavekp:
**Is your feature request related to a problem? Please describe.**
We have huge databases so we need to use CSI for stagingStorage, and since csi we use does not natively support ephemeral, we have persistent pvc even we dont need it at the time and pay for it.

**Describe the solution you'd like**

Just like described in official docs - https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#generic-ephemeral-volumes
it would solve our problem having the staging storage volume bound to pods lifecycle.

**Describe alternatives you've considered**

**Environment details**:
- Kubernetes version: 1.35
- Kubernetes distribution: Talos
- MariaDB Operator version: 25.10.3
- MariaDB topology: Standalone
- MariaDB Server version: 11.3
- Install method: Helm

**Additional context**
We are dealing this on our cluster on Hetzner Cloud.

PR #1873, opened 2026-08-28
add ephemeral volume source option

dmaes:
Allows using Ephemeral volumes, mostly useful for staging storage for backups, or storage for the PhysicalBackup used to fix replication.

also refs #1573


dmaes:
Hmm, seems like EphemeralStorage was deliberately added only for `MariaDB` by use of new `MariaDBStorageVolumeSource` , not for general `StorageVolumeSource`. (https://github.com/mariadb-operator/mariadb-operator/pull/1618#discussion_r2901662472) . But there actually _is_ a use case for having it in `StorageVolumeSource`.

@mmontes11 thoughts ?

Issue #1791, opened 2026-06-08
[Feature] MariaDB 12.3 support


mmontes11:
Currently, integration tests with [mariadb:12.3.2](https://hub.docker.com/layers/library/mariadb/12.3.2/images/sha256-181bbc639b0bef9ded18543efba92730b6ae8815102e090bf2c98aea7a247099) are failing:
https://github.com/mariadb-operator/mariadb-operator/actions/runs/27136282252/job/80089459603

Both the tests above this [slack thread ](https://mariadb-community.slack.com/archives/C02R1CJ0LD6/p1780910262062239) are reporting the same error:
```bash
Error 4243 (HY000): Cannot execute RESET MASTER as the binlog is in use by a connected slave or other RESET MASTER or binlog reader. Check SHOW PROCESSLIST for \"Binlog Dump\" commands and use KILL to stop such readers
```

Additional information about MariaDB 12.3:
- https://mariadb.com/docs/release-notes/community-server/12.3/mariadb-12.3-changes-and-improvements
- https://mariadb.com/docs/server/server-management/install-and-upgrade-mariadb/upgrading/mariadb-community-server-upgrade-paths/upgrading-from-mariadb-11-8-to-mariadb-12-3
- https://mariadb.com/docs/server/server-management/server-monitoring-logs/binary-log/innodb-based-binary-log

PR #1883, opened 2026-08-31
fix(replication): do not perform RESET MASTER as part of replica configuration

mmontes11:
Closes MDB-10
Closes MDB-21
Related to https://github.com/mariadb-operator/mariadb-operator/issues/1791

## To be evaluated:
- https://mariadb.com/docs/server/reference/sql-statements/administrative-sql-statements/replication-statements/change-master-to#master_demote_to_slave

## What

`RESET MASTER;` is no longer issued as part of the replica configuration.

The statement purges the replica's binary logs. That is not needed to configure replication, and it is destructive once the node is promoted: its binary log history is gone, and it drifts from whatever has already been archived to object storage. The operator had accumulated two opt-outs for exactly those reasons — one for point-in-time recovery and one for multi-cluster replicas — which is a good signal the step should not have been there in the first place. Dropping it removes the option along with all its call sites.

## Changes

- `pkg/controller/replication/topology.go:157` — `singleClusterTopology.ConfigureReplica` no longer calls `client.ResetMaster`; the replica configuration now starts at `STOP SLAVE`.
- `pkg/controller/replication/topology.go:24` — `ResetMaster` removed from `ConfigureReplicaOpts`, along with the `WithResetMaster` option and the now-unused `slices` import.
- `pkg/controller/replication/topology.go:286` — `multiClusterTopology.ConfigureReplica` forwards `replicaOpts` unchanged instead of cloning them to append `WithResetMaster(false)`.
- `pkg/controller/replication/controller.go:322` and `pkg/controller/replication/switchover.go:446` — the point-in-time recovery opt-out is gone from both replica-opts builders.

No API, CRD or Pod template surface is touched, so existing `MariaDB` resources are unaffected on operator upgrade and no rolling update is triggered.

## Follow-up required before release — MDB-12

Not issuing `RESET MASTER;` makes it the norm for a promoted node to still hold binary logs from a **previous primary term of its own**. That exposes a pre-existing bug in `singleClusterTopology.ConfigurePrimary` (`pkg/controller/replication/topology.go:127`): when clearing `gtid_slave_pos` fails with `Error 1948`, the function returns `nil` early and leaves the new primary half configured — `read_only` still enabled and the replication user SQL not reconciled, so `gtid_binlog_pos` never advances past the leftover logs. The stale position is then handed to the demoted primary, rejected with `Error 1947`, and the switchover is aborted for good, leaving every node read-only.

In practice this means the **second** switchover of a replication cluster wedges it in a permanent write outage. It is tracked separately as MDB-12 and must land alongside this PR, or at least ahead of any release including it.

## Verification

- `make build` — passes.
- `make lint` — 0 issues.
- `make test` — 35 suites; only `internal/helmtest` fails, and it fails identically on an unmodified tree (`helm dependency build` has not been run in this environment: *missing in charts/ directory: mariadb-operator-crds*). Unrelated to this change.
- `go test ./pkg/controller/replication/... ./pkg/sql/...` — `ok`.
- Integration tests were **not** run — no KIND cluster was provisioned in this environment. The switchover behaviour above is the path that needs integration coverage, and MDB-12 carries the regression spec for it.

### DIFF
--- a/internal/controller/mariadb_controller_replication_test.go
+++ b/internal/controller/mariadb_controller_replication_test.go
@@ -139,27 +139,6 @@ var _ = Describe("MariaDB replication", Ordered, func() {
 			return mdb.IsReady() && mdb.Status.CurrentPrimaryPodIndex != nil
 		}, testHighTimeout, testInterval).Should(BeTrue())
 
-		switchPrimaryTo := func(podIndex int) {
-			By(fmt.Sprintf("Expecting MariaDB to eventually update primary to index '%d'", podIndex))
-			Eventually(func(g Gomega) bool {
-				g.Expect(k8sClient.Get(testCtx, key, mdb)).To(Succeed())
-				mdb.Spec.Replication.Primary.PodIndex = &podIndex
-				g.Expect(k8sClient.Update(testCtx, mdb)).To(Succeed())
-				return true
-			}, testTimeout, testInterval).Should(BeTrue())
-
-			By(fmt.Sprintf("Expecting MariaDB to eventually change primary to index '%d'", podIndex))
-			Eventually(func() bool {
-				if err := k8sClient.Get(testCtx, key, mdb); err != nil {
-					return false
-				}
-				if !mdb.IsReady() || mdb.Status.CurrentPrimaryPodIndex == nil {
-					return false
-				}
-				return *mdb.Status.CurrentPrimaryPodIndex == podIndex
-			}, testHighTimeout, testInterval).Should(BeTrue())
-		}
-
 		originalPrimary := *mdb.Status.CurrentPrimaryPodIndex
 		var newPrimary int
 		for i := 0; i < int(mdb.Spec.Replicas); i++ {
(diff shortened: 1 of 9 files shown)


mmontes11:
**Hard-failover spec is realistic — reproduced 4/4 on local KIND**

I ran the `should fail and switch over primary` spec's exact trigger against a local KIND cluster (3-replica native replication, operator built from this PR's head `a6064aa`): hard-delete the current primary pod with `gracePeriodSeconds: 0` + foreground propagation, then watch. In **every** configuration the cluster **wedges at 2/3**: `status.currentPrimary` moves, but `mdb.IsReady()` never flips back to true, so the spec's `Eventually(mdb.IsReady())` (`internal/controller/mariadb_controller_replication_test.go:150`) times out. It reproduces deterministically — not a flake.

| log_slave_updates | empty | 50k rows | wedged node | error |
|---|---|---|---|---|
| OFF (single-cluster default) | WEDGE 2/3 | WEDGE 2/3 | demoted primary | 1236 |
| ON | WEDGE 2/3 | WEDGE 2/3 | a secondary replica | 1950 |

Data volume doesn't change the outcome — a handful of DDL GTIDs is enough.

**Two distinct signatures, both GTID-state incoherence after a failover that promotes a different server-UUID:**

`log_slave_updates=OFF` (the operator only sets it in multi-cluster mode, so this is the default here): the **demoted primary** rejoins requesting its pre-failover GTID, which the new primary never wrote to its binlog (replicated txns aren't re-logged when the var is off):

```
[ERROR] Slave I/O: Got fatal error 1236 from master: 'connecting slave requested to start from GTID 0-10-8, which is not in the master's binlog ... the slave has diverged'
```

`log_slave_updates=ON`: the demoted primary rejoins cleanly, but a **secondary replica** is re-pointed to the new primary **by binlog position** (relay log starts at `GTID position ''`), so its SQL thread re-reads already-applied GTIDs out of order and `gtid_strict_mode` aborts it:

```
[ERROR] Slave SQL: An attempt was made to binlog GTID 0-10-1 which would create an out-of-order sequence number with existing GTID 0-10-99, and gtid strict mode is enabled ... error code: 1950
```

Either way the replica I/O/SQL thread dies, the agent liveness probe (`pkg/agent/handler/replication/probe.go:75`) returns 500, the node crashloops, and the operator never re-establishes 3/3 — so the spec's readiness gate never opens.

**Why this PR surfaces it:** removing `RESET MASTER` from `ConfigureReplica` dropped the step that reset a rejoining node's GTID history. `RESET MASTER` is the wrong tool (destructive — it discards binlog/GTIDs), but it was masking the gap: there's no coherent GTID-reconciliation path for a node re-joining after a same-cluster failover. Worth deciding how the rejoin should be GTID-safe before closing MDB-10 — e.g. force GTID-based replication (`log_slave_updates=ON` + `CHANGE MASTER ... USING GTID` / `MASTER_AUTO_POSITION=1`) and/or a targeted GTID fixup for the demoted node. Happy to open a follow-up issue.

colans:
FYI from the neighbouring PR: #1853 (the #1852 pre-promotion switchover-deadlock fix) also touches `switchover.go`/`topology.go`/`sql.go`, and I've just updated it to be compatible with this change — its replica-reconnection path no longer references `WithResetMaster`, so the two merge cleanly in either order.

One scope note in case it helps triage: removing `RESET MASTER` doesn't affect the #1852 failure mode — there the stuck replica is wired to the not-yet-promoted target, which has nothing to serve it regardless of binlog retention (replicas don't run `log_slave_updates`), while `waitForReplicaSync` holds the current primary read-only across unbounded retries. So the two fixes are complementary rather than overlapping.

mmontes11:
Thanks for looking into it @colans, `RESET MASTER` is really a no-go for the MariaDB 12.3 we will be supporting in 26.10. We still have some PRs to go through, we'll eventually get there to review.

PR #1894, opened 2026-09-03
Demote rejoining nodes with self-owned GTIDs via MASTER_DEMOTE_TO_SLAVE

mmontes11:
## Summary

Follow-up to #1883 (removal of `RESET MASTER`). Fixes the hard-failover wedge: when a primary is hard-failed (pod deleted) and recreates as a replica, it keeps its self-originated GTIDs (non-empty `gtid_binlog_pos`). Re-pointing it with `MASTER_USE_GTID` cannot reconcile those self-owned GTIDs, so the node never converges — `Enabling readonly` loops and the pod's readiness churns.

## Change

In `changeMaster`, read `gtid_binlog_pos` and, when it is non-empty, emit `CHANGE MASTER ... MASTER_DEMOTE_TO_SLAVE=1` instead of `MASTER_USE_GTID`. That merges `gtid_binlog_pos` into `gtid_slave_pos` — the non-destructive replacement for the `RESET MASTER` that used to run here.

Fresh replicas (empty `gtid_binlog_pos`) are untouched and keep the `MASTER_USE_GTID` path.

## Files

- `pkg/controller/replication/topology.go:241` — read `gtid_binlog_pos`; gate demote on non-empty
- `pkg/sql/sql.go` — `Demote` field, `WithChangeMasterDemote`; mutually-exclusive template clause (`MASTER_DEMOTE_TO_SLAVE=1` or `MASTER_USE_GTID=...`)
- `pkg/sql/sql_test.go` — demote query case

## Verification

- `go build ./...`, `go test ./pkg/sql/... ./pkg/controller/replication/...`, `make lint` — all green.
- Operator-failover matrix across 4 scenarios (empty / with-data × `log_slave_updates` OFF/ON) evaluated on a local KIND cluster — results in the comment below.

Closes MDB-21

### DIFF
--- a/make/dev.mk
+++ b/make/dev.mk
@@ -249,7 +249,7 @@ dump:  ## Dump cluster state for debugging purposes
 	@for pod in $$($(KUBECTL) get pods -n $(DUMP_NAMESPACE) -o jsonpath='{.items[*].metadata.name}'); do \
 		if [[ "$$pod" == csi* ]]; then continue; fi; \
 		printf "\n---- Logs for $$pod ----\n\n"; \
-		$(KUBECTL) logs $$pod -n $(DUMP_NAMESPACE) --all-containers=true; \
+		$(KUBECTL) logs $$pod -n $(DUMP_NAMESPACE) --all-containers=true || true; \
 	done
 
 ##@ Profile
--- a/pkg/agent/handler/replication/probe.go
+++ b/pkg/agent/handler/replication/probe.go
@@ -72,8 +72,10 @@ func (p *ReplicationProbe) Liveness(w http.ResponseWriter, r *http.Request) {
 
 		replicaIORunning := ptr.Deref(status.SlaveIORunning, false)
 		if !replicaIORunning {
-			p.livenessLogger.Error(nil, "Replica IO thread not running")
-			p.responseWriter.WriteError(w, "Replica IO thread not running")
+			p.livenessLogger.Error(nil, "Replica IO thread not running",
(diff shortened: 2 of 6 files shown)


mmontes11:
## Local KIND verification — hard-failover path (3/3 stable)

Follow-up to the failover matrix above: I ran the focused integration spec for this exact regression (`should fail and switch over primary`, plus `should reconcile` and `should reconcile with MaxScale`) three independent times on a local KIND cluster (3-replica, AutoFailover, TLS, MetalLB LB) to check whether the hard-failover wedge is gone or only intermittent.

| Run | Result | Duration |
|-----|--------|----------|
| 1 | PASS (3/3 specs) | 204s |
| 2 | PASS (3/3 specs) | 172s |
| 3 | PASS (3/3 specs) | 184s |

Each run is from-scratch (fresh `BeforeSuite` fixtures, torn down by `DeferCleanup` on clean exit) — three independent failovers, not one shared cluster.

The spec genuinely exercises the regression path: it deletes the current primary pod, then asserts the cluster reconverges with a new `currentPrimary`. Across all three runs the old primary re-joined as a replica cleanly — no `Enabling readonly` loop, no `Last_IO_Errno=1236`, no readiness churn — and the operator completed the switchover in order (`Lock primary` → `read_only` → `Wait sync (gtid)` → `Configure new primary` → `Ready`).

**Verdict:** the `MASTER_DEMOTE_TO_SLAVE` path holds up under repeated hard failover locally and matches the green CI run. The earlier local failures were environment noise (a stale `:8080` holder + a non-idempotent SSEC secret), not the bug.

Issue #1915, opened 2026-09-18
[Bug] PITR silently skips binlog replay: nil pointer panic when s3.tls.enabled is true and caSecretKeyRef is omitted


hrens:
**Documentation**
- [x] I acknowledge that I have read the relevant [documentation](https://github.com/mariadb-operator/mariadb-operator/tree/main/docs).

**Describe the bug**

A point-in-time restore completes with `Ready=True` but recovers **none** of the requested data: only the base physical backup is restored, no binary logs are replayed, and no `BinlogsReplayed` condition is ever set. There is no event, no error condition and nothing in the `MariaDB` status to suggest the recovery was incomplete.

The operator is in fact panicking on every reconcile of the restoring instance:

```
{"level":"error","msg":"Observed a panic","controller":"mariadb",
 "MariaDB":{"name":"mariadb-test-restore","namespace":"tsa-test"},
 "panic":"runtime error: invalid memory address or nil pointer dereference",
 "stacktrace":"...
   github.com/mariadb-operator/mariadb-operator/v26/pkg/minio.NewMinioClientFromS3Config
     /app/pkg/minio/minio.go:146
   internal/controller.(*MariaDBReconciler).getS3Client
     /app/internal/controller/mariadb_controller_pitr.go:394
   internal/controller.(*MariaDBReconciler).getStorageClient
     /app/internal/controller/mariadb_controller_pitr.go:339
   internal/controller.(*MariaDBReconciler).reconcileReplayBinlogsError
 ..."}
```

`pkg/minio/minio.go:143-146` dereferences `caSecretKeyRef` without a nil check, although the field is optional:

```go
tls := ptr.Deref(s3.TLS, v1alpha1.TLSConfig{})
if tls.Enabled {
    minioOpts = append(minioOpts, WithTLS(true))
    caCertBytes, err := refResolver.SecretKeyRef(ctx, *s3.TLS.CASecretKeyRef, namespace)  // <-- panics
    ...
}
```

[...]

PR #1917, opened 2026-09-18
Treat s3 and abs tls.caSecretKeyRef as optional

harshilv17:
Closes the nil dereference reported in #1915, and the same bug on the Azure Blob path.

### The bug

`CASecretKeyRef` is marked `+optional` and documented to fall back to the system trust chain:

> By default, the system trust chain will be used, but you can use this field to add more CAs to the bundle.

Both storage clients dereferenced it unconditionally whenever `tls.enabled` was true:

- **`pkg/minio/minio.go`** — the panic in the report. In the PITR path it fires inside `reconcileReplayBinlogsError` *before* any status patch, so binlog replay silently does not happen: no `BinlogsReplayed` condition is set, and the instance still reports `Ready` while holding only the base backup.
- **`internal/controller/mariadb_controller_pitr.go`** (`getABSClient`) — the identical nil dereference on the Azure Blob path. Not in the report, same root cause.

`pkg/builder` already guards this correctly in all six places it reads the field (`tls.Enabled && tls.CASecretKeyRef != nil`); only the two runtime client paths did not.

### Why the Azure transport also changed

Guarding the dereference alone was not enough for Azure. `getTransport` built an empty pool and called `AppendCertsFromPEM` unconditionally, so TLS with no CA bundle turned the panic into `unable to add CA cert to pool`. It now starts from the system trust chain, matching the minio transport and the documented behaviour that a CA bundle *adds to* the bundle rather than replacing it.

Worth a closer look in review, since it also means an Azure CA bundle now augments the system pool instead of replacing it.

### Tests

Both new tests fail without the fix and pass with it:

- `pkg/minio` — `TestNewMinioClientFromS3ConfigOptionalCACert` reproduces the reported `invalid memory address or nil pointer dereference`. A nil-backed `RefResolver` is deliberate: with no `CASecretKeyRef` no Secret may be resolved at all.
- `pkg/azure` — `TestGetTransport` covers TLS off, TLS on with no CA bundle, and an invalid bundle.

### Not included

The second ask in #1915 — surfacing a panic in the PITR phase as `BinlogsReplayed=False`. That is defensive hardening of the reconcile phase rather than this root cause, so I left it out to keep the fix reviewable. Happy to follow up in a separate PR if you would like it.


harshilv17:
Thanks for approving the run. It caught a real lint failure of mine — `MinioOpts` is embedded in `MinioClient`, so `client.MinioOpts.TLS` tripped `QF1008`. Fixed in cf94690, which uses the field directly.

The other failure was `should create replica cluster from physical backup [multi-cluster]`, which timed out after 300s on:

```
error reconciling Replication: error getting agent clientset:
  error reading TLS CA bundle: Secret "mariadb-eu-central-ca-bundle" not found
```

I don't think that one is mine — this PR only touches the S3 and ABS backup storage CA handling, and that failure is in the cluster's own agent/replication TLS. The same `test-int-multi-cluster` suite passed on `main` earlier today (30 passed), and my run also logged a flake in the suite before it, so it looks environmental. I'd rather not assert that without evidence though, so a re-run would settle it.

Pushing the lint fix moved the head, so the workflows are gated again — whenever you have a moment, another approval would confirm both.

PR #1923, opened 2026-09-19
Add maxscale filters and volumes

### LINKED ISSUE #1924: [Feature] Add Filters support to MaxScale CRD
Add Filters support to MaxScale CRD:

* https://mariadb.com/docs/maxscale/mariadb-maxscale-tutorials/filters
* https://mariadb.com/docs/maxscale/reference/maxscale-filters

hedgieinsocks:
* Add Filters support to MaxScale CRD
* Add Volumes support to MaxScale CRD

Closes https://github.com/mariadb-operator/mariadb-operator/issues/1924


pkg/maxscale/client/generic_client.go
```diff
@@ -31,7 +31,7 @@ func WithQuery(q map[string]string) Option {
 func WithForceQuery() Option {
 	return func(o *Options) {
 		o.query = map[string]string{
-			"force": "true",
```
mmontes11, line 34: Any issues in your MaxScale version? I believe both are accepted:
https://mariadb.com/docs/maxscale/maxscale-management/deployment/installation-and-configuration/maxscale-configuration-guide#booleans

mmontes11 (changes requested):
Will be worth adding:
- Example manifest with filters added
- Minimal docs to setup a filter, something in the same line as [Server Configuration](https://github.com/mariadb-operator/mariadb-operator/blob/main/docs/maxscale.md#server-configurations). Or even update that same section adding a filter to one of the services, leave it up to you.
- Add [filter definition ](https://mariadb.com/docs/maxscale/reference/maxscale-filters) in our [small glossary](https://github.com/mariadb-operator/mariadb-operator/blob/main/docs/maxscale.md#maxscale-resources)

Happy to support this, and thank you very much for your contribution! Left some comments.

Actually, we left something behind and we need to release a patch version: [26.10.1](https://github.com/mariadb-operator/mariadb-operator/pull/1922), so we might be able to squeeze this in.

pkg/maxscale/client/generic_client.go
hedgieinsocks: I faced an issue when a test filter config file refused to be deleted when I removed it from maxscale crd (default 28.03.13-2). I asked claude to debug it and it blamed this bool. Changing it resolved the issue and I accepted the solution (yeah, lazy friday meh). But now, I think I should rebuild it with `true` and give it another test, maybe it was just a coincidence. 

Also, so far, I've only verified it builds the config properly, edits it, and removes it. But I also want to test whether the configured filter actually works lol. But I'll do it only on Sunday/Monday. So this should not block your hotfix release.

mmontes11: Sure, no rush! Hit request another review when review ready, I'll take another look.

mmontes11:
Docs look good! Thanks for updating!

pkg/maxscale/client/generic_client.go
hedgieinsocks: You were right, `true` still works, reverted, trust in claude lowered :D

hedgieinsocks:
Did some manual tests with the following mariadb-cluster helm release:

```yaml
mariadb:
  galera:
    enabled: false
  tls:
    enabled: false
  replicas: 2
  replication:
    enabled: true
    semiSyncEnabled: false
    standaloneProbes: true
    replica:
      gtid: SlavePos
      maxLagSeconds: 1
maxscale:
  enabled: true
  filters:
    - name: throttle
      module: throttlefilter
      params:
        max_qps: "1"
        throttling_duration: 2000ms
  services:
  - listener:
      name: rw-router-listener
      port: 3306
      protocol: MariaDBProtocol
    name: rw-router
    params:
      max_replication_lag: 1s
    router: readwritesplit
    filters:
    - throttle
  monitor:
    interval: 1s
    params:
      auto_failover: "false"
      auto_rejoin: "false"
users:
  - name: infra
    host: "%"
    # kubectl create secret generic infra --from-literal=password=password
    passwordSecretKeyRef:
      name: infra
      key: password
grants:
  - name: infra
    username: infra
    host: "%"
    database: "*"
    table: "*"
    privileges:
      - ALL PRIVILEGES
```

```sh
❯ k port-forward service/mariadb-cluster-maxscale 3306:3306

❯ mariadb -u infra -ppassword -h 127.0.0.1 --skip-ssl -e "select 1; select 2; select 3; select 4; select 5; select 6; select 7;"
+---+
| 1 |
+---+
| 1 |
+---+
--------------
select 2
--------------

ERROR 1927 (08S01) at line 1: Connection killed by MaxScale

❯ stern maxscale -i throttle
+ mariadb-cluster-maxscale-0 › maxscale
mariadb-cluster-maxscale-0 maxscale 2026-09-20 13:39:51   notice : Module 'throttlefilter' loaded from '/usr/lib64/maxscale/libthrottlefilter.so'.
mariadb-cluster-maxscale-0 maxscale 2026-09-20 13:39:51   notice : Created filter 'throttle'
mariadb-cluster-maxscale-0 maxscale 2026-09-20 13:54:14   notice : (1255) [throttlefilter] (rw-router); Query throttling Session 1255 user infra, throttling limit reached. Disconnect.
```

Editing and removing filter seems to work as expected, filter file in `/var/lib/maxscale/maxscale.cnf.d` is created and removed, the reference in router file is also added and removed fine.

pkg/maxscale/client/generic_client.go
mmontes11: Still helps a lots with the heavy lifting, but yeah intelligent sand decisions are not always fully accurate.

mmontes11 (approved):
LGTM. I can confirm it works using the example in the PR:

```bash
 mariadb -u  maxscale-repl-client -p'<redacted>'   -h maxscale-repl  --skip-ssl  -e "select 1; select 2; select 3; select 4; select 5; select 6; select 7;"
+---+
| 1 |
+---+
| 1 |
+---+
--------------
select 2
--------------

ERROR 1927 (08S01) at line 1: Connection killed by MaxScale
```

mmontes11:
@hedgieinsocks FYI When reviewing the [26.10.1 PR](https://github.com/mariadb-operator/mariadb-operator/pull/1922/changes#diff-df46f240b8f1052c37bce587db7356c71acba67aea6b6c1356b825b19f982b10) I thought we should probably keep the filters functionality more scoped, and remove any side effect in the rest of MaxScale objects (i.e. keep `RelationshipData` as it currently was and introduce a dedicated `FilterRelationshipData`):
- https://github.com/mariadb-operator/mariadb-operator/pull/1932

